PR #18, opened 2019-01-12
Add support for system alerts

adamstep:
[TAD](https://instawork.atlassian.net/wiki/spaces/EN/pages/794001412/Proposal+Alert+element)

Added support for "alert" behaviors. These behaviors describe the alert options as `<alert:option>` elements, that can then contain behaviors that trigger when the user presses the option:

```xml
<view style="Button">
  <behavior
    trigger="press"
    action="alert"
    alert:title="Alert title"
    alert:message="Alert message"
  >
    <alert:option alert:label="Push" trigger="press" href="/index.xml" action="push" />
    <alert:option alert:label="Back" trigger="press" href="#" action="back" />
    <alert:option alert:label="Cancel" />
  </behavior>
  <text style="Button__Label">Navigation options</text>
</view>
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,7 @@ website/build/
 website/yarn.lock
 website/node_modules
 website/i18n/*
+
+.yalc
+.yalcignore
+demo/yalc.lock
--- a/.babelrc
+++ b/.babelrc
@@ -5,5 +5,8 @@
       "root": ["."],
       "alias": {"hyperview": "./"}
     }]
-  ]
+  ],
+  "ignore": [
+    "demo"
+  ],
 }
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -11,3 +11,7 @@ website/build/
 website/yarn.lock
 website/node_modules
 website/i18n/*
+
+.yalc
```
flochtililoch: Can you add details about yalc dependency - I had a brief look at https://github.com/whitecolor/yalc, and I assume this has to be installed globally according to their instructions, so we should document it in the README. Any way we could make it a hard dependency instead of installing globally (to skip the extra manual step when setting up the demo app)?

adamstep: I'm not sure yet if this should be included as a hard dependency, it was just a tool I was testing out to get something working. But I'd prefer to discuss with more people before deciding whether we should use it or a different approach.

adamstep: Also, I realized I accidentally checked in changes to package.json that use yalc... I'll revert that.

PR #22, opened 2019-01-18
Add call-or-text custom action

dikarel:
This custom action is added for a feature wherein we pop up a bottom sheet asking if the user wants to call or send an SMS to someone. This is a common-enough pattern that I thought it might belong here.


adamstep:
As discussed, we should close this PR since it's not needed immediately. Let's try to build system action sheet into Hyperview as a first-class behavior.

PR #27, opened 2019-02-05
Show helpful XML parse errors

adamstep:
https://app.asana.com/0/518926233438259/1106456911336593/f

When parsing a screen, show errors if the XML does not parse correctly. In addition, show errors if the XML is missing `<doc>`, `<screen>`, or `<body>` tags. These errors show the default error screen that allows the screen to be reloaded, allowing faster iteration.

screen shot 2019-02-05 at 11 16 28 am
screen shot 2019-02-05 at 11 16 46 am
screen shot 2019-02-05 at 11 16 58 am
screen shot 2019-02-05 at 11 24 13 am

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -490,15 +490,21 @@ export default class HyperScreen extends React.Component {
   constructor(props) {
     super(props);
 
+    this.onUpdate = this.onUpdate.bind(this);
+    this.shallowClone = this.shallowClone.bind(this);
+    this.shallowCloneToRoot = this.shallowCloneToRoot.bind(this);
+    this.reload = this.reload.bind(this);
+    this.parseError = this.parseError.bind(this);
+
     this.navActions = ['push', 'new', 'back', 'close', 'navigate'];
     this.updateActions = ['replace', 'replace-inner', 'append', 'prepend'];
 
     this.parser = new DOMParser({
       locator: {},
       errorHandler: {
-        warning: (w) => console.error(w),
-        error: (e) => console.error(e),
-        fatalError: (e) => console.error(e),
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -490,25 +490,45 @@ export default class HyperScreen extends React.Component {
   constructor(props) {
     super(props);
 
+    this.onUpdate = this.onUpdate.bind(this);
```
flochtililoch: FWIW, because we're using arrow functions in this components, we should be able to remove these bind operations

adamstep: good point, I can clean that up in a follow-up PR.

src/index.js
```diff
@@ -610,11 +630,34 @@ export default class HyperScreen extends React.Component {
...
+        let error = false;
         const animations = createAnimations(doc);
         const stylesheets = Stylesheets.createStylesheets(doc);
         ROUTE_KEYS[getHrefKey(url)] = this.props.navigation.state.key;
 
+        // Make sure the XML has the required elements: <doc>, <screen>, <body>.
+        const docElement = getFirstTag(doc, 'doc');
+        if (!docElement) {
+          console.error(`No <doc> tag found in the response from ${url}.`);
+          doc = null;
```
flochtililoch: why do we need to set doc to `null`?
also wondering if a more readable pattern would be to throw an error and let the `catch` handler log the error/set the error state for us

adamstep: just cleaning up the state to not have references to the previous or current document when an error occurs. The previous document can be set in cases where we did a reload of the current screen, or if there was a custom preloading state.

PR #28, opened 2019-02-06
Remove animation implementation

adamstep:
Hyperview had an old animation implementation that wasn't fully supported. This was designed before the full behavior system. With behaviors, there's a better path for supporting animations in the future, so we can ditch the old system.

Also: fixed prettier config to not change double quotes to single quotes in Markdown files. This was breaking document generation.


adamstep:
@divyanshu013 you're right... I can try reverting my prettier config and remove unecessary quotes in the md file. That may work.

adamstep:
@divyanshu013 ok that worked... code in the docs now follows conventions.

PR #32, opened 2019-02-11
infra: add hyperview watch mode for demo app

divyanshu013:
With this PR we can develop hyperview features in watch mode. Changes in src should directly get reflected in the demo app with the `build:demo` command running. This also saves us from the "duplicate modules" error which happens if we're using `yarn link`.

Also updated readme.


adamstep:
@flochtililoch can you lead the review on this? I know you've been doing some thinking here too.

PR #39, opened 2019-03-02
Do not attempt to send a POST request using an empty FormData

dikarel:
On Android, attempting to send a POST request using an empty `FormData` will cause a crash.

**Test plan**

- [x] Verify fix on Android


src/index.js
```diff
@@ -729,7 +729,14 @@ export default class HyperScreen extends React.Component {
           formData.append(name, input.getAttribute('value'));
         }
       });
-    return formData;
+
+    let formHasData = false;
+    for (var key of formData.keys()) {
```
dikarel: Couldn't find a quick way to see if a `FormData` is empty. Consequently, had to rely on the good 'ole for-loop. Happy to accept suggestions for how to do this non-imperatively

flochtililoch: Would this work?
```js
const hasFormData = formData.entries().length > 0;
```

dikarel: @flochtililoch tried it and didn't work. It appears ([from docs I found](
https://developer.mozilla.org/en-US/docs/Web/API/FormData)) that although `FormData` has helpers that return an iterator, there's no quick way to grab length.

screen shot 2019-03-02 at 7 21 17 pm

flochtililoch: After doing some experimenting in the "Basic forms" case study of the demo app, it does not appear that `FormData` instances have `keys` or `entries` methods. The object looks like:
```json
{"_parts":[["data", "Test"]]}
```
(`Test` being the value typed in the text input field)

adamstep: Is that part of the internal implementation though? Might not be a good idea to rely on the internals, which may change in the future.

divyanshu013:
```suggestion
    for (let key of formData.keys()) {
```

dikarel: @adamstep I checked for properties of `FormData` (by printing `Object.getOwnPropertyNames(formData)`), and there doesn't seem to be other methods or properties other than `_parts`

dikarel: I settled on another way that doesn't involve having to inspect `formData`

PR #43, opened 2019-03-05
Infra: iterate on the build:demo command

flochtililoch:
`yarn build:demo` now copies the content of `src/` into the demo app hyperview dependency module, and runs a watcher. This enables the developer to see the actual source code instead of the minified version when attaching to debugger

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   },
   "scripts": {
     "build": "babel src --out-dir lib",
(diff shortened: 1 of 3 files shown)


flochtililoch:
I’m actually thinking moving this script to the `demo/package.json` As it really pertain to getting the demo app synced with the hyperview code, less about the other way around. Re naming, how about `yarn sync`? :trollface:

PR #45, opened 2019-03-06
Add SMS custom behavior example page

dikarel:
| Page | On click |
| - | - |
| screen shot 2019-03-05 at 4 10 12 pm | screen shot 2019-03-05 at 4 15 38 pm |

**Test plan**

- [x] Test locally on iOS simulator


flochtililoch:
In my upcoming PR (expected tomorrow), I’m extracting the behaviors away from hyperview core and will move them to our apps codebase (and will deprecate at the same time these props you’ve added stubs for) - the only two behavior that will remain part of the core will be `share` and `alert`, so I think this code change won’t be needed

PR #48, opened 2019-03-06
Infra: Behavior registry

flochtililoch:
Extract 3rd party behaviors and isolate core behaviors in their own module

https://app.asana.com/0/244065166232575/938009554003242/f


src/types.js
```diff
@@ -283,3 +283,12 @@ export type HvComponent<Props> = ComponentType<Props> & {
   localName: LocalName | string,
   namespaceURI: NamespaceURI,
 };
+
+export type HvBehavior = {
+  action: string,
+  callback: (element: Element, onUpdate: HvComponentOnUpdate) => any,
```
dikarel: - Curious about the `any`. What happens to any returned values here? If we don't do anything with them, I think it's best to make this `void` or `Promise<void>`
- I think it might be good to add documentation around the Flow type to explain the effects of each attribute (e.g. what do we do with the return value of `callback`? Is `action` supposed to be the name of the action?)

src/services/namespaces/index.js
```diff
@@ -13,9 +13,4 @@ export const HYPERVIEW = 'https://hyperview.org/hyperview';
 export const HYPERVIEW_ALERT = 'https://hyperview.org/hyperview-alert';
 
 // Custom namespaces
```
dikarel: Perhaps this comment is now outdated? Since we've moved out all "custom" 3rd party behavior out of this lib

src/services/index.js
```diff
@@ -116,3 +116,6 @@ export const createProps = (
 
   return props;
 };
+
+export const later = (delay: number): Promise<void> =>
```
dikarel: Should we add units suffix here? As per engineering patterns

src/behaviors/hv-alert/index.js
```diff
@@ -0,0 +1,86 @@
...
+
+import * as Namespaces from 'hyperview/src/services/namespaces';
+import type { Element, HvComponentOnUpdate } from 'hyperview/src/types';
+import { getBehaviorElements, later } from 'hyperview/src/services';
+import { Alert } from 'react-native';
+
+export default {
+  action: 'alert',
+  callback: (element: Element, onUpdate: HvComponentOnUpdate) => {
+    const title = element.getAttributeNS(Namespaces.HYPERVIEW_ALERT, 'title');
```
dikarel: Should we additionally add type-checks here? There's nothing stopping someone from inserting a boolean or numbers here by accident, and if we explicitly catch that here the debugging will be easier.

src/behaviors/hv-share/types.js
```diff
@@ -0,0 +1,20 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+export type Content =
+  | { title?: string, message: string }
+  | { title?: string, url: string };
+
+export type Options = {
```
dikarel: Maybe it'd be good to point out where this came from?

src/behaviors/hv-alert/index.js
dikarel: Looks like getAttributeNS always returns `string` or `null`. In that case, all we have to do is ensure required fields aren't `null`. Out of scope for this PR though

flochtililoch: There are no flow warnings because these fields are optional in [Alert.alert of RN 0.52.2](https://github.com/facebook/react-native/blob/4080141a93127ba64e2d418c73e95b36f4eb5eee/Libraries/Alert/Alert.js#L44-L45)

src/services/namespaces/index.js
flochtililoch: the `SHARE` namespace is still "custom" (url being on instawork.com instead of hyperview.org)

PR #62, opened 2019-04-11
Feature: collapse multiple whitespaces within text nodes.

flochtililoch:
This allows using white-spaces to format source code without affecting rendering and brings Hyperview parsing behavior closer to [HTML whitespace spec](https://www.w3.org/TR/REC-html40/struct/text.html#h-9.1).

For example:
```xml
<text>
  Velit voluptas et alias atque provident sapiente consequuntur deserunt.
  Dolorem et non error dolorem voluptate amet accusantium. Corporis rerum
  sed labore quae sed qui quis quasi. Illo pariatur sint qui. Quasi quaerat id molestias.
  Necessitatibus et ipsa quia asperiores laborum neque. Quisquam dolorem consequatur illum.
  Ut magni iusto explicabo blanditiis quasi laborum incidunt earum. Eius ut in rerum ipsam.
  Officiis dolores suscipit consequatur placeat commodi eum. Vel possimus placeat aut eos
  tempore saepe. Esse assumenda eum illo sed aut earum quia voluptatibus. Recusandae qui
  iusto corporis sed atque. Veniam et possimus praesentium. Cum molestiae non velit minus
  voluptatibus quos illo sed. Et omnis ut soluta qui inventore molestias. Dolores voluptatem
  perspiciatis exercitationem consectetur minus illum. Quos quaerat omnis aut eius eos
  dolores velit. Et quia vel ea unde eum repudiandae. Repellat et ab sed.
</text>
```

would still render as a continuous text blob despite line breaks and white spaces used for indentation.

https://app.asana.com/0/923014529014430/1116373124321410/f

### DIFF
--- a/src/services/render/index.js
+++ b/src/services/render/index.js
@@ -104,7 +104,7 @@ export const renderElement = (
   if (element.nodeType === NODE_TYPE.TEXT_NODE) {
     // Render non-empty text nodes
     if (element.nodeValue && element.nodeValue.trim().length > 0) {
-      return element.nodeValue.trim();
+      return element.nodeValue.trim().replace(/\s+/g, ' ');
     }
   }
 


adamstep:
So this behavior should be 1:1 with what happens in HTML, right?

adamstep:
I think there was also a task to properly handle leading whitespace, where we collapse it to a single space. I think the code here will still trim all leading whitespace.

flochtililoch:
> I think the code here will still trim all leading whitespace.

I initially thought we'd need this, but It appears that's also HTML's behavior - any leading and trailing spaces within a node are removed (at least from my testing in Safari).

PR #65, opened 2019-04-24
Make navigation prop optional

flochtililoch:
This allows to use the root Hyperview component in a context that is not aware of react-navigation. For now, the implementation consist in defaulting the `params` property to an object so that code accessing keys without checking existance first does not crash. This should be better addressed with added typing when refactoring the component in https://app.asana.com/0/923014529014430/933532962114370/f.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -437,15 +437,15 @@ export default class HyperScreen extends React.Component {
     });
   }
 
-  getNavigationState = () => {
-    if (this.props.navigation) {
-      return this.props.navigation.state;
+  getNavigationState = (props) => {
+    if (props.navigation) {
+      return props.navigation.state;
     }
     return { params: {} };
   }
 
   componentDidMount() {
-    const { params } = this.getNavigationState();
+    const { params } = this.getNavigationState(this.props);
     // The screen may be rendering via a navigation from another HyperScreen.
     // In this case, the url to load in the screen will be passed via navigation props.
     // Otherwise, use the entrypoint URL provided as a prop to the first HyperScreen.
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -470,10 +478,11 @@ export default class HyperScreen extends React.Component {
    * preload screen and URL to load.
    */
   componentWillReceiveProps(nextProps) {
-    const newUrl = nextProps.navigation.state.params.url;
-    const oldUrl = this.props.navigation.state.params.url;
-    const newPreloadScreen = nextProps.navigation.state.params.preloadScreen;
-    const oldPreloadScreen = this.props.navigation.state.params.preloadScreen;
+    const { params } = this.getNavigationState();
+    const newUrl = params.url;
+    const oldUrl = params.url;
```
divyanshu013: Looks like `oldUrl` and `newUrl` point to same thing. Looks like the earlier code picked them differently from the current and next props.

src/index.js
```diff
@@ -470,10 +478,11 @@ export default class HyperScreen extends React.Component {
...
+    const newPreloadScreen = params.preloadScreen;
+    const oldPreloadScreen = params.preloadScreen;
```
divyanshu013: The `oldPreloadScreen` and `newPreloadScreen` are pointing to the same `params.preloadScreen`.

src/index.js
flochtililoch: good catch

flochtililoch: addressed in f5396f3

src/index.js
flochtililoch: addressed in f5396f3

PR #50, opened 2019-03-13
Refactor hv-image and add typings

divyanshu013:
Still WIP. Needs changes to make the `<image />` element work.

PR #66, opened 2019-04-25
HyperRef: refactor

flochtililoch:
Extract `HyperRef` component into its own module. Put it under `src/core` instead of `src/components` as its usage is different from other components, represented by a node in the HAXML doc.

This is a sub-PR of #50, props to @divyanshu013 for the original work.

https://app.asana.com/0/923014529014430/1119995435707633/f


src/types.js
```diff
@@ -293,3 +294,46 @@ export type HvBehavior = {
...
+  PRESS_OUT: 'pressOut',
+  PRESS: 'press',
+  REFRESH: 'refresh',
+  SELECT: 'select',
+  VISIBLE: 'visible',
+});
+
+export type Trigger = $Values<typeof TRIGGERS>;
+
+export const PRESS_TRIGGERS = Object.freeze([
```
divyanshu013: Should we leave `freeze` on `TRIGGERS` and `PRESS_TRIGGERS`?

flochtililoch: I think we can revisit this in the entire project - enums should be immutable and types from enum should be marked as such (`$ReadOnly`)

PR #72, opened 2019-05-19
Format example xml files

adamstep:
- add script to format XML (prettier doesn't support it). It will sort attributes alphabetically, but `id` will be at the top.
- run script on all XML examples


adamstep:
@divyanshu013 I didn't do that yet because I wasn't sure how to force `tidy` to install as part of `package.json`. Any ideas how to make that happen? I don't think it's in `npm`.

flochtililoch:
the program would need to be installed as part of the circleci steps

PR #73, opened 2019-05-19
Picker field implementation

adamstep:
- Add `<picker-field>` and `<picker-item>` elements.
- `<picker-field>` renders as a `View` and `Text` RN component.
- Pressing the picker field brings up a modal with a bottom sheet.
- The bottom sheet contains a "Cancel" and "Save" button, and a picker below.
- Pressing "Cancel" dismisses the modal.
- Pressing "Save" will save the selection and populate the text in the picker field.
- Styles can be customized for the field, text, modal, and modal buttons.

### Example
```
<picker-field
    field-style="Input"
    field-text-style="Input__Text"
    modal-style="PickerModal"
    modal-text-style="PickerModal__text"
    value="1"
>
    <picker-item label="Choice 0" value="0" />
    <picker-item label="Choice 1" value="1" />
    <picker-item label="Choice 2" value="2" />
</picker-field>
```

### Screenshots
picker

### To-do
- update docs (separate PR)
- verify Android behavior, implement special-casing to avoid modal.


src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+
+    // Gets all of the <picker-item> elements. All picker item elements
+    // with a value and label are turned into options for the picker.
+    const children = Array.from(
+      element.getElementsByTagNameNS(
+        Namespaces.HYPERVIEW,
+        LOCAL_NAME.PICKER_ITEM,
+      ),
+    ).map(item => {
+      if (!item) {
```
divyanshu013: Instead of returning `null` here, perhaps it would be better to `filter` out truthy values before `map`

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+    };
+
+    // Gets all of the <picker-item> elements. All picker item elements
+    // with a value and label are turned into options for the picker.
+    const children = Array.from(
+      element.getElementsByTagNameNS(
+        Namespaces.HYPERVIEW,
+        LOCAL_NAME.PICKER_ITEM,
+      ),
+    ).map(item => {
```
divyanshu013: Flow typings

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+
+    return React.createElement(Picker, props, ...children);
+  };
+
+  /**
+   * Renders a bottom sheet with cancel/save buttons and a picker component.
+   * Uses styles defined on the <picker-field> element for the modal and buttons.
+   */
+  renderPickerModal = () => {
+    const { element, stylesheets, options } = this.props;
```
divyanshu013: Avoid destructuring

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+    const modalStyle = createStyleProp(element, stylesheets, {
```
divyanshu013: Should have flow typings for all the style variables

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+
+    return (
+      <Modal
+        animationType="slide"
+        transparent
+        visible={this.state.showPicker}
+        onRequestClose={this.cancelPickerModal}
+      >
+        <View
+          style={{
```
divyanshu013: Should be referenced from the stylesheet. We usually don't use inline styles.

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+            flex: 1,
+            justifyContent: 'flex-end',
+          }}
+        >
+          <View style={modalStyle}>
+            <View
+              style={{
+                flexDirection: 'row',
```
divyanshu013: Styles should be referenced from stylesheet

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+                justifyContent: 'space-between',
+              }}
+            >
+              <TouchableWithoutFeedback
+                onPressIn={() => this.setState({ cancelPressed: true })}
```
divyanshu013: You can create a toggle state function for `cancelPressed` that would inverse the state

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+                onPressOut={() => this.setState({ cancelPressed: false })}
+                onPress={this.cancelPickerModal}
+              >
+                <Text style={cancelTextStyle}>{cancelLabel}</Text>
+              </TouchableWithoutFeedback>
+              <TouchableWithoutFeedback
+                onPressIn={() => this.setState({ savePressed: true })}
```
divyanshu013: Here as well, perhaps a good idea to have a single toggle function

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+            </View>
+            {this.renderPicker()}
+          </View>
+        </View>
+      </Modal>
+    );
+  };
+
+  render() {
+    const { element, stylesheets, options } = this.props;
```
divyanshu013: Avoid destructuring the props

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+    if (element.getAttribute('hide') === 'true') {
+      return null;
+    }
+
+    const focused = this.state.showPicker;
+    const pressed = this.state.fieldPressed;
+    const props = createProps(element, stylesheets, {
+      ...options,
```
divyanshu013: We should probably add typings for all these variables

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+      focused,
+      pressed,
+      styleAttr: 'field-text-style',
+    });
+    const value = element.getAttribute('value');
+    const placeholderTextColor = element.getAttribute('placeholderTextColor');
+
+    if (!value && placeholderTextColor) {
+      fieldTextStyle.push({ color: placeholderTextColor });
```
divyanshu013: This could be moved where `fieldTextStyle` is initialized to avoid mutation. Usually not a good idea to mutate objects.

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    const label = value
+      ? this.getLabelForValue(value) || value
+      : element.getAttribute('placeholder') || '';
+
+    return (
+      <TouchableWithoutFeedback
+        onPressIn={() => this.setState({ fieldPressed: true })}
```
divyanshu013: We can have a single state toggler for these

src/index.js
```diff
@@ -446,7 +446,8 @@ export default class HyperScreen extends React.Component {
     const formData = new FormData();
     let formHasData = false;
 
-    ['text-area', 'text-field', 'select-single', 'select-multiple']
+    // TODO: It would be more flexible to grab any element with a name and value.
+    ['text-area', 'text-field', 'select-single', 'select-multiple', 'picker-field']
```
divyanshu013: We should probably move these to a constant (maybe to the top of the file or function) and create an enum type for these

src/services/index.js
```diff
@@ -50,6 +50,54 @@ export const getFirstTag = (document: Document, localName: LocalName) => {
   return null;
 };
 
+export const createStyleProp = (
+  element: Element,
+  stylesheets: StyleSheets,
+  options: HvComponentOptions,
+) => {
```
divyanshu013: We should add a return flow typing

src/services/index.js
```diff
@@ -50,6 +50,54 @@ export const getFirstTag = (document: Document, localName: LocalName) => {
...
+  const styleAttr = options.styleAttr || 'style';
```
divyanshu013: We should add flow typings to these variables

src/services/index.js
```diff
@@ -50,6 +50,54 @@ export const getFirstTag = (document: Document, localName: LocalName) => {
...
+  if (!element.getAttribute(styleAttr)) {
+    return [];
+  }
+
+  const styleValue = element.getAttribute(styleAttr) || '';
+  const styleIds = styleValue.split(' ');
+  let styleRules = styleIds.map(s => stylesheets.regular[s]);
```
divyanshu013: We should add a more descriptive variable name instead of `s`

src/services/index.js
```diff
@@ -50,6 +50,54 @@ export const getFirstTag = (document: Document, localName: LocalName) => {
...
+  stylesheets: StyleSheets,
+  options: HvComponentOptions,
+) => {
+  const styleAttr = options.styleAttr || 'style';
+  if (!element.getAttribute(styleAttr)) {
+    return [];
+  }
+
+  const styleValue = element.getAttribute(styleAttr) || '';
+  const styleIds = styleValue.split(' ');
```
divyanshu013: This separator `' '` should be perhaps moved to a constant in the types

src/services/index.js
```diff
@@ -50,6 +50,54 @@ export const getFirstTag = (document: Document, localName: LocalName) => {
...
+  let styleRules = styleIds.map(s => stylesheets.regular[s]);
+
+  if (options.pressed) {
+    let pressedRules = styleIds
+      .map(s => stylesheets.pressed[s])
+      .filter(r => !!r);
+    if (pressedRules.length === 0) {
+      pressedRules = [{ opacity: 0.7 }];
```
divyanshu013: We should move the opacity `0.7` to some constant for easily referencing it.

src/services/index.js
```diff
@@ -50,6 +50,54 @@ export const getFirstTag = (document: Document, localName: LocalName) => {
...
+    return [];
+  }
+
+  const styleValue = element.getAttribute(styleAttr) || '';
+  const styleIds = styleValue.split(' ');
+  let styleRules = styleIds.map(s => stylesheets.regular[s]);
+
+  if (options.pressed) {
+    let pressedRules = styleIds
+      .map(s => stylesheets.pressed[s])
```
divyanshu013: Perhaps more descriptive variable name

src/services/index.js
```diff
@@ -50,6 +50,54 @@ export const getFirstTag = (document: Document, localName: LocalName) => {
...
+      .filter(r => !!r);
```
divyanshu013: We can simplify this for filtering truthy values:

```suggestion
      .filter(Boolean);
```

src/components/hv-picker-field/index.js
```diff
@@ -0,0 +1,247 @@
...
+      styleAttr: 'field-style',
+    });
+    const fieldTextStyle = createStyleProp(element, stylesheets, {
+      ...options,
+      focused,
+      pressed,
+      styleAttr: 'field-text-style',
+    });
+    const value = element.getAttribute('value');
+    const placeholderTextColor = element.getAttribute('placeholderTextColor');
```
flochtililoch: we typically have used dash case for attributes
```suggestion
    const placeholderTextColor = element.getAttribute('placeholder-text-color');
```

adamstep: I thought about that, but we already use `placeholderTextColor` camel-cased for the `<text-field>` component. I thought it would be more consistent to keep the same attribute.

long term, I think we should support an `empty` modifier in style sheets rather than specifying the placeholder style directly as an attribute.

adamstep:
@divyanshu013 I don't think I can combine the state for `value` and `pickerValue`. We need state for the value of the picker (it's a controlled component). Since the picker selection can be canceled, we can't override the value that's displayed in the field.

I do think I can get rid of `label` in the state, since this data is derived from `value` and the element.

src/components/hv-picker-field/index.js
adamstep: @divyanshu013 I don't agree in this case, mutation is the cleanest way to express the desired outcome, and there are no side-effects outside of the function. If there's a way to express the same thing without mutation, while keeping the code easy-to-read, I'm open to it.

src/index.js
adamstep: I don't think it's really a type, just a single list of elements.

adamstep:
@divyanshu013 @flochtililoch adressed all of the comments.

Big change from earlier was to add Android support (verified on device). On Android, the picker field can be rendered inline, on iOS we need to use the modal.

PR #77, opened 2019-05-22
Add example to clear all & select all in a form

adamstep:
As a reference for @scripterkaran.

clear_select

### DIFF
--- a/examples/case_studies/form_select_clear/index.xml
+++ b/examples/case_studies/form_select_clear/index.xml
@@ -1,0 +1,217 @@
+<doc xmlns="https://hyperview.org/hyperview">
+  <screen>
+    <styles>
+      <style id="Body"
+             backgroundColor="white"
+             flex="1"
+             paddingTop="40" />
+      <style id="Select"
+             marginTop="24" />
+      <style id="Select__Separator"
+             borderTopColor="#e1e3e3"
+             borderTopWidth="1"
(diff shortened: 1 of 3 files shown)


scripterkaran:
Awesome Thanks! @adamstep

scripterkaran:
@adamstep  I think we missed a case:

1# Select any option 
2# Click Clear Selection
3# Select any option
4# Click Clear Selection -> Does not clear selection

adamstep:
@scripterkaran it should work now. I moved the hidden view above the visible screen. That way the referenced id is first.

It's not a solution I feel great about but it works. There are some alternative solutions we can consider now or for the long run:
1. rather than replacing the fragment from the doc, make a request to the server to replace the `<select-multiple>` with a filled element or cleared element.
    - Pros: will work without any changes to the Hyperview client
    - Cons: simple interaction would require a request to the backend, which can feel like overkill
2. add a `select` and `deselect` behavior, which would change the selection state of an `<option>` element. Then, pressing "clear all" would trigger a bunch of `deselect` behaviors to change the state of the form.
    - Pros: interaction is explicit, XML is clean, everything happens on client
    - Cons: requires adding and testing a new behavior in the Hyperview client

adamstep:
Closed in favor of the new `set-value` action: https://github.com/Instawork/hyperview/pull/98

PR #82, opened 2019-06-30
Bump React dependencies for Hyperview and demo app

adamstep:
Now that our apps are using RN 0.59, I'm increasing the minimum peer deps for hyperview, and I'm upgrading the demo app to Expo 33.

The motivation is to require versions of React that support the new Context API, which we want to start using for features including the forthcoming date picker.

For QA: I tested every single screen in the demo app for correct functionality.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,8 +43,8 @@
     "xmldom": "0.1.27"
   },
   "peerDependencies": {
-    "react": ">= 16.2.0",
-    "react-native": " >= 0.52.2",
+    "react": ">= 16.6.0",
+    "react-native": " >= 0.57.2",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -43,8 +43,8 @@
     "xmldom": "0.1.27"
   },
   "peerDependencies": {
-    "react": ">= 16.2.0",
-    "react-native": " >= 0.52.2",
+    "react": ">= 16.6.0",
```
divyanshu013: The context API should work with `>= 16.3.0` as well

adamstep: The `contextType` syntax was added in 16.6.0. Since all of the docs refer to it, I thought it was better to use a min version that matches the docs.

PR #83, opened 2019-07-01
Implement date-field element

adamstep:
[TAD](https://instawork.atlassian.net/wiki/spaces/EN/pages/901677061/Date+field+element)

#### Feature
Implements the `<date-field>` element for selecting dates. This works similar to the `<picker>` field:
- renders a field (view with text) on the screen. The view and text can be styled.
- tapping the field focuses it and brings up the date selector:
    - on iOS, we show a bottom sheet with the spinning date picker
    - on Android, we show the system date picking modal
- Selected dates are shown in the field, rendered in the given format.

#### Implementation details
- The date formatting (for the field) is handled via dependency injection from the client. In the demo app, we use the `moment` library to demonstrate the formatting.
- The date formatting function is passed to date fields from the root screen via a React context.

#### To-do
- [x] Android implementation
- [ ] Update website docs

#### Screenshots
iOS:
date-field

### DIFF
--- a/src/components/hv-date-field/index.js
+++ b/src/components/hv-date-field/index.js
@@ -1,0 +1,321 @@
+// @flow
+
+/**
+ * Copyright (c) Garuda Labs, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import * as Namespaces from 'hyperview/src/services/namespaces';
+import type {
+  DOMString,
+  Element,
+  HvComponentOptions,
+  StyleSheets,
+} from 'hyperview/src/types';
+import {
+  DatePickerIOS,
+  Modal,
+  Platform,
+  Text,
+  TouchableWithoutFeedback,
+  View,
+} from 'react-native';
+import type { Props, State } from './types';
+import React, { PureComponent } from 'react';
+import { createProps, createStyleProp } from 'hyperview/src/services';
+import { FormatDateContext } from 'hyperview/src';
+import { LOCAL_NAME } from 'hyperview/src/types';
+import type { Node as ReactNode } from 'react';
+import type { StyleSheet as StyleSheetType } from 'react-native/Libraries/StyleSheet/StyleSheetTypes';
+import styles from './styles';
+
+/**
+ * A date field renders a form field with ISO date fields (YYYY-MM-DD).
+ * Focusing the field brings up a system-appropriate UI for date selection:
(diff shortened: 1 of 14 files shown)


src/components/hv-date-field/index.js
```diff
@@ -0,0 +1,321 @@
...
+  toggleFieldPress = () => {
+    this.setState({ fieldPressed: !this.state.fieldPressed });
+  };
+
+  toggleCancelPress = () => {
+    this.setState({ cancelPressed: !this.state.cancelPressed });
+  };
+
+  toggleSavePress = () => {
+    this.setState({ donePressed: !this.state.donePressed });
```
divyanshu013: Here as well:

```suggestion
    this.setState(prevState => ({ donePressed: !prevState.donePressed }));
```

src/components/hv-date-field/index.js
```diff
@@ -0,0 +1,321 @@
...
+  };
+
+  /**
+   * Hides the picker and applies the chosen value to the field.
+   */
+  onModalDone = () => {
+    const element: Element = this.props.element;
+    this.setState({
+      focused: false,
+      value: this.state.pickerValue,
```
divyanshu013: Minor: since this is a derived state as well, it should use a callback pattern to update state as well.

src/index.js
```diff
@@ -126,6 +126,10 @@ export function text(element, stylesheets, onUpdate, options) {
     addHref(component, element, stylesheets, onUpdate, options);
 }
 
+// Provides the date format function to use in date fields
+// in the screen.
+export const FormatDateContext = React.createContext();
```
divyanshu013: Very minor: I think `DateFormatContext` is a better name since it provides the date format. FormatDate sounds like an action or function which formats the date.

adamstep:
@divyanshu013 thanks, I implemented all of your suggestions and added support for Android:
android

src/components/hv-date-field/index.js
```diff
@@ -0,0 +1,355 @@
...
+      // Date shown when the picker opens, must be set to a default to display.
+      pickerValue,
+      focused: false,
+      fieldPressed: false,
+      donePressed: false,
+      cancelPressed: false,
+    };
+  }
+
+  componentDidUpdate = async (prevProps: Props, prevState: State) => {
```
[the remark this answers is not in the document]
adamstep: the API for the date picker on Android is an async function. RN docs recommend using it with the `await` syntax.

PR #89, opened 2019-08-01
Blog: element dependency with events

divyanshu013:
I wrote a blog to introduce the team to using events. Any feedback or suggestions?

Relates to #81


adamstep:
@divyanshu013 looks like a good write-up overall. I'd recommend a section at the end giving some more examples of when to use this feature.
- actions on one screen affecting another screen
- decoupling elements
    - behaviors don't need to refer to other elements (don't need to know the target id)
    - UI elements can be responsive (keep their behaviors nested under the element, rather than in a different place)

Also, it would be nice to acknowledge Florent's contribution to the design.

divyanshu013:
Thanks for review @adamstep, I've updated content with the suggestions.

PR #90, opened 2019-08-13
Implement "toggle" action

adamstep:
### Summary
A new action `toggle` will hide or show a target element on the screen.

In this example, pressing the label will alternately hide and show the text below.
```xml
<text action="toggle" target="content" >Toggle</text>
<text id="content" style="Toggle">Content to be toggled</text>
```
Result:
toggle

The `toggle` action also works with conventional behavior attributes like `delay`, `show-during-load`, `hide-during-load`. Note that showing/hiding indicators will only work if `delay` is non-zero.

### Implementation details
In order to prevent adding more code to `index.js`, I've extended the custom behavior interface to support all of our behaviors. The main change is to pass 2 extra arguments:
- `getRoot()` is a function that returns a `Document`. This allows the behavior code to get the current state of the DOM for mutation.
- `updateRoot(newRoot)` is a function that allows the behavior to set the new state of the DOM.
With these changes, toggle (and other update behaviors) can be re-written as custom behaviors.

To implement toggle as a custom behavior, I pulled out some code into helper methods. Primarily, `shallowCloneToRoot()` now lives in services.

### Next Steps
Once this PR goes through a review and gets merged, I can follow up with a similar approach for `hide` and `show` actions. These will be similar other than the toggling behavior.


src/behaviors/utils.js
```diff
@@ -0,0 +1,56 @@
+// @flow
```
flochtililoch: These functions seem to be specific to the `toggle` behavior, so they would better be located under `src/behavior/toggle` directory. If some are generic enough to be reused somewhere else, we can turned into services under `src/services`

src/behaviors/toggle/index.js
```diff
@@ -0,0 +1,84 @@
+// @flow
```
flochtililoch: Other core components and behaviors are prefixed with `hv-` (directory) / `Hv` (when imported) - could you adjust this?

src/behaviors/toggle/index.js
```diff
@@ -0,0 +1,84 @@
...
+      // If there's a delay, first trigger the indicators before the toggle.
+      const newRoot = setIndicatorsBeforeLoad(
+        showIndicatorIds,
+        hideIndicatorIds,
+        getRoot(),
+      );
+      // Update the DOM to reflect the new state of the indicators.
+      updateRoot(newRoot);
+      // Wait for the delay then toggle the target.
+      later(delay).then(toggleElement);
```
flochtililoch: can we handle the failure case (catch)? it would otherwise throw a warning on the UI

src/behaviors/utils.js
```diff
@@ -0,0 +1,56 @@
...
+ * visible or hidden based on showIndicators.
+ * Returns a new Document object with the toggled indicators.
+ */
+export const toggleIndicators = (
+  ids: Array<string>,
+  showIndicators: boolean,
+  root: Document,
+): Document => {
+  let newRoot = root;
+  ids.forEach(id => {
```
flochtililoch: (not a blocker) As a general reco for new code added, I'd suggest to try avoid variables mutations when possible, i.e.

```js
return ids.reduce((newRoot, id) => {
  const indicatorElement: ?Element = root.getElementById(id);
	if (!indicatorElement) {
    return root;
  }
  indicatorElement.setAttribute('hide', showIndicators ? 'false' : 'true');
	return shallowCloneToRoot(indicatorElement);
}, root);
```

src/behaviors/utils.js
adamstep: These are generic, I plan on using them to implement `hide` and `show` actions. As we migrate other behaviors to `src/behaviors`, it will use them as well.

Do you recommend I move it to `src/services/behaviors/index.js`?

flochtililoch: yes would be a good place if they are shared among behaviors and not used by any other part of the code

src/behaviors/toggle/index.js
adamstep: `later()` just does a setTimeout, it never rejects. But I suppose I can call `toggleElement` in the catch?

adamstep:
@flochtililoch addressed most of the comments but I had two questions to clarify.

PR #92, opened 2019-08-20
Add webview from react-native-webview

reshavsingla:
Add webview from react-native-webview instead of react-native


adamstep:
@reshavsingla can you test the screens that use JS injection into the webview? We have a few in our backend, it would be good to make sure these work correctly.

PR #93, opened 2019-09-03
Fix: testID / accessibilityLabel props

flochtililoch:
Fix typo in testID prop, and add accessibilityLabel prop.

Ref:
- [testID](https://facebook.github.io/react-native/docs/0.59/view#testid)
- [accessibilityLabel](https://facebook.github.io/react-native/docs/0.59/view#accessibilitylabel)

Before:
before

After:
after

### DIFF
--- a/src/services/index.js
+++ b/src/services/index.js
@@ -125,10 +125,11 @@ export const createProps = (
       } else {
         props[attr.name] = attr.value;
 
-        // Add the id attribute as a test id
+        // Add the id attribute as a test id and accessibility label
         // (for testing automation purposes).
         if (attr.name === 'id') {
-          props.testId = attr.value;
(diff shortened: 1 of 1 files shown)


adamstep:
@flochtililoch I was thinking a bit more about this... should we be auto-setting the accessibility label? If we want to actually have good accessibility support in the future, we would want this label to be meaningful to users.

flochtililoch:
@adamstep We've been using the `accessibilityLabel` as a work-around for `testID` not being correctly set on Android / not discoverable by Appium. This is not ideal, but the alternative would be to exposing a new prop `accessibility-label` to every views, and explicitly setting it in the xml markup everywhere we currently set `id`. Or we could make a breaking change in the future when that root issue is resolved. Thoughts?

PR #96, opened 2019-10-04
Navigation tweaks: support form data for navigation and reloads

adamstep:
Implemented some missing features needed for the HV gigs list experiment. These are mostly to support interactions with the gigs filter modal:

1. Navigation actions will now serialize form data. If a navigation behavior triggers from within a `<form>`, the form data will be included as query params. Use case: apply filters to a screen via a modal:
```xml
<!-- loaded from /gigs/filters -->
<form>
  <date-field name="min-date" value="2019-05-06" />
  <text action="close" href="/gigs">Save</text>
</form>
```
When the user taps save, the modal will close and the previous screen will reload with contents from `/gigs?min-date=2019-05-06`.

2. The `reload` action can take an `href` attribute. This will cause the screen to reload with the response from the given href, rather than from the original URL of the screen. Use case: clear the filters on a screen:
```xml
<!-- loaded from /gigs?min-date=2019-05-06 -->
<view>
  <text>No gigs for the given filters</text>
  <text action="reload" href="/gigs">Clear filters</text>
</view>
```

### Other changes:
- Updated expo version of demo app to support iOS 13
- Added examples to demo app
- Updated docs on website
- Moved form data code to services for easier re-use

### DIFF
--- a/src/core/hyper-ref/index.js
+++ b/src/core/hyper-ref/index.js
@@ -9,12 +9,6 @@
  */
 
 import * as Render from 'hyperview/src/services/render';
-import { ATTRIBUTES, PRESS_TRIGGERS_PROP_NAMES } from './types';
-import type {
-  Element,
-  HvComponentOnUpdate,
-  PressTrigger,
-} from 'hyperview/src/types';
 import {
   ACTIONS,
   NAV_ACTIONS,
@@ -23,6 +17,13 @@ import {
   TRIGGERS,
   UPDATE_ACTIONS,
 } from 'hyperview/src/types';
+import { ATTRIBUTES, PRESS_TRIGGERS_PROP_NAMES } from './types';
+import type {
+  Element,
+  HvComponentOnUpdate,
+  PressTrigger,
+} from 'hyperview/src/types';
+
 import type { PressHandlers, Props, State } from './types';
 import React, { PureComponent } from 'react';
 import { RefreshControl, ScrollView, TouchableOpacity } from 'react-native';
--- a/src/core/hyper-ref/index.js
+++ b/src/core/hyper-ref/index.js
@@ -16,6 +16,7 @@ import type {
   PressTrigger,
 } from 'hyperview/src/types';
 import {
+  ACTIONS,
   NAV_ACTIONS,
   ON_EVENT_DISPATCH,
   PRESS_TRIGGERS,
@@ -106,7 +107,12 @@ export default class HyperRef extends PureComponent<Props, State> {
     const action =
       behaviorElement.getAttribute(ATTRIBUTES.ACTION) || NAV_ACTIONS.PUSH;
 
-    if (Object.values(NAV_ACTIONS).indexOf(action) >= 0) {
+    if (action === ACTIONS.RELOAD) {
+      return () => {
+        const href = behaviorElement.getAttribute(ATTRIBUTES.HREF);
+        onUpdate(href, action, element, {});
+      };
+    } else if (Object.values(NAV_ACTIONS).indexOf(action) >= 0) {
       return () => {
         const href = behaviorElement.getAttribute(ATTRIBUTES.HREF);
         const showIndicatorId = behaviorElement.getAttribute(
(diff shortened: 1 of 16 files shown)


src/core/hyper-ref/index.js
```diff
@@ -106,7 +107,12 @@ export default class HyperRef extends PureComponent<Props, State> {
     const action =
       behaviorElement.getAttribute(ATTRIBUTES.ACTION) || NAV_ACTIONS.PUSH;
 
-    if (Object.values(NAV_ACTIONS).indexOf(action) >= 0) {
+    if (action === ACTIONS.RELOAD) {
```
flochtililoch: wouldn't it be desirable to also support delay and loading indicator like `NAV_ACTIONS` case?

src/index.js
```diff
@@ -340,11 +328,17 @@ export default class HyperScreen extends React.Component {
 
   /**
    * Reload if an error occured.
+   * @param opt_href: Optional string href to use when reloading the screen. If not provided,
+   * the screen's current URL will be used.
    */
-  reload = () => {
+  reload = (opt_href) => {
```
flochtililoch: I think you need to use camel case and flow type annotations for linting to succeed

src/index.js
```diff
@@ -357,7 +351,7 @@ export default class HyperScreen extends React.Component {
       return (
         <View style={{ backgroundColor: 'white', flex: 1, alignItems: 'center', justifyContent: 'center' }}>
           <Text>An error occured</Text>
-          <TouchableOpacity onPress={this.reload}>
+          <TouchableOpacity onPress={() => this.reload()}>
```
flochtililoch: curious why did you need this change?

src/services/index.js
```diff
@@ -204,3 +205,71 @@ export const shallowCloneToRoot = (element: Element): Document => {
   parentNode.replaceChild(elementClone, element);
   return shallowCloneToRoot((parentNode: any));
 };
+
+/**
+ * Searches the parent chain from the given element until it finds an
+ * element with the given tag name. If no ancestor with the tagName is found,
+ * returns null.
+ */
+export const getAncestorByTagName = (element: Element, tagName: string) => {
```
flochtililoch: can you add an annotation for the return type?

src/services/index.js
```diff
@@ -204,3 +205,71 @@ export const shallowCloneToRoot = (element: Element): Document => {
...
+  return parentNode;
+};
+
+/**
+ * Creates a FormData object for the given element. Finds the closest form element ancestor
+ * and adds data for all inputs contained in the form. Returns null if the element has no
+ * form ancestor, or if there is no form data to send.
+ * If the given element is a form element, its form data will be returned.
+ */
+export const getFormData = (element: Element) => {
```
flochtililoch: Here as well

src/services/index.js
```diff
@@ -204,3 +205,71 @@ export const shallowCloneToRoot = (element: Element): Document => {
...
+            formElement.getElementsByTagNameNS(Namespaces.HYPERVIEW, tag),
+          ),
+        ),
+      [],
+    )
+    // Append the form data for each input
+    .forEach(input => {
+      const name = input.getAttribute('name');
+      if (
+        input.tagName === 'select-single' ||
```
flochtililoch: minor: might be good using constants defined in types.js

src/services/index.js
```diff
@@ -204,3 +205,71 @@ export const shallowCloneToRoot = (element: Element): Document => {
...
+        input.tagName === 'select-multiple'
+      ) {
+        // Add each selected option to the form data
+        Array.from(input.getElementsByTagNameNS(Namespaces.HYPERVIEW, 'option'))
```
flochtililoch: same for `option`

src/core/hyper-ref/index.js
adamstep: it would be, I'll add an Asana task for this:
https://app.asana.com/0/923014529014430/1143270339705747/f

adamstep:
@flochtililoch added flow annotations and used constants. Had to re-write some stuff due to a subtle bug: Flow's default `NodeList` type is different from the one for `xmldom`, so `Array.from()` doesn't work properly.

PR #97, opened 2019-10-06
Implement `<switch>` element for boolean inputs

adamstep:
[Asana link](https://app.asana.com/0/923014529014430/1143036055176777/f)

Implement a `<switch>` input element to use in forms. When selected, the form will serialize the value "on". When deselected, the form will serialize the value "off". This PR also adds demo examples and website documentation.

Screen Shot 2019-10-06 at 3 48 16 PM


adamstep:
@divyanshu013 that's a good question. I was generally trying to follow the conventions of the closest HTML element, `<checkbox>`. `<checkbox>` by default sends a value of `on` when checked. (It can actually send arbitrary values when checked). I thought it would be easier to integrate into server-side code if we stick to the convention. But I don't feel strongly about it, open to changing it if others feel strongly.

src/components/hv-switch/index.js
```diff
@@ -0,0 +1,63 @@
...
+    const { element } = props;
+    super(props);
+    const initialValue = element.getAttribute('value') === 'on';
+    this.state = {
+      value: initialValue,
+    };
+  }
+
+  render() {
+    const { element, stylesheets } = this.props;
```
divyanshu013: Generally we don't destructure props.

src/components/hv-switch/index.js
```diff
@@ -0,0 +1,63 @@
...
+import { createStyleProp } from 'hyperview/src/services';
+
+export default class HvSwitch extends PureComponent<Props, State> {
+  static namespaceURI = Namespaces.HYPERVIEW;
+  static localName = LOCAL_NAME.SWITCH;
+  constructor(props: Props) {
+    const { element } = props;
+    super(props);
+    const initialValue = element.getAttribute('value') === 'on';
+    this.state = {
```
divyanshu013: Should have a flow typing for `State`

PR #98, opened 2019-10-07
Implement set-value behavior for form inputs

adamstep:
[Asana task](https://app.asana.com/0/923014529014430/1143376661956577/f)

Proof of concept of a new `set-value` action that allows settings form input values via a behavior. Example syntax:
`<behavior action="set-value" target="myDateField" value="new value" />`

When this behavior triggers, it will set the text field with id `myDateField` to `new value`.

I've implemented support for this behavior in `<text-field>`, `<text-area>`, `<picker-field>`, `<date-field>`, and `<select-single>`. Need to add `<select-multiple>` and `<switch>`.

**NOTE**: In order to get this feature working, I needed to update the component state in `componentDidUpdate` by reading the attributes in the `element` prop. This is against our eslint rules which I slienced for now. To properly address this issue, the input elements would need to be re-written to not keep any local state (we could maybe keep the local state on the `element`). This is a bigger re-write that I didn't want to do now, but I added a tracking task for it [here](https://app.asana.com/0/923014529014430/1144293982403313/f).

### DIFF
--- a/src/behaviors/hv-set-value/index.js
+++ b/src/behaviors/hv-set-value/index.js
@@ -1,5 +1,6 @@
 // @flow
 
+import * as Behaviors from 'hyperview/src/services/behaviors';
 import type {
   DOMString,
   Document,
@@ -9,10 +10,8 @@ import type {
   HvUpdateRoot,
 } from 'hyperview/src/types';
 import { later, shallowCloneToRoot } from 'hyperview/src/services';
-import {
-  setIndicatorsAfterLoad,
-  setIndicatorsBeforeLoad,
-} from 'hyperview/src/services/behaviors';
+
+const ID_SEPARATOR = ' ';
 
 export default {
   action: 'set-value',
@@ -24,6 +23,7 @@ export default {
   ) => {
     const targetId: ?DOMString = element.getAttribute('target');
     if (!targetId) {
+      console.warn('[behaviors/set-value]: missing "target" attribute');
       return;
     }
 
@@ -35,10 +35,10 @@ export default {
 
     const showIndicatorIds: Array<string> = (
       element.getAttribute('show-during-load') || ''
-    ).split(' ');
+    ).split(ID_SEPARATOR);
     const hideIndicatorIds: Array<string> = (
       element.getAttribute('hide-during-load') || ''
-    ).split(' ');
+    ).split(ID_SEPARATOR);
 
     const setValue = () => {
       const doc: Document = getRoot();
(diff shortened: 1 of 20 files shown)


src/components/hv-date-field/index.js
```diff
@@ -70,6 +70,19 @@ export default class HvDateField extends PureComponent<Props, State> {
...
+      const newValue = element.getAttribute('value') || '';
+      const newDate = this.createDateFromString(newValue);
+      // NOTE(adam): We convert from date to strings for the comparison to normalize the representation.
+      if (
+        this.createStringFromDate(newDate) !==
+        this.createStringFromDate(this.state.value)
+      ) {
+        this.setState({ value: newDate, pickerValue: newDate || new Date() });
+      }
+    }
```
divyanshu013: It looks to me this section has no side effects and can be therefore moved to `getDerivedStateFromProps` as well.

src/behaviors/hv-set-value/index.js
```diff
@@ -0,0 +1,85 @@
...
+    ).split(' ');
+    const hideIndicatorIds: Array<string> = (
+      element.getAttribute('hide-during-load') || ''
+    ).split(' ');
+
+    const setValue = () => {
+      const doc: Document = getRoot();
+      const targetElement: ?Element = doc.getElementById(targetId);
+      if (!targetElement) {
+        return;
```
divyanshu013: Should we add a dev mode warning if target is not set?

src/behaviors/hv-set-value/index.js
```diff
@@ -0,0 +1,85 @@
...
+
+    const newValue: string = element.getAttribute('value') || '';
+
+    const delayAttr: string = element.getAttribute('delay') || '0';
+    const parsedDelay: number = parseInt(delayAttr, 10);
+    const delay: number = isNaN(parsedDelay) ? 0 : parsedDelay;
+
+    const showIndicatorIds: Array<string> = (
+      element.getAttribute('show-during-load') || ''
+    ).split(' ');
```
divyanshu013: There are a couple of usage of the space `' '` separator. Shall we make this and use as a constant say `EMPTY_SPACE`?

flochtililoch:
> On a side note, we can also remove the eslint rule globally since its fine to do a state update in componentDidUpdate as long as it's done conditionally.

Are you sure of this? I didn't see any mention of using conditions in [the rule definition](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-did-update-set-state.md) - shouldn't we instead solely rely on `getDerivedStateFromProps` instead of `componentDidUpdate`?

divyanshu013:
Yes, it should be better to use `getDerivedStateFromProps` for such usecases. Might be good to leave the rule as it is then since it would warn someone if they unknowingly make a mistake of updating state unconditionally.

flochtililoch:
I addressed the comments left by @divyanshu013 and fixed a bunch of existing linting warnings. @ashwinibm @olegpesok if you guys could take a look on the last few commits, mostly 1bd8a55 is where the PR comments are addressed. Thanks!

PR #99, opened 2019-10-12
Re-add src/ to published package

flochtililoch:
There's a few high frequency errors in Sentry stemming from Hyperview, but it's difficult to get a good grasp on these because the source code is compiled/obfuscated. This PR removes the `src/` directory from ignored directories when publishing the package to NPM, and will allows us to temporarily swap the entrypoint file of the module via an additional patch on the mobile repo.
Also add to the ignore list other dev-only files.

Related errors in Sentry:
https://sentry.io/organizations/instawork/issues/1266895318/?environment=release&project=1549809&query=TypeError&statsPeriod=14d
https://sentry.io/organizations/instawork/issues/1268973916/?environment=release&project=1549809&query=TypeError&statsPeriod=14d


flochtililoch:
> Should we also include source maps in published releases?

I thought about it, but I'm not sure if react native bundler would be able to use the sourcemap in the bundle's generated sourcemap - do you know if that would be the case?

divyanshu013:
There are some sourcemap related flags in the CLI but I'm not sure either how well / if it works.

flochtililoch:
> There are some sourcemap related flags in the CLI but I'm not sure either how well / if it works.

These are those I saw, but they relate to the main bundle it seems - not sure if there's a way to indicate how to include "3rd party" source maps when bundling the app, nor whether it would try to do that automagically. Might need to experiment with this.
```
--sourcemap-output [string]        File name where to store the sourcemap file for resulting bundle, ex. /tmp/groups.map
--sourcemap-sources-root [string]  Path to make sourcemap's sources entries relative to, ex. /root/dir
--sourcemap-use-absolute-path      Report SourceMapURL using its full path
```

PR #109, opened 2019-12-11
Fix: action="close" / action="back" behavior causing incorrect reload

flochtililoch:
Default `href` to `#` when attribute not set, so that navigation service does not try to load a different URL.

https://app.asana.com/0/244065166232575/1153051694506464/f


adamstep:
I guess this task would be related? https://app.asana.com/0/923014529014430/500468286004730/f

PR #110, opened 2019-12-16
HvText / HvView

flochtililoch:
Create new `HvText` and `HvView` and remove outdated code. Introduce a new static prop `localNameAliases` so that HvText can have `body` / `form` / `item` and `section-title` aliases. Add tests to `Components.getRegistry()` helper.

https://app.asana.com/0/0/933653114164339/f
https://app.asana.com/0/0/933653114164340/f

Sanity check:

| View | Text |
|------|-----|
| Screen Shot 2019-12-16 at 1 57 06 PM | Screen Shot 2019-12-16 at 1 57 10 PM |


src/components/hv-date-field/index.js
```diff
@@ -42,6 +42,7 @@ import styles from './styles';
 export default class HvDateField extends PureComponent<Props, State> {
   static namespaceURI = Namespaces.HYPERVIEW;
   static localName = LOCAL_NAME.DATE_FIELD;
+  static localNameAliases = [];
```
divyanshu013: If an element has no aliases, is there a benefit in adding this static variable? Same for the rest of the elements.

src/services/components/index.js
```diff
@@ -47,7 +51,13 @@ export const getRegistry = (
...
       [component.namespaceURI]: {
         ...registry[component.namespaceURI],
-        [component.localName]: component,
+        ...[component.localName, ...(component.localNameAliases || [])].reduce(
+          (acc, localName) => ({
+            ...acc,
+            [localName]: component,
+          }),
+          {},
+        ),
```
divyanshu013: The `getRegistry` function is quite complex. I would recommend extracting out the reducers into seperate functions and then creating tests for each instead of a single test for `getRegistry`.

src/types.js
```diff
@@ -303,6 +303,7 @@ export type HvComponentProps = {|
 
 export type HvComponent<Props> = ComponentType<Props> & {
   localName: LocalName | string,
+  localNameAliases: Array<LocalName | string>,
```
divyanshu013: When can this alias include a string other than the ones in `LocalName`?

src/components/hv-view/index.js
```diff
@@ -0,0 +1,78 @@
...
+    LOCAL_NAME.BODY,
+    LOCAL_NAME.FORM,
+    LOCAL_NAME.HEADER,
+    LOCAL_NAME.ITEM,
+    LOCAL_NAME.SECTION_TITLE,
+  ];
+  props: Props;
+
+  render() {
+    const { element, stylesheets, onUpdate, options } = this.props;
```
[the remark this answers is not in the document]
flochtililoch: the rule is not enforced in this repo - here I merely moved the existing code in its own module but haven't went all the way refactoring it

src/components/hv-view/index.js
```diff
@@ -0,0 +1,78 @@
...
+    let viewOptions = options;
+    const { skipHref } = viewOptions || {};
+    const props: InternalProps = createProps(element, stylesheets, viewOptions);
+    const scrollable = !!element.getAttribute('scroll');
+    let c = View;
```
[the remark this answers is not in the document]
flochtililoch: same here

src/components/hv-date-field/index.js
flochtililoch: I originally added it for every components, as flow would not play nice with it's absence, even when the component type definition was:

```js
export type HvComponent<Props> = ComponentType<Props> & {
  localName: LocalName | string,
  localNameAliases?: ?Array<LocalName | string>,
  namespaceURI: NamespaceURI,
};
```

can you think of a way to make this optional while keeping flow happy?

src/types.js
flochtililoch: When the component is not part of the core (i.e. our own custom components)

src/components/hv-text/index.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+
+/**
+ * Copyright (c) Garuda Labs, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
```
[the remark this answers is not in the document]
flochtililoch: I created [a task](https://app.asana.com/0/923014529014430/1154294366555337/f) if you feel strongly about this, so this can be applied to the entire codebase

src/components/hv-date-field/index.js
divyanshu013: I wasn't sure why the property was added for each component, that's why I wanted to get it clarified. So the following is totally optional since the current approach also works.

This could be possibly due to [type variance](https://flow.org/en/docs/types/interfaces/#toc-interface-property-variance-read-only-and-write-only) since interface properties are invariant by default.

Making the property co-variant should solve this.

```diff
export type HvComponent<Props> = ComponentType<Props> & {
  localName: LocalName | string,
-  localNameAliases: Array<LocalName | string>,
+  +localNameAliases?: Array<LocalName | string>,
  namespaceURI: NamespaceURI,
};
```

Alternatively since the components themselves are not interested in the `localNameAliases` we can also extract all aliases to a common service or helper.

No strong feelings for any approaches.

flochtililoch: making the property covariant does not seem to help here, the errors I'm getting are:

```sh
Error: src/components/hv-date-field/stories/index.js:15
 15: const createStory = Helpers.stories(HvDateField);
                                         ^^^^^^^^^^^ statics of `HvDateField`. This type is incompatible with the expected param type of
                                                             v
304: export type HvComponent<Props> = ComponentType<Props> & {
305:   localName: LocalName | string,
306:   +localNameAliases?: Array<LocalName | string>,
307:   namespaceURI: NamespaceURI,
308: };
     ^ object type. See: src/types.js:304
  Property `localNameAliases` is incompatible:
                                                                 v
    304: export type HvComponent<Props> = ComponentType<Props> & {
    305:   localName: LocalName | string,
    306:   +localNameAliases?: Array<LocalName | string>,
    307:   namespaceURI: NamespaceURI,
    308: };
         ^ property `localNameAliases`. Property not found in. See: src/types.js:304
     15: const createStory = Helpers.stories(HvDateField);
                                             ^^^^^^^^^^^ statics of `HvDateField`
```

divyanshu013: Seems like some issue with incorrect inference. I don't have a solution to this currently.

src/services/components/index.js
flochtililoch: addressed in 112e4a0

src/components/hv-date-field/index.js
flochtililoch: Will leave as is for the time being, we can revisit when we update flow or migrate to typescript

PR #112, opened 2019-12-19
Feature: Scroll context

flochtililoch:
Keep track of scrollable components' positions, and make them available via a React context.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -24,9 +24,11 @@
       }
     ],
     "import/no-internal-modules": [
-      "warn", {
+      "warn",
+      {
         "allow": [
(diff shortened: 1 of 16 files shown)


flochtililoch:
@divyanshu013 Thanks for the review - you make a good point, I didn't think about this potential perf issue. If you have any suggestions with regards to profiling it, let me know - I'll try to look into this.

divyanshu013:
[RN debugger](https://github.com/jhen0409/react-native-debugger/) has the profiler integrated (among things I've used). Besides that we can hook the build into our app for a more real world analysis.

PR #125, opened 2020-01-27
Chore: remove duplicate prettier configs

DanielSchaffer:
Updates linting configurations, dependencies, and runtime versions to allow using latest `eslint-plugin-prettier`

- upgraded `babel-eslint`, `eslint`, `prettier`, `pretty-quick` to match versions used by `eslint-plugin-prettier`
- added `@babel/core`, required by `eslint`
- set NodeJS runtime to `10.18.1`: `@babel/core` does not support the existing build version (8.7), and the 8.x line is EOL as of 12/31/2019. `10.18.1` is the most up to date from 10.x, which is still in extended maintenance
- temporarily disabled eslint rules triggered by existing code to avoid touching source files in this PR - to be fixed in #126


DanielSchaffer:
Looks like the Netlify build is failing due to it attempting to install `fsevents` - this should be marked as an optional dependency, and I'm wondering if updating the version of Yarn used (appears to be `1.3`, current is 1.21`) would allow it to behave correctly.

package.json
```diff
@@ -50,27 +50,27 @@
...
-    "babel-eslint": "8.2.1",
+    "babel-eslint": "10.0.1",
     "babel-plugin-module-resolver": "3.0.0",
     "babel-preset-react-native": "4.0.0",
     "chokidar": "2.1.2",
-    "eslint": "4.17.0",
-    "eslint-config-prettier": "3.3.0",
-    "eslint-plugin-instawork": "0.0.2",
-    "eslint-plugin-prettier": "3.0.0",
+    "eslint": "5.16.0",
```
divyanshu013: Are there any breaking changes in this eslint upgrade that may require changes in the custom rules in `eslint-plugin-instawork`?

.eslintrc
```diff
@@ -24,21 +26,39 @@
...
         ]
       }
-    ]
+    ],
+    "max-len": [
+      "warn",
+      {
+        "code": 100
+      }
+    ],
```
divyanshu013: I think indentation related configurations should be a part of prettierrc

.eslintrc
```diff
@@ -24,21 +26,39 @@
...
+    "import/no-cycle": "off",
+    "lines-between-class-members": "off",
+    "no-else-return": "off",
+    "no-restricted-globals": "off",
+    "prefer-destructuring": "off",
+    "react/destructuring-assignment": "off",
+    "react/jsx-sort-props": "off",
+    "react/no-access-state-in-setstate": "off",
+    "sort-keys": "off"
```
divyanshu013: We follow a convention of sorting keys alphabetically. This seems to bypass that if I'm not mistaken?

.eslintrc
```diff
@@ -24,21 +26,39 @@
...
+        "code": 100
+      }
+    ],
+    "import/no-cycle": "off",
+    "lines-between-class-members": "off",
+    "no-else-return": "off",
+    "no-restricted-globals": "off",
+    "prefer-destructuring": "off",
+    "react/destructuring-assignment": "off",
+    "react/jsx-sort-props": "off",
```
divyanshu013: We follow a convention of sorting props alphabetically. This seems to bypass that if I'm not mistaken?

.eslintrc
divyanshu013: I think some of them should be added to `eslint-plugin-instawork` instead of hyperview repo individually. Could you also consult @flochtililoch for these changes? Some of them were intentionally added in `eslint-plugin-instawork`.

package.json
DanielSchaffer: Not that I'm aware of - `5.16.0` is the same version used by `eslint-plugin-instawork`

.eslintrc
DanielSchaffer: Upgrading to `eslint-plugin-instawork@0.4.0` resulted in 200+ new linting errors. I decided to leave fixing the errors to different PR - see #126 so this one could be kept as small as possible.

.eslintrc
DanielSchaffer: Correct, though it is intended to be temporary. Upgrading to `eslint-plugin-instawork@0.4.0` resulted in 200+ new linting errors. I decided to leave fixing the errors to different PR - see #126 so this one could be kept as small as possible.

.eslintrc
DanielSchaffer: See my other comments - these are intended to be temporary.

flochtililoch:
Thanks for your first contribution to Hyperview repo, @DanielSchaffer!
As this update requires some code changes that are spread out across several PRs, it would be good to have it opened against an integration branch, so that we can continue iterating on master branch in parallel. The final integration in master can happen once the refactor is complete.
Note: each intermediate PR in your integration branch don't necessarily need to build, as long as the final PR to master does - so you won't need to temporarily disable any linting rules just to satisfy CI.

flochtililoch:
Some of the changes here overlap with #215, closing this for now

PR #129, opened 2020-01-31
Proposal: cumulative mode for select-single

adamstep:
Recently, we realized it's a bit cumbersome to implement star-ratings in Hyperview since we need to use hidden elements and swap out the star widget with different ones.

I realized that the rating widget is basically a `<select-single>`, except that it renders all `<option>` elements as selected if they precede the pressed one. So, we could implement ratings widgets if we added a `cumulative` mode to `<select-single>`:

```xml
<select-single name="rating" cumulative="true">
  <option value="1" style="bar" />
  <option value="2" style="bar" />
  <option value="3" style="bar" />
</select-single>
```
With this syntax, selecting "2" will render both "1" and "2" as selected. The form's data will only include `rating=2`.

I re-wrote the business rating case study example to mimic the UI recently implemented by @flochtililoch , with much less markup:

rating

### DIFF
--- a/src/components/hv-select-single/index.js
+++ b/src/components/hv-select-single/index.js
@@ -46,16 +46,51 @@ export default class HvSelectSingle extends PureComponent<HvComponentProps> {
     const newElement = element.cloneNode(true);
     const options = newElement.getElementsByTagNameNS(
       Namespaces.HYPERVIEW,
-      'option',
+      LOCAL_NAME.OPTION,
     );
+
+    const isCumulative: boolean =
+      newElement.getAttribute('cumulative') === 'true';
+
+    // Calculate the index of the selected index. This is
+    // used in the cumulative mode, to render the preceding
+    // options as selected.
+    let selectedIndex: ?number = null;
     for (let i = 0; i < options.length; i += 1) {
       const opt = options.item(i);
       if (opt) {
         const value = opt.getAttribute('value');
+        if (value === selectedValue) {
+          selectedIndex = i;
+        }
+      }
+    }
+
+    for (let i = 0; i < options.length; i += 1) {
+      const opt = options.item(i);
+      if (opt) {
+        const value = opt.getAttribute('value');
+        // Only the option with the selected value
+        // should be rendered as selected.
         opt.setAttribute(
           'selected',
           value === selectedValue ? 'true' : 'false',
         );
+        // If using the cumulative mode, all options
+        // should be updated with a flag indicating if
+        // they should render as selected. Options
+        // preceding the selected one should be selected,
+        // the later options should be unselected.
+        if (
+          isCumulative &&
+          selectedIndex !== null &&
+          selectedIndex !== undefined
+        ) {
+          opt.setAttribute(
+            'selected-as-cumulative',
+            i < selectedIndex ? 'true' : 'false',
+          );
+        }
       }
     }
     onUpdate('#', 'swap', element, { newElement });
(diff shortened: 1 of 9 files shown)


src/components/hv-select-single/index.js
```diff
@@ -46,16 +46,51 @@ export default class HvSelectSingle extends PureComponent<HvComponentProps> {
...
     );
+
+    const isCumulative: boolean =
+      newElement.getAttribute('cumulative') === 'true';
+
+    // Calculate the index of the selected index. This is
+    // used in the cumulative mode, to render the preceding
+    // options as selected.
+    let selectedIndex: ?number = null;
+    for (let i = 0; i < options.length; i += 1) {
```
flochtililoch: should we skip this loop when `isCumulative` is `false` ?

src/core/hyper-ref/index.js
```diff
@@ -79,13 +79,15 @@ export default class HyperRef extends PureComponent<Props, State> {
...
     });
+    let time = 0;
     onEventBehaviors.forEach(behaviorElement => {
       const handler = this.createActionHandler(
         this.props.element,
         behaviorElement,
         this.props.onUpdate,
       );
-      handler();
+      setTimeout(handler, time);
```
flochtililoch: This change is a bit obscure, what purpose does it serve?

adamstep: This is an approach the codebase uses to make sure multiple triggering behaviors get applied properly to the DOM. For example with press handlers: https://github.com/Instawork/hyperview/blob/0cf2411492cfb9324b26a6180aa554b860cb8a8b/src/core/hyper-ref/index.js#L227

We need it because executing behavior A can modify the DOM resulting in a new top-level document node. Then, when the execute B, it's holding on to DOM elements that are no longer part of the top-level document, resulting in errors.

I think there's a better way to handle this by putting all behaviors in a queue in the core update code, rather than simulating a queue using timeouts. But for now, there's some precedent for this approach in the code.

src/components/hv-select-single/index.js
adamstep: yes, I'll make that change.

src/core/hyper-ref/index.js
flochtililoch: I guess the thing that I find confusing is the increment of `time` here - are we just deferring the execution to the next tick, in which case `setTimeout(handler, 0)` would work too, or does the extra millisecond waiting between each behavior has an incidence on the execution, and if so, what would happen if any given behavior took longer than 1ms to complete its DOM changes?

adamstep: We want each behavior to execute in a separate loop so that it operates on the latest DOM root, which is why we tick up the time for each behavior.

flochtililoch: I think using `setTimeout(handler, 0)` would have the same effect - each call to `setTimeout` would schedule some code to be run on the next loop, each of these blocking the loop until the next tick, causing subsequent timers to be pushed to the next event loop. To be verified / tested, I might be wrong but I think this is how it works.

adamstep: I think you're right, let me test it out.

adamstep: it worked, updated.

adamstep:
In my testing, I found that this approach doesn't work for initial rendering. I'm going to play around with different implementations to make sure it works properly.

adamstep:
Closing to revisit the implementation.

PR #130, opened 2020-02-07
Fix: keyboard avoiding view prop

flochtililoch:
Set `scrollToInputAdditionalOffset` prop of [`react-native-keyboard-aware-scrollview`](https://github.com/wix/react-native-keyboard-aware-scrollview). We were previously setting the prop `extraScrollHeight`, which is a prop of the module [`react-native-keyboard-aware-scroll-view`](https://github.com/APSL/react-native-keyboard-aware-scroll-view#readme)

Also allow overriding the default value of `120` with the prop `scroll-to-input-offset`
https://app.asana.com/0/987416201877098/1158164359079221/f

### DIFF
--- a/docs/reference_view.md
+++ b/docs/reference_view.md
@@ -49,6 +49,7 @@ A `<view>` element can only appear anywhere within a `<screen>` element.
 - [`style`](#style)
 - [`scroll`](#scroll)
 - [`scroll-orientation`](#scroll-orientation)
+- [`scroll-to-input-offset`](#scroll-to-input-offset)
 - [`id`](#id)
 - [`hide`](#hide)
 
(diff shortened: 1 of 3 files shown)


src/components/hv-view/index.js
```diff
@@ -50,7 +50,13 @@ export default class HvView extends PureComponent<HvComponentProps> {
       const hasFields = textFields.length > 0 || textAreas.length > 0;
       c = hasFields ? KeyboardAwareScrollView : ScrollView;
       if (hasFields) {
-        props.extraScrollHeight = 32;
+        const scrollToInputAdditionalOffset = element.getAttribute(
+          'scroll-to-input-offset',
+        );
+        props.scrollToInputAdditionalOffset = scrollToInputAdditionalOffset
+          ? parseInt(scrollToInputAdditionalOffset, 10)
```
adamstep: maybe good to check that this string parses correctly in case someone provides bad input

flochtililoch: addressed in adeb20b

docs/reference_view.md
```diff
@@ -80,6 +81,14 @@ An attribute indicating whether the content in the can be scrollable. The style
 
 An attribute indicating the direction in which the view will scroll.
 
+#### `scroll-to-input-offset`
+
+| Type   | Required                |
+| ------ | ----------------------- |
+| number | No (defauls to **120**) |
+
+An attribute defining an additional scroll offset to be applied to the view, when a `<text-field>` or `<text-area>` is focused. Only work in combination with attribute `scroll` set to `"true"`.
```
DanielSchaffer:
```suggestion
An attribute defining an additional scroll offset to be applied to the view, when a `<text-field>` or `<text-area>` is focused. Only valid in combination with attribute `scroll` set to `"true"`.
```

PR #141, opened 2020-04-23
Cleanup: simplify update actions code

adamstep:
I'm working on improvements to how we handle behaviors that use the `delay` attribute. As part 1, I'm doing a simplification of the code to make my improvements easier.

This PR makes changes to `onUpdateFragment()`:
- uses the standard `setIndicatorsBeforeLoad` and `setIndicatorsAfterLoad` functions instead of custom code
- pulls out the DOM manipulations into a function called `performUpdate`.
- adds Flow types to `performUpdate`
- fixes a bug when showing or hiding multiple indicators
- fixes a bug when prepending to an empty target.


src/services/behaviors/index.js
```diff
@@ -88,3 +90,50 @@ export const setIndicatorsAfterLoad = (
...
+/**
+ * Returns a new Document object where the given action was applied to the target element
+ * with the new element.
+ */
+export const performUpdate = (
+  action: UpdateAction,
+  targetElement: Element,
+  newElement: Element,
+): Document => {
+  if (action === 'replace') {
```
flochtililoch, line 103: we have a dictionary `ACTIONS` in `src/types.js` which might be good to use in place of hardcoded action values

PR #143, opened 2020-04-28
Docs: Dispatch Event Example

ashwinibm:
- Added an example page for Event Dispatch in the website
- Added example for multiple events dispatch in demo app


docs/example_event_dispatch.md
```diff
@@ -0,0 +1,184 @@
...
+sidebar_label: Event dispatch
+---
+
+By using **trigger** [`on-event`](/docs/reference_behavior_attributes#on-event) and the **behavior action** [`dispatch-event`](/docs/reference_behavior_attributes#dispatch-event), we can communicate between different hyperview screens or between elements in the same screen.
+
+Consider a screen which shows a list of contacts along with a status.
+Another screen shows the details of a single person with an option add them to your network.
+Every time a person is added, we want the list on the first screen to show the updated status.
+
+<!-- Replace with correct gif -->
```
adamstep, line 13: is this comment still relevant? I see a few of these in the markup.

ashwinibm: I've removed them now.

docs/example_event_dispatch.md
```diff
@@ -0,0 +1,176 @@
...
+Let's start with the markup for the list of people and the default status. Clicking on an item will open a new screen with the person's information.
+
+**List Screen**
+
+```xml
+<doc xmlns="https://hyperview.org/hyperview">
+  <screen>
+    <body style="Main">
+      <list>
+        <item style="Item" href="/item1">
```
adamstep, line 28: Each item should have a `key` attribute here like in the xml file

docs/example_event_dispatch.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+**List Screen**
+
+```xml
+<doc xmlns="https://hyperview.org/hyperview">
+  <screen>
+    <body style="Main">
+      <list>
+        <item style="Item" href="/item1">
```
adamstep, line 108: add `key` here as well

flochtililoch:
A couple thoughts I had when looking at the generate docs:
- I wonder if that would make sense to leave out the styles attributes on the markup snippets, so that they are more succinct and focused on the important element / attributes that highlight the feature. We're also using BEM style on these, and readers might not be familiar with this, so these might add confusion.
- Having a gif showing the animation is great, but the permanent loop makes it somewhat difficult to spot the start and end of the animation. It would be nice to have the animation only play once, and allow the user to click the image to replay the animation.

These are more general comments and we might want to incorporate these at some point during the quarter, so not a blocker for this PR. I'm curious if @adamstep or @DanielSchaffer also have opinions with this points above.

ashwinibm:
> A couple thoughts I had when looking at the generate docs:
> 
>     * I wonder if that would make sense to leave out the styles attributes on the markup snippets, so that they are more succinct and focused on the important element / attributes that highlight the feature. We're also using BEM style on these, and readers might not be familiar with this, so these might add confusion.
> 
>     * Having a gif showing the animation is great, but the permanent loop makes it somewhat difficult to spot the start and end of the animation. It would be nice to have the animation only play once, and allow the user to click the image to replay the animation.
> 
> 
> These are more general comments and we might want to incorporate these at some point during the quarter, so not a blocker for this PR. I'm curious if @adamstep or @DanielSchaffer also have opinions with this points above.

I agree with both points. I was keeping it consistent with other examples. I've merged this PR but I can incorporate this in a future PR, if we want to do this for all examples.

PR #144, opened 2020-04-28
Hyperview XML Schema

adamstep:
Adding an XML schema for Hyperview that can be used for validation, docs generation, IDE tools, etc.

- The schema files live under the `schema/` directory. `hyperview.xsd` is the main schema, and it imports other schemas for the various namespaces we use.
- I added a command `yarn test:validate-xml` that validates the schema against the XML files under `examples/`.
- I fixed up the examples that contained easy-to-fix issues like wrong namespaces, missing attributes, etc.

However, some examples cannot be validated due to an issue in the Hyperview spec. The schema revealed a conflict between two features:
- Many elements like `<option>` use the attribute `value` to hold the input value.
- The `set-value` behavior uses the attribute `value` to update any input element.
This means that an `<option>` field that triggers a behavior to set another field is ambiguous, because `value` has two meanings.

We will need to resolve this issue in the spec before completing the schema.

### DIFF
--- a/examples/advanced_behaviors/alert/behavior.xml
+++ b/examples/advanced_behaviors/alert/behavior.xml
@@ -139,8 +139,8 @@
                     alert:title="Choose an action"
                     trigger="press"
                     xmlns:alert="https://hyperview.org/hyperview-alert"
-                    xmlns:phone="https://hyperview.org/hyperview-phone"
-                    xmlns:sms="https://hyperview.org/hyperview-sms">
+                    xmlns:phone="https://instawork.com/hyperview-phone"
+                    xmlns:sms="https://instawork.com/hyperview-sms">
             <alert:option action="phone"
                           alert:label="Call phone"
                           phone:number="(555) 555-5555" />
--- a/examples/case_studies/photos/index.xml
+++ b/examples/case_studies/photos/index.xml
@@ -168,7 +168,7 @@
         </view>
       </view>
       <section-list>
-        <section key="stories">
+        <section>
           <section-title></section-title>
           <item key="stories">
             <view scroll-orientation="horizontal"
@@ -211,7 +211,7 @@
             </view>
           </item>
         </section>
-        <section key="1">
+        <section>
           <section-title style="ImageHeader">
             <view style="ImageHeader__Left">
               <image href="/case_studies/photos/index.xml"
@@ -286,7 +286,7 @@
             </view>
           </item>
         </section>
-        <section key="2">
+        <section>
           <section-title style="ImageHeader">
             <view style="ImageHeader__Left">
               <image source="/case_studies/photos/avatars/picard.jpg"
@@ -343,7 +343,7 @@
             </view>
           </item>
         </section>
-        <section key="3">
+        <section>
           <section-title style="ImageHeader">
             <view style="ImageHeader__Left">
               <image source="/case_studies/photos/avatars/worf.jpg"
@@ -392,7 +392,7 @@
             </view>
           </item>
         </section>
-        <section key="4">
+        <section>
           <section-title style="ImageHeader">
             <view style="ImageHeader__Left">
               <image source="/case_studies/photos/avatars/deanna.png"
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "test:render": "jest test/render.test.js --forceExit",
     "test:unit": "jest --testPathPattern src",
     "test:xmlserver": "http-server -c-1 -p 8085 ./examples",
+    "test:validate-xml": "find examples/ -name '*.xml' | xargs xmllint --schema schema/hyperview.xsd --noout",
     "test": "yarn generate test && yarn test:flow && yarn test:lint && yarn test:render && yarn test:unit"
   },
   "dependencies": {
--- a/schema/alert.xsd
+++ b/schema/alert.xsd
@@ -1,0 +1,23 @@
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  attributeFormDefault="qualified"
+  elementFormDefault="qualified"
+  targetNamespace="https://hyperview.org/hyperview-alert"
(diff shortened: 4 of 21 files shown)


examples/case_studies/photos/index.xml
```diff
@@ -168,7 +168,7 @@
         </view>
       </view>
       <section-list>
-        <section key="stories">
```
flochtililoch, line 171: shouldn't these keys preserved / added to the schema? (as they generate RN warnings when missing)

schema/alert.xsd
```diff
@@ -0,0 +1,23 @@
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  attributeFormDefault="qualified"
+  elementFormDefault="qualified"
+  targetNamespace="https://hyperview.org/hyperview-alert"
+  xmlns:hv="https://hyperview.org/hyperview"
+>
+<xs:import namespace="https://hyperview.org/hyperview" schemaLocation="hyperview.xsd"/>
```
flochtililoch, line 7:
```suggestion
  <xs:import namespace="https://hyperview.org/hyperview" schemaLocation="hyperview.xsd"/>
```

schema/phone.xsd
```diff
@@ -0,0 +1,10 @@
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  attributeFormDefault="qualified"
+  elementFormDefault="qualified"
+  targetNamespace="https://instawork.com/hyperview-phone"
+>
+
```
flochtililoch, line 6:
```suggestion
```

schema/share.xsd
```diff
@@ -0,0 +1,14 @@
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  attributeFormDefault="qualified"
+  elementFormDefault="qualified"
+  targetNamespace="https://instawork.com/hyperview-share"
+>
+
```
flochtililoch, line 6:
```suggestion
```

schema/sms.xsd
```diff
@@ -0,0 +1,10 @@
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  attributeFormDefault="qualified"
+  elementFormDefault="qualified"
+  targetNamespace="https://instawork.com/hyperview-sms"
+>
+
```
flochtililoch, line 6:
```suggestion
```

package.json
```diff
@@ -36,6 +36,7 @@
     "test:render": "jest test/render.test.js --forceExit",
     "test:unit": "jest --testPathPattern src",
     "test:xmlserver": "http-server -c-1 -p 8085 ./examples",
+    "test:validate-xml": "find examples/ -name '*.xml' | xargs xmllint --schema schema/hyperview.xsd --noout",
```
flochtililoch, line 39: is this too early to chain this  to the test command?

examples/advanced_behaviors/alert/behavior.xml
```diff
@@ -139,8 +139,8 @@
                     alert:title="Choose an action"
                     trigger="press"
                     xmlns:alert="https://hyperview.org/hyperview-alert"
```
DanielSchaffer, line 141: Why are these changing from `hyperview.org` to `instawork.com`?

schema/alert.xsd
```diff
@@ -0,0 +1,23 @@
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  attributeFormDefault="qualified"
+  elementFormDefault="qualified"
+  targetNamespace="https://hyperview.org/hyperview-alert"
```
DanielSchaffer, line 4: Is there any kind of standard for defining nested/sub-namespaces? Using forward slashes (`/`) instead of dashes seems like it might be more appropriate.

schema/alert.xsd
DanielSchaffer, line 7: I think `schemaLocation` might need to be changed to a real URI before we can publish this anywhere.

examples/case_studies/photos/index.xml
adamstep: `key` is required on `<item>` elements, but I don't think it's required on `<section>`. I checked our templates in the instawork repo and we don't use keys there.

package.json
adamstep: yes, like I mentioned some of the examples don't pass yet due to a conflict in the spec.

schema/alert.xsd
adamstep: yes, once we publish we'd want to replace it with a URL. But this works locally with `xmllint` so I'm leaving it for now. I think we'll need a publish command that replaces these locations with the global ones before pushing to production.

schema/alert.xsd
adamstep: There's no standard for sub-namespaces. Technically, namespaces don't even need to be valid URLs. That became the convention in order to minimize conflicts between different groups putting out schemas.

examples/advanced_behaviors/alert/behavior.xml
adamstep: The namespace distinguishes features that are supported in the core Hyperview client vs the ones that are custom extensions included in our apps. Some features get moved from our custom extensions to the core once they mature. For example, `alert` and `share` behaviors use the `hyperview.org` namespace. Things like `intercom`, `amplitude`, `redux` will likely always use the `instawork.com` namespace.

The one outlier is `share`, we didn't change the namespace immediately since it was in wide use, but there's a task to migrate to the `hyperview.org` namespace.

schema/alert.xsd
DanielSchaffer: Thoughts on using slashes vs the dashes?

adamstep: I have no preference either way, so we can use slashes in the future. It is a bit of a pain to change the existing namespaces though while maintaining backwards compatibility with older versions of the Hyperview client that are out there

PR #145, opened 2020-04-29
Update flow-type definition for jest

flochtililoch:
Match it with actual jest version - will be needed for upcoming PR


DanielSchaffer:
Why are these source controlled? Shouldn't these just be dev dependencies?

flochtililoch:
> Why are these source controlled? Shouldn't these just be dev dependencies?

That's how flow expects it to be setup - see https://github.com/flow-typed/flow-typed

DanielSchaffer:
> That's how flow expects it to be setup

huh. alrighty!

PR #147, opened 2020-04-29
Extract XML fetching code in its own service

flochtililoch:
Goals of this code change:
- centralize code that loads remote content, currently spread out between `load` and `fetchElement`
- use async/await style instead of chained promise to ease readability and avoid variable mutations
- in a future PR, have a better handling for error responses that can containt non-xml content type (i.e. Django errors responses are text/html)
- more tests and flow coverage (src/index.js is explicitely ignored by linters as it contains too many issues)

Regarding the naming of the service, I choose `dom` even if right now it has little to do with DOM manipulation. The reasoning is to have a common place for abstractions of the `xmldom` module, and the goal would be to eventually move most of the DOM helpers currently living in [src/services/index.js (currently used a dumping ground for shared functions)](https://github.com/Instawork/hyperview/blob/master/src/services/index.js#L33-L36). I'm open to suggestions for a better name if any.

Implementation notes:
I choose a instantiable class over a simple function, so that the component props `fetch`, `onParseBefore` and `onParseAfter` could only be injected once.
I've also incorporated in this class the responsability to set the `Accept=application/xml` header, hence why it was removed from the demo app.

### DIFF
--- a/demo/screens/HyperviewScreen.js
+++ b/demo/screens/HyperviewScreen.js
@@ -58,7 +58,6 @@ export default class HyperviewScreen extends React.PureComponent {
       headers: {
         // Don't cache requests for the demo
         'Cache-Control': 'no-cache, no-store, must-revalidate',
-        Accept: 'application/xml',
         Pragma: 'no-cache',
         Expires: 0,
         ...init.headers,
--- a/src/index.js
+++ b/src/index.js
@@ -296,18 +296,16 @@ export default class HyperScreen extends React.Component {
    */
   fetchElement = async (href, method, root, formData) => {
     if (href[0] === '#') {
-      return new Promise((resolve, reject) => {
-        const element = root.getElementById(href.slice(1));
-        if (element) {
-          resolve(element.cloneNode(true));
-        }
-        reject();
-      });
+      const element = root.getElementById(href.slice(1));
+      if (element) {
+        return element.cloneNode(true);
+      }
+      throw new Error();
     }
 
     try {
       const url = UrlService.getUrlFromHref(href, this.state.url, method);
-      const document = await this.parser.load(url, formData, );
+      const document = await this.parser.load(url, formData);
       return document.documentElement;
     } catch (err) {
       this.setState({
--- a/src/index.js
+++ b/src/index.js
@@ -16,34 +16,25 @@ import * as UrlService from 'hyperview/src/services/url';
 import {
   ActivityIndicator,
   Alert,
-  Dimensions,
   Easing,
   Linking,
   Text,
   TouchableOpacity,
   View,
 } from 'react-native';
-import { DOMParser, XMLSerializer } from 'xmldom-instawork';
+import { XMLSerializer } from 'xmldom-instawork';
 import HyperRef from 'hyperview/src/core/hyper-ref';
 import Navigation, { ANCHOR_ID_SEPARATOR } from 'hyperview/src/services/navigation';
+import { Parser } from 'hyperview/src/services/dom';
 import React from 'react';
 import VisibilityDetectingView from './VisibilityDetectingView.js';
 import { createProps, getBehaviorElements, getFirstTag, later, shallowCloneToRoot, getFormData, getElementByTimeoutId, removeTimeoutId, setTimeoutId } from 'hyperview/src/services';
-import { version } from '../package.json';
 import { ACTIONS, FORM_NAMES, NAV_ACTIONS, ON_EVENT_DISPATCH, UPDATE_ACTIONS } from 'hyperview/src/types';
 import urlParse from 'url-parse';
 
 
 // Shared instance, used in dev mode only
 const devXMLSerializer: ?XMLSerializer = __DEV__ ? new XMLSerializer() : null;
-
-function getHyperviewHeaders() {
-  const { width, height } = Dimensions.get('window');
-  return {
-    'X-Hyperview-Version': version,
-    'X-Hyperview-Dimensions': `${width}w ${height}h`,
-  };
-}
 
 // Provides the date format function to use in date fields
 // in the screen.
@@ -61,18 +52,14 @@ export default class HyperScreen extends React.Component {
 
     this.onUpdate = this.onUpdate.bind(this);
     this.reload = this.reload.bind(this);
-    this.parseError = this.parseError.bind(this);
 
     this.updateActions = ['replace', 'replace-inner', 'append', 'prepend'];
+    this.parser = new Parser(
+      this.props.fetch,
+      this.props.onParseBefore,
+      this.props.onParseAfter
+    );
 
-    this.parser = new DOMParser({
-      locator: {},
-      errorHandler: {
-        warning: this.parseError,
-        error: this.parseError,
-        fatalError: this.parseError,
-      },
-    });
     this.needsLoad = false;
     this.state = {
       styles: null,
@@ -88,17 +75,6 @@ export default class HyperScreen extends React.Component {
       this.state.doc,
       this.getNavigation(),
     )
-  }
-
-  /**
-   * Callback for parser errors. Logs error to console and shows error screen.
-   */
-  parseError(e) {
-    console.error(e);
-    this.setState({
-      doc: null,
-      error: true,
-    });
   }
 
   getNavigationState = (props) => {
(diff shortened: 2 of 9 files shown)


adamstep:
In terms of creating and naming the service:

I agree that we should have a common place for all of the functions related to DOM manipulation. But that doesn't seem completely related to making and processing requests. Maybe this service should be called `fetch`?

src/services/dom/index.js
```diff
@@ -0,0 +1,88 @@
...
+    this.fetch = fetch;
+    this.onBeforeParse = onBeforeParse;
+    this.onAfterParse = onAfterParse;
+  }
+
+  load = async (
+    baseUrl: string,
+    data: ?FormData,
+    method: ?HttpMethod = HTTP_METHODS.GET,
+  ): any => {
```
adamstep, line 61: Can the return type be well-defined here?

flochtililoch:
> In terms of creating and naming the service:
> 
> I agree that we should have a common place for all of the functions related to DOM manipulation. But that doesn't seem completely related to making and processing requests. Maybe this service should be called `fetch`?

Are you suggesting the DOM manipulation functions should not be in the same module that generates the DOM object? I can rename the module `fetch`, what about the class `Parser`?

src/services/dom/types.js
```diff
@@ -0,0 +1,37 @@
...
+export const HTTP_HEADERS = {
+  ACCEPT: 'Accept',
+  CONTENT_TYPE: 'Content-Type',
+  X_HYPERVIEW_VERSION: 'X-Hyperview-Version',
+  X_HYPERVIEW_DIMENSIONS: 'X-Hyperview-Dimensions',
+};
+
+export const HTTP_METHODS = {
+  GET: 'GET',
+  POST: 'POST',
```
adamstep, line 20: We technically also support `put`, `patch`, and `delete`, even though we don't use them from the IW backend:  https://hyperview.org/docs/reference_behavior_attributes#verb

Can you add some tests for those methods as well?

adamstep:
> > In terms of creating and naming the service:
> > I agree that we should have a common place for all of the functions related to DOM manipulation. But that doesn't seem completely related to making and processing requests. Maybe this service should be called `fetch`?
> 
> Are you suggesting the DOM manipulation functions should not be in the same module than the one that generates the DOM object? I can rename the module `fetch`, what about the class `Parser`?

Parsing makes sense in the DOM service since it's creating the DOM

src/index.js
```diff
@@ -16,35 +16,26 @@ import * as UrlService from 'hyperview/src/services/url';
 import {
```
DanielSchaffer, line 16: A general note on file naming / structure:

`index.js` should be reserved for "barrel" modules. "Barrel" modules are files, similar to `__init__.py` in Python, that are used specifically for defining the public API of a particular "sub-module" or section of the package. The barrel module can be used in a couple different ways depending on what the intention is.

For smaller / less complex packages, it can be used just as a way to break up the definition of the public API:

Given the following folder structure:
```
[root]
src
  dom
    index.js
    parser.js
  other-section-a
    ...
    index.js
  other-section-b
    ...
    index.js
  index.js
```

```javascript
// ./src/dom/parser.js
const parser = {...}
export class Parser { ... }

// ./src/dom/index.js
export * from './parser'

// ./src/index.js - root entry point for the package
export * from './dom'
export * from './other-section-a'
export * from './other-section-b'
```
In the above example, the entire public API is available from a single entry point - so the consumer would only ever `import {...} from 'hyperview'`. 

Alternatively, a slightly different pattern can be used to define sub-"namespaces" within a package by effectively defining multiple endpoints: 

```
[root]
dom
  src
    parser.js
  index.js
src
  other-section-b
    ...
    index.js
  other-section-a
    ...
    index.js
  index.js
```

```javascript
// ./dom/src/parser.js
const parser = {...}
export class Parser { ... }

// ./dom/index.js
export * from './src/parser'

// ./src/index.js - root entry point for the package
export * from './other-section-b'
export * from './other-section-b'
```

In this example, the root entry point does _not_ export anything from `dom` itself, instead leaving `dom` to create its own endpoint. In this case, the consumer would `import { Parser } from 'hyperview/dom'` and, `import { everything, else } from 'hyperview'`

src/index.js
```diff
@@ -195,69 +171,48 @@ export default class HyperScreen extends React.Component {
   /**
    * Performs a full load of the screen.
    */
-  load = () => {
+  load = async () => {
```
DanielSchaffer, line 174: Promises to `async` ... YES PLEASE.

src/index.js
```diff
@@ -195,69 +171,48 @@ export default class HyperScreen extends React.Component {
...
-          error: true,
-        });
-        throw reason;
+      const url = this.state.url;
+      const doc = await this.parser.load(url);
+
+      // Make sure the XML has the required elements: <doc>, <screen>, <body>.
+      const docElement = getFirstTag(doc, 'doc');
+      if (!docElement) {
+        throw new Error(`No <doc> tag found in the response from ${url}.`);
```
DanielSchaffer, line 188: You'd created specific error types elsewhere - why not for these? Seems like these could be abstracted into some kind of `ParserError` that takes `elementName` and `sourceUrl` params. Also, maybe this validation functionality could be integrated directly into `getFirstTag`?

src/index.js
```diff
@@ -195,69 +171,48 @@ export default class HyperScreen extends React.Component {
...
-      later(parseInt(delay, 10)).then(fetchPromise);
-    } else {
-      fetchPromise();
+    } catch (err) {
+      this.setState({
+        doc: null,
+        styles: null,
+        error: true,
+      });
+      console.error(err.message);
```
DanielSchaffer, line 215: Personally, I consider it bad form for a package to write to the console on its own. If there's a logging abstraction, we should use that - but otherwise, I think we should either leave logging to the consumer, or throw an error they have to handle.

Also, is there a reason we can't include the actual error in the state update here? Maybe just set `error: err` instead of `true`?

src/index.js
```diff
@@ -339,8 +294,7 @@ export default class HyperScreen extends React.Component {
    *   used to render the screen.
    * Returns a promise that resolves to a DOM element.
    */
-  fetchElement = (href, verb, root, formData) => {
-    verb = verb || 'GET';
+  fetchElement = async (href, method, root, formData) => {
     if (href[0] === '#') {
       return new Promise((resolve, reject) => {
```
DanielSchaffer, line 299: This should not be a promise, especially if you've converted the wrapping function to `async`. It doesn't look it needs to be a promise at all actually since there's nothing asynchronous happening.

src/index.js
```diff
@@ -351,31 +305,18 @@ export default class HyperScreen extends React.Component {
...
-          this.props.onParseBefore(url);
-        }
-        const parsed = this.parser.parseFromString(responseText).documentElement;
-        if (typeof this.props.onParseAfter === 'function') {
-          this.props.onParseAfter(url);
-        }
-        return parsed;
+    try {
+      const url = UrlService.getUrlFromHref(href, this.state.url, method);
+      const document = await this.parser.load(url, formData, );
```
DanielSchaffer, line 310:
```suggestion
      const document = await this.parser.load(url, formData);
```

src/index.js
```diff
@@ -351,31 +305,18 @@ export default class HyperScreen extends React.Component {
...
+      return document.documentElement;
+    } catch (err) {
+      this.setState({
+        doc: null,
+        styles: null,
+        error: true,
       });
+      console.error(err.message);
```
DanielSchaffer, line 318: Same comment as above RE console logging and error value.

src/services/dom/types.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Copyright (c) Garuda Labs, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+export const HTTP_HEADERS = {
```
DanielSchaffer, line 11: The common convention for object members in JavaScript is to use camelCase. Another thing to consider is that these are really enums - I know we're not using TypeScript right now, but the convention there for enum is to use PascalCase (like classes with static members)

src/index.js
flochtililoch: I deliberately left these as they are (they already exist even if the diff shows them as new code) as this seemed out of scope for the current refactor. As mentioned in the PR description, this file `src/index.js` is explicitly ignored by our linting tools as there are too many existing issues, so I'd rather limit as much as possible new code added to this file.

src/index.js
flochtililoch: a couple thing with this regards:
- `console.error` in react native triggers the "redbox" alert on the device, only for debug builds - it's noop for prod builds.
- it expects a string in order to do proper rendering of the error - just passing the error would effectively trigger a redbox with the error constructor name only and no details

demo/screens/HyperviewScreen.js
```diff
@@ -58,7 +58,6 @@ export default class HyperviewScreen extends React.PureComponent {
       headers: {
         // Don't cache requests for the demo
         'Cache-Control': 'no-cache, no-store, must-revalidate',
-        Accept: 'application/xml',
```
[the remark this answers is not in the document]
flochtililoch: See the PR description about this.

src/index.js
flochtililoch: I agree with most of what's mentioned there. Additionally, we use `import * ` style when possible, which allows for better namespacing of the module public API in the module consumers. i.e. 
```js

import * as Hyperview;

...

Hyperview.Events.dispatch('foo');

```

src/services/dom/types.js
flochtililoch: Good to know 👍 
We use the uppercase convention in the mobile project, and this naturally got applied in Hyperview too, but I'm open to adopting a more widespread accepted convention in Hyperview.

src/services/dom/types.js
flochtililoch: As discussed on Slack, the use case for PUT/PATCH/DELETE does not exist with Hyperview, so we agreed to remove its support and update the doc. Done in 5619e0d

src/services/dom/index.js
flochtililoch: addressed in 5619e0d

src/index.js
flochtililoch: addressed in 746f667

src/index.js
flochtililoch: addressed in 746f667

src/index.js
DanielSchaffer: Unless I'm misunderstanding how/where this is getting used, `import * ` is bad form for a browser JS package. It effectively prohibits any tree-shaking of the source, because it forces ALL modules to be loaded and included in the consumer's application bundle.

Bundlers like Webpack (combined with tools like Terser or Uglify) have the ability to remove unused code by inspecting the dependency tree and removing modules that are never accessed. This includes removing unused exports from a single file with multiple exports. However, this only works so long as Webpack can determine that a module is never used - and if modules are imported using `*`, there's nothing to tell Webpack what is actually getting used, so nothing can be "shaken".

more info here: https://webpack.js.org/guides/tree-shaking/

demo/screens/HyperviewScreen.js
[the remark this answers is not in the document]
DanielSchaffer: My bad, I'd skipped right over that.

src/index.js
flochtililoch: Fair warnings and good to learn about this. It appears that metro (the React Native bundler) [does not support tree shaking nor would support it anytime soon](https://github.com/facebook/metro/issues/227#issuecomment-583358386)

DanielSchaffer: Oh. Well that's a bummer :(

PR #151, opened 2020-04-30
Allow selectable="true" attribute

flochtililoch:
This is to be used on `<text>` elements and allow user selection on long press.

Relates to https://app.asana.com/0/518926233438259/1155044724972251/f


adamstep:
Can you update the docs/examples/schema as well?

flochtililoch:
> Can you update the docs/examples/schema as well?

Addressed in f7f8670

PR #153, opened 2020-05-04
Update react-native-webview to latest

flochtililoch:
- Update `react-native-webview` to `9.4.0`
- Set `onMessage` to no-op to preserve `injectedJavaScript` functionality (as [not-so-well documented here](https://github.com/react-native-community/react-native-webview/issues/1311))


src/components/hv-web-view/index.js
```diff
@@ -33,6 +35,7 @@ export default class HvWebView extends PureComponent<HvComponentProps> {
     return (
       <WebView
         injectedJavaScript={injectedJavaScript}
+        onMessage={NOOP}
```
adamstep, line 38: Can you link to reason here in the code?

PR #157, opened 2020-05-07
chore: add yarn resolution for fsevents

DanielSchaffer:
### Summary
- Make sure Yarn uses the latest version of `fsevents` to ensure compatibility with NodeJS >= 12
- `fsevents` is an optional dependency used to improve functionality for applications running on MacOS - it will not affect anything about the built package.

### Background
The `fsevents` library is used to add extra functionality working the filesystem on MacOS, and it's included as an optional dependency for various utility libraries. Since it's only for MacOS, it can't be installed on Linux, but since it's optional, Yarn is supposed to just show an info log and move on. However, older versions of fsevents are not compatible with NodeJS 12.x+, and the install fails. We can work around this with Yarn by using a resolution to force it to install the fixed version of fsevents, since upgrading the dependencies that use it is a much more involved level of effort. However number 2: Yarn needs to be run with `--ignore-optional` in order for it to not fail with an optional dependency that has a Yarn resolution.

_**Update:**_
I've updated the Netflify setup to use NodeJS 12 and Yarn 1.22.4. According to their docs, Netlify includes the `--ignore-optional` flag by default, but it's still failing on the incompatible module error, even when I explicitly add the --ignore-optional to their `YARN_FLAGS` env var.

I've tried installing all the dependencies from Hyperview that use `fsevents` on an Ubuntu box of mine, using the same version of Yarn (1.22.4) and NodeJS (12.16.3), and it works fine there so long as I use the `--ignore-optional` flag. Netlify just keeps failing either way, even after clearing its cache.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
       - image: instawork/hyperview-ci:1.0.0
     steps:
       - checkout
-      - run: yarn
+      - run: yarn --ignore-optional
       - run: yarn test:generators
       - run: yarn test:flow
       - run: yarn test:lint
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,7 @@ jobs:
       - image: instawork/hyperview-ci:1.0.0
     steps:
       - checkout
+      - run: curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.22.4
       - run: yarn --ignore-optional
       - run: yarn test:generators
       - run: yarn test:flow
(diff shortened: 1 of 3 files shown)


DanielSchaffer:
Note: This currently held up because Netlify installs yarn packages without the `--ignore-optional` flag, which is required when using a resolution to ensure that failing optional dependencies are ignored.

flochtililoch:
This issue was due to a misconfiguration in netlify, where our root package.json file was influencing what was installed by netlify. Settings changed below:

| Before | After |
|--------|------|
| Screen Shot 2021-01-12 at 3 49 44 PM | Screen Shot 2021-01-12 at 3 55 35 PM |

PR #163, opened 2020-05-19
Fix: Doc state

flochtililoch:
Add `doc` instance property to `HyperviewScreen`, and update it at the same time as `state.doc`. When accessing the DOM for read/write operation, use `doc` / use `state.doc` when rendering. Also simplify `Navigation` constructor.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -68,13 +68,32 @@ export default class HyperScreen extends React.Component {
       error: false,
     };
 
+    // <HACK>
+    // In addition to storing the document on the react state, we keep a reference to it
+    // on the instance. When performing batched updates on the DOM, we need to ensure every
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -68,13 +68,32 @@ export default class HyperScreen extends React.Component {
...
     this.componentRegistry = Components.getRegistry(this.props.components);
-    this.navigation = new Navigation(
-      this.state.url,
-      this.state.doc,
-      this.getNavigation(),
-    )
+    this.navigation = new Navigation(props.entrypointUrl, this.getNavigation());
+  }
+
+  setDoc = (doc) => {
```
adamstep, line 94: Is this still needed?

flochtililoch: good catch, I'll remove it

PR #165, opened 2020-05-20
Handle HTML server errors

flochtililoch:
**Open for early feedback**

Add support for `text/html` server error responses.

- changes `state.error` type from `boolean` to `Error`, and keep reference of thrown errors
- separate out `Dom.load` in `Dom.loadDocument` and `Dom.loadElement`
- add XML validation for both methods above
- add ability to view details when error is instance of `ServerError`

https://app.asana.com/0/1173541402094433/1173541402094440/f

hyperview-django-error mov


adamstep:
Are there any limits to the size of the data URI? If the server returns a lot of HTML, I'm wondering if this could possibly break (other than out-of-memory)

flochtililoch:
[There are probably some limits on data URI size](https://stackoverflow.com/a/41755526/237637), so there is a chance this could break (though this would be limited to the __DEV__ environment). Using data URI seemed the shortest path to getting this feature working without an heavy refactor in the current handling of navigation, which currently is tightly coupled with the notion of URI.

adamstep:
OK, seems reasonable.

adamstep:
Also, it looks like this only works when loading new screens, right? It won't work for server errors when updating partials on a screen. Maybe a good option for partials is to show a system alert with an option to reload, view details, or dismiss

flochtililoch:
> Maybe a good option for partials is to show a system alert with an option to reload, view details, or dismiss

I think the `<LoadError>` component should still take over the screen as the component state `error` would be populated. I haven't tested it but will do a round of QA around this before merging

flochtililoch:
@adamstep I verified, the error screen also takes over when loading/updating a partial

PR #169, opened 2020-06-03
Fix: btoa missing crash

flochtililoch:
Add check to ensure `btoa` is defined on the global scope. Default to rendering alternate text suggesting to enable debug mode to see more details.
Given that the HTML error response parsing is already only enabled during `__DEV__` mode, I preferred only allowing the feature to work when debug mode is also enabled, instead of adding a 3rd party library to perform the conversion of the raw HTML to base 64.

### DIFF
--- a/src/core/components/load-error/index.js
+++ b/src/core/components/load-error/index.js
@@ -20,7 +20,12 @@ export default class LoadError extends PureComponent<Props> {
 
   getDetailsUri = () => {
     const error = this.props.error;
-    if (!__DEV__ || !error || !(error instanceof Dom.ServerError)) {
+    if (
+      !__DEV__ ||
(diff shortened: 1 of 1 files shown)


adamstep:
I guess this is a decent fix for now, but I try to avoid running the debugger most of the time because I find it slows down my machine quite a bit. Ideally, I'd like to find a way to see the error without requiring the debugger.

One idea to avoid `btoa`: What if we support HTML content in the `<web-view>` element? Then we wouldn't need to use base64 encoding to load a separate page
```xml
<view>
  <text>Server error</text>
  <web-view>
  <![CDATA[
    HTML ERROR RESPONSE ADDED HERE
  ]]>
  </web-view>
</view>
```

PR #172, opened 2020-06-18
Webview: Dispatch Hyperview event

flochtililoch:
Allow dispatching Hyperview events from a webview

webview-dispatch-event mov


ashwinibm:
Great to have this! 💯
Would be useful for the all the quiz stories that have web views.

PR #176, opened 2020-06-30
Fix date picker

flochtililoch:
Generate dates in local timezone instead of UTC, so that formatting remains consistent. Also update demo app Hyperview version.

https://app.asana.com/0/244065166232575/1182609012386243/f

Tested by manually setting local TZ to IST.
| Before | After |
|-------|-------|
| date-picker-before mov | date-picker-after mov |

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -9,7 +9,7 @@
   },
   "dependencies": {
     "expo": "~37.0.3",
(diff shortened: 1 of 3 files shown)


adamstep:
Thanks for tracking this down!

PR #181, opened 2020-08-26
Set test props for <picker-field> on Android

adamstep:
Asana: https://app.asana.com/0/923014529014430/1190686856226427/f

On iOS, we do some custom rendering for the `<picker-field>` input. The input was rendered as a `View` component with the appropriate `id`, `testID`, and `accessibilityLabel` props.

On Android, we were using the native `Picker` directly and weren't propagating props. I created a method `createTestProps` to just set the test-related props on the Android component.

### DIFF
--- a/src/services/index.js
+++ b/src/services/index.js
@@ -110,7 +110,7 @@ export const createStyleProp = (
  * Sets the element's id attribute as a test id and accessibility label
  * (for testing automation purposes).
  */
-export const createTestProps = (element: Element): {} => {
+export const createTestProps = (element: Element): {testID?: string, accessibilityLabel?: string} => {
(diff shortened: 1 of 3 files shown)


src/services/index.js
```diff
@@ -105,6 +106,22 @@ export const createStyleProp = (
   return styleRules;
 };
 
+/**
+ * Sets the element's id attribute as a test id and accessibility label
+ * (for testing automation purposes).
+ */
+export const createTestProps = (element: Element): {} => {
```
flochtililoch, line 113:
```suggestion
export const createTestProps = (element: Element): {testID?: string, accessibilityLabel?: string} => {
```

PR #182, opened 2020-08-26
POC: progress bar component

adamstep:
Asana: https://app.asana.com/0/1184657605951713/1185457480046931/f

Prototyping a simple `<progress-bar>` component.
* You can apply styling with `style` and `bar-style`:
* The value of the progress bar is set with the `percent` attribute.

Screen Shot 2020-08-26 at 12 12 19 PM

Animating the progress bar can be done in two ways:
1. To animate the bar on initial load, you add a `start-percent` prop. When the component displays on screen, the bar will animate from `start-percent` to `percent`:

2. To animate the bar changing values based on interactions, you can `replace` the XML of the progress bar with one that has a different value of `percent`. The mounted React component will animate between the old value and new value:

### DIFF
--- a/src/components/hv-progress-bar/index.js
+++ b/src/components/hv-progress-bar/index.js
@@ -1,0 +1,116 @@
+// @flow
+
+/**
+ * Copyright (c) Garuda Labs, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import * as Namespaces from 'hyperview/src/services/namespaces';
+import React, { PureComponent } from 'react';
+import { createProps } from 'hyperview/src/services';
+import type { HvComponentProps } from 'hyperview/src/types';
+import { Animated, View } from 'react-native';
+import { LOCAL_NAME } from 'hyperview/src/types';
+
+export default class HvImage extends PureComponent<HvComponentProps> {
+  static namespaceURI = Namespaces.HYPERVIEW;
+  static localName = LOCAL_NAME.PROGRESS_BAR;
+  static localNameAliases = [];
+  props: HvComponentProps;
+
+  constructor(props) {
+    super(props);
+    const element = this.props.element;
+    const startPercent: ?number = this.getStartPercent(element);
+    const initialPercent: number =
+      startPercent === null ? this.getPercent(element) || 0 : startPercent;
+    this.state = {
+      barPercent: new Animated.Value(initialPercent),
+    };
+  }
+
+  getStartPercent = (element): ?number => {
+    const startPercent: ?DOMString = element.getAttribute('start-percent');
+    if (startPercent === null || startPercent === '') {
+      return null;
+    }
+    return parseFloat(startPercent);
+  };
+
+  getPercent = (element): ?number => {
+    const percent: ?DOMString = element.getAttribute('percent');
+    if (percent === null || percent === '') {
+      return null;
+    }
+    return parseFloat(percent);
+  };
+
+  getMaxDuration = (element): ?number => {
+    const duration: ?DOMString = element.getAttribute('max-duration');
+    if (duration === null || duration === '') {
+      return 2000;
+    }
+    return parseInt(duration, 10);
+  };
+
+  componentDidMount() {
+    const element = this.props.element;
+    const startPercent: ?number = this.getStartPercent(element);
+    const percent: ?number = this.getPercent(element);
+    const maxDuration: number = this.getMaxDuration(element);
+    const duration: numer =
+      (Math.abs(startPercent - percent) / 100) * maxDuration;
+
+    if (startPercent !== null && percent !== null) {
+      Animated.timing(this.state.barPercent, {
+        toValue: percent,
+        duration,
+      }).start();
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    const prevElement = prevProps.element;
+    const prevPercent = this.getPercent(prevElement);
+
+    const element = this.props.element;
+    const percent = this.getPercent(element);
+    const maxDuration: number = this.getMaxDuration(element);
+    const duration: numer =
+      (Math.abs(prevPercent - percent) / 100) * maxDuration;
+
+    if (prevPercent !== percent) {
+      Animated.timing(this.state.barPercent, {
+        toValue: percent,
+        duration,
+      }).start();
+    }
+  }
+
+  render() {
+    const { element, stylesheets, onUpdate, options } = this.props;
+
+    const barWidthStyle = {
+      width: this.state.barPercent.interpolate({
+        inputRange: [0, 100],
+        outputRange: ['0%', '100%'],
+      }),
+    };
+    const props = createProps(element, stylesheets, options);
+    const barProps = createProps(element, stylesheets, {
+      ...options,
+      styleAttr: 'bar-style',
+    });
+    barProps.style = [...barProps.style, barWidthStyle];
+
+    const component = (
+      <View {...props}>
+        <Animated.View {...barProps} />
+      </View>
+    );
+    return component;
+  }
+}
--- a/src/components/hv-progress-bar/index.js
+++ b/src/components/hv-progress-bar/index.js
@@ -79,7 +79,7 @@ export default class HvImage extends PureComponent<HvComponentProps> {
     const element = this.props.element;
     const percent = this.getPercent(element);
     const maxDuration: number = this.getMaxDuration(element);
-    const duration: numer =
+    const duration: number =
       (Math.abs(prevPercent - percent) / 100) * maxDuration;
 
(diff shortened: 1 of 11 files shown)


src/components/hv-progress-bar/index.js
```diff
@@ -0,0 +1,116 @@
...
+    const element = this.props.element;
+    const startPercent: ?number = this.getStartPercent(element);
+    const initialPercent: number =
+      startPercent === null ? this.getPercent(element) || 0 : startPercent;
+    this.state = {
+      barPercent: new Animated.Value(initialPercent),
+    };
+  }
+
+  getStartPercent = (element): ?number => {
```
flochtililoch, line 35: These questions apply to `getStartPercent`, `getPercent` and `getMaxDuration`:
- can `element` be flow-typed?
- shouldn't the if statement be `if (!startPercent)` ? Are we omitting `undefined` value for a specific reason?

src/components/hv-progress-bar/index.js
```diff
@@ -0,0 +1,116 @@
...
+
+  componentDidUpdate(prevProps) {
+    const prevElement = prevProps.element;
+    const prevPercent = this.getPercent(prevElement);
+
+    const element = this.props.element;
+    const percent = this.getPercent(element);
+    const maxDuration: number = this.getMaxDuration(element);
+    const duration: numer =
+      (Math.abs(prevPercent - percent) / 100) * maxDuration;
```
flochtililoch, line 83: `prevPercent` / `percent` can be `null` (from `getPercent`), should we instead default them to `0`?

src/components/hv-progress-bar/index.js
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  componentDidUpdate(prevProps) {
+    const prevElement = prevProps.element;
+    const prevPercent = this.getPercent(prevElement);
+
+    const element = this.props.element;
+    const percent = this.getPercent(element);
+    const maxDuration: number = this.getMaxDuration(element);
+    const duration: numer =
```
flochtililoch, line 82:
```suggestion
    const duration: number =
```

src/components/hv-progress-bar/index.js
adamstep: Yes, I will add full Flow types in the final PR, I didn't focus on it too much yet in case we wanted to go in a different direction. Same for handling falsy values.

adamstep:
@flochtililoch at a high level, do you think the API and interactions of this component make sense? Is it useful enough to be part of the Hyperview core, or should it be a custom component of our app? I think it's generic enough that it could go here, but let me know your thoughts.

flochtililoch:
Regarding the API:

```
start-percent="10" percent="70"
```
It's a little strange to me that we have two separate attributes to set what the "progress" is. Ideally, we would only have a single attribute, that the component can persist on mount so that it knows how to animate any change to it. I also think `progress` terminology instead of `percent` would make more sense. If we need to confer that percent is the unit, we could name it `progress-pct`

```
max-duration="15000"
```
naming is also a bit odd here, I think `animation-duration` would better represent what this is? Same, if we want to include the unit, then `animation-duration-ms` may be appropriate.

Regarding the presence of this component as part of the core, I would suggest to keep the core lean and perhaps at some point provide a UI library that could be installed alongside the core. Most core components currently abstract basic react native components or provide core/essential features of Hyperview, and I'm not sure if this animated progress bar can be considered as such.
Note: we have two internal components in the mobile app that render these kind of UI, for which some code might be made reusable? ([progress-bar](https://github.com/Instawork/mobile/tree/master/core-mobile/src/components/progress-bar) and [progress-circle](https://github.com/Instawork/mobile/tree/master/core-mobile/src/components/progress-circle))

DanielSchaffer:
This looks great so far!

+1 to everything @flochtililoch said RE naming and inclusion in core

src/components/hv-progress-bar/index.js
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+import * as Namespaces from 'hyperview/src/services/namespaces';
+import React, { PureComponent } from 'react';
+import { createProps } from 'hyperview/src/services';
+import type { HvComponentProps } from 'hyperview/src/types';
+import { Animated, View } from 'react-native';
+import { LOCAL_NAME } from 'hyperview/src/types';
+
+export default class HvImage extends PureComponent<HvComponentProps> {
```
ashwinibm, line 18:
```suggestion
export default class HvProgressBar extends PureComponent<HvComponentProps> {
```

ashwinibm:
Looks good! I tried it out with the quiz screen too and the animations work well.

adamstep:
> It's a little strange to me that we have two separate attributes to set what the "progress" is. Ideally, we would only have a single attribute, that the component can persist on mount so that it knows how to animate any change to it.

When the component's props change (due to replacing of the XML), I only read the data from the `percent` attribute. However, I wanted to also support the use case of showing an animation during the initial mount. In that case, there's no existing attribute to compare the `percent` to, hence the need for `start-percent`. I see two other options:
1. If we render a progress bar with an animation duration, assume we're animating from a value of 0.
2. Don't support animation on mount. If a developer wants to show aprogress bar animation, they need to use a behavior to swap the progress bar after the page loads.

I don't like option 1 because it may not always be desirable to animate progress from 0 (for example, if someone is leveling up, you only want to show the bar increase from the previous level). I'm ok with option 2 because it simplifies the API and implementation of the progess bar, while still allowing someone to "animate on load" if they really want to. 

> 
> ```
> max-duration="15000"
> ```
> 
> naming is also a bit odd here, I think `animation-duration` would better represent what this is? Same, if we want to include the unit, then `animation-duration-ms` may be appropriate.

This naming is trying to reflect how I calculate the animation duration. Our current progress bars in the app exhibit what I consider to be a UI glitch: the animation duration is constant. Thus, animating from a value of `0` to `10` takes as long as animating from `0` to `100`. The two animations will appear to have different speeds.

With `max-duration`, I was attempting to convey that this would be the duration from `0` to `100`, and the actual animation duration will be interpolated based on the progress bar's current and future values.

We can use a simple `animation-duration-ms` attribute, but then the server would need to know the current state of the progress bar to properly calculate the duration.

DanielSchaffer:
What I've seen done for similar controls is to intentionally avoid specifying units of any kind, which actually makes it clearer since you don't have to worry about making assumptions. You have "start value", "value", "min value", and "max value". min/max can probably be defaulted to 0 and 100. The progress is determined based on `value / (max - min)` (and if specified, the start progress can be calculated the same way).

I like the `max-duration` concept, and it makes sense once explained, I just wish there was a different name we could use that would make it more immediately apparent.

flochtililoch:
> Don't support animation on mount. If a developer wants to show aprogress bar animation, they need to use a behavior to swap the progress bar after the page loads.

I support this, I think that makes more sense to let a behavior handle a value change post mount, when desired.

> We can use a simple animation-duration-ms attribute, but then the server would need to know the current state of the progress bar to properly calculate the duration.

Got it - I didn't think about that, nor understood the param this way, but this make a little more sense now. I would find it more obvious if we were representing the speed per percent. Assuming that `max-duration="15000"` is 150ms per percent, how about `amimation-ms-per-pct="150"`?

adamstep:
> What I've seen done for similar controls is to intentionally avoid specifying units of any kind, which actually makes it clearer since you don't have to worry about making assumptions. You have "start value", "value", "min value", and "max value". min/max can probably be defaulted to 0 and 100.

> This is basically how the native HTML [`<progress>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress)

@DanielSchaffer It looks like `<progress>` only supports a max value, not min. And it defaults to a range of 0-1 instead of 0-100. Do you have any opinions about that? Is there a need to support a min value other than 0?

adamstep:
> > Don't support animation on mount. If a developer wants to show aprogress bar animation, they need to use a behavior to swap the progress bar after the page loads.
> 
> I support this, I think that makes more sense to let a behavior handle a value change post mount, when desired.

Ok, I can make that change.

> > We can use a simple animation-duration-ms attribute, but then the server would need to know the current state of the progress bar to properly calculate the duration.
> 
> Got it - I didn't think about that, nor understood the param this way, but this make a little more sense now. I would find it more obvious if we were representing the speed per percent. Assuming that `max-duration="15000"` is 150ms per percent, how about `amimation-ms-per-pct="150"`?

That could work. It may conflict with Daniel's suggestion though, since he'd advocating not using percents, but letting the user set a max value or arbitrary units.

DanielSchaffer:
For us, there probably won't be a need for a min value other than 0 - but as a general use component, it could be useful to represent things that aren't zero-based. For example, a time span (e.g. the years 1920-2020).

adamstep:
Given the feedback, I'm going to close this PR and implement in our internal repo with the suggestions.

PR #187, opened 2020-09-09
POC: disappearing tooltip

adamstep:
Example of a floating tooltip that disappears after 5 seconds:

NOTE: to get this working, I needed to add z-index support for styles. This requires a new HV version in the app. On older versions, the tooltip will be partially under the scrolling list.


ashwinibm:
Looks good. 
I've seen the arrow on top generally done using the ::after pseudo selector. Will that be possible here? Could we use the `modifier`?

PR #194, opened 2020-10-16
Switch to new date picker component

adamstep:
[Asana](https://app.asana.com/0/923014529014430/1198498895658493/f)

The `<date-field>` component was using the deprecated `DatePickeriOS` and `DatePickerAndroid` components from core. These don't render correctly on iOS 14. To fix the issue, I'm switching to the [community version](https://github.com/react-native-datetimepicker/datetimepicker) of this component.

Changes:
- `DateTimePicker` now has a unified component API for iOS and Android. I removed the old async callback for Android.
- New `renderPickerModalAndroid` shares code with iOS but adds focus handling and a callback that handles dismiss action.
- Updated to the prop changes documented [here](https://github.com/react-native-datetimepicker/datetimepicker#migration-from-the-older-components).

### Testing
- [x] iOS 13 Simulator
- [x] iOS 14 Simulator
- [x] Android

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "test": "yarn generate test && yarn test:flow && yarn test:lint && yarn test:render && yarn test:unit && yarn test:validate-xml"
   },
   "dependencies": {
+    "@react-native-community/datetimepicker": "^3.0.3",
     "react-native-webview": "9.4.0",
     "tiny-emitter": "2.1.0",
     "url-parse": "1.4.3",
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "test": "yarn generate test && yarn test:flow && yarn test:lint && yarn test:render && yarn test:unit && yarn test:validate-xml"
   },
   "dependencies": {
-    "@react-native-community/datetimepicker": "^3.0.3",
+    "@react-native-community/datetimepicker": "3.0.3",
     "react-native-webview": "9.4.0",
     "tiny-emitter": "2.1.0",
     "url-parse": "1.4.3",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -40,6 +40,7 @@
     "test": "yarn generate test && yarn test:flow && yarn test:lint && yarn test:render && yarn test:unit && yarn test:validate-xml"
   },
   "dependencies": {
+    "@react-native-community/datetimepicker": "^3.0.3",
```
flochtililoch, line 43:
```suggestion
    "@react-native-community/datetimepicker": "3.0.3",
```

src/components/hv-date-field/index.js
```diff
@@ -237,7 +191,28 @@ export default class HvDateField extends PureComponent<
...
+   * Unlike iOS, the Android picker natively uses a modal. So we don't need
+   * to wrap it in an extra component, just render it when we want the modal
+   * to appear.
+   */
+  renderPickerModalAndroid = (): ?ReactNode => {
+    if (!this.state.focused) {
+      return null;
+    }
+    const onChange = (evt: Event, date: Date) => {
+      if (date === undefined) {
```
flochtililoch, line 207: it looks like `date` cannot be undefined from the flowtype definition. You might need to change the signature of this handler everywhere it's used to reflect this

adamstep: Good catch, updating to match the library:
https://github.com/react-native-datetimepicker/datetimepicker/blob/master/src/types.js#L50

vikynandha:
We currently use the `spinner` mode (I think that's what it's called) on Android, because selecting a date from long back -- e.g. DOB -- isn't very intuitive in the default calendar UI.

screenshot_20201020-102846

Just wanted to make sure this would be supported in the new component as well.

flochtililoch (approved):
Verified with iOS 14 / Xcode 12, looks as expected:

Screen Shot 2020-10-29 at 10 05 24 AM

adamstep:
> We currently use the `spinner` mode (I think that's what it's called) on Android, because selecting a date from long back -- e.g. DOB -- isn't very intuitive in the default calendar UI.
> 
> Just wanted to make sure this would be supported in the new component as well.

@vikynandha yes, spinner mode is still supported. I tested both modes on Android.

vikynandha:
Awesome, thanks @adamstep

PR #195, opened 2020-10-19
Copy to clipboard behavior

ashwinibm:
[Asana](https://app.asana.com/0/378088913887644/1198199975357167/f)

### Summary
- Copy to clipboard functionality implemented as a behavior. 
- Updated docs

Note: Although this component has been moved to a community library, there is an [issue](https://github.com/react-native-clipboard/clipboard/issues/71) which makes it unusable. So I've used the component from `react-native` itself for now. 

### Demo
copy_to_clipboard

### DIFF
--- a/docs/reference_behavior_attributes.md
+++ b/docs/reference_behavior_attributes.md
@@ -296,9 +296,9 @@ The `verb` attribute defines the HTTP method used to request the content specifi
 
 ## action
 
-| Type                                                                                                                                                                     | Required |
-| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | -------- |
-| **push** (default), new, back, close, navigate, deep-link, open-settings, replace, replace-inner, append, prepend, reload, dispatch-event, show, hide, toggle, set-value | No       |
+| Type                                                                                                                                                                                        | Required |
+| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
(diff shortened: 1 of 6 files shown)


flochtililoch:
I see that [the clipboard API is now deprecated](https://reactnative.dev/docs/clipboard), so we should integrate the react-native-community version.

ashwinibm:
> I see that [the clipboard API is now deprecated](https://reactnative.dev/docs/clipboard), so we should integrate the react-native-community version.

Yes, I had tried the community version first. But it was not working and I then saw this [issue](https://github.com/react-native-clipboard/clipboard/issues/71) that is still open

adamstep (approved):

ashwinibm:
@adamstep The schema doesn't allow duplicate attributes when performing the union operation. I can't have `value` as a `behavior` attribute since `option` already does and `option` also incorporates `behavior` attributes. I've renamed the attribute to `use-value`. If you have a suggestion for a better name, I'll update it.

References:
https://app.circleci.com/jobs/github/Instawork/hyperview/828
https://www.w3.org/2001/05/xmlschema-rec-comments.xml -> section "Question about unions of attribute uses"

ashwinibm:
> @adamstep The schema doesn't allow duplicate attributes when performing the union operation. I can't have `value` as a `behavior` attribute since `option` already does and `option` also incorporates `behavior` attributes. I've renamed the attribute to `use-value`. If you have a suggestion for a better name, I'll update it.
> 
> References:
> https://app.circleci.com/jobs/github/Instawork/hyperview/828
> https://www.w3.org/2001/05/xmlschema-rec-comments.xml -> section "Question about unions of attribute uses"

Reverted to have `value` as the behavior attribute name. Removed the duplicate `value` attribute from `option`.

PR #196, opened 2020-10-20
Added support for keyboard avoiding view

ashwinibm:
[Asana](https://app.asana.com/0/491474088351836/1198503649002413/f)

### Summary
The keyboard is hiding the footer section only on iOS as described in the asana ticket.
We have `KeyboardAwareScrollView` but that can't be used for fixed footers since they need to be outside the scroll view. 

**Changes**:
Allow specifying `avoid-keyboard` attribute on a view. That will wrap the contents in a `KeyboardAvoidingView` for iOS only. Android has built-in support for this functionality.

**Usage**
`<view avoid-keyboard="true">
    {% include 'worker/quiz/questionnaire/footer.xml' %}
  </view>`

### Demo

1. iOS 
keyboard_avoiding_ios 

2. Android
keyboard_avoid_android

3. Example Demo
hv_keyboard_avoiding_view

### DIFF
--- a/docs/reference_view.md
+++ b/docs/reference_view.md
@@ -112,5 +112,5 @@ If `hide="true"`, the element will not be rendered on screen. If the element or 
 | ------------------------- | -------- |
 | true, **false** (default) | No       |
 
-An attribute to solve the common problem of views that need to move out of the way of the virtual keyboard. It can automatically adjust its position based on the keyboard height. This is useful when you want keyboard avoiding behavior in non-scrollable views. It is applied only in iOS since
+An attribute to solve the common problem of views that need to move out of the way of the virtual keyboard. It can automatically adjust the position of its children based on the keyboard height. This is useful when you want keyboard avoiding behavior in non-scrollable views. It is applied only in iOS since
 Android has built-in support for avoiding keyboard.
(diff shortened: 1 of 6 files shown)


src/components/hv-view/types.js
```diff
@@ -12,6 +12,7 @@ import type { StyleSheet } from 'react-native/Libraries/StyleSheet/StyleSheetTyp
 
 export type InternalProps = {|
   accessibilityLabel?: ?string,
+  behavior?: ?string,
```
flochtililoch, line 15: perhaps restrict to the allowed values, i.e.

```suggestion
  behavior?: 'position',
```

docs/reference_view.md
```diff
@@ -104,3 +105,12 @@ A global attribute uniquely identifying the element in the whole document.
...
 If `hide="true"`, the element will not be rendered on screen. If the element or any of the element's children have a behavior that triggers on "load" or "visible", those behaviors will not trigger while the element is hidden.
+
+#### `avoid-keyboard`
+
+| Type                      | Required |
+| ------------------------- | -------- |
+| true, **false** (default) | No       |
+
+An attribute to solve the common problem of views that need to move out of the way of the virtual keyboard. It can automatically adjust the position of its children based on the keyboard height. This is useful when you want keyboard avoiding behavior in non-scrollable views. It is applied only in iOS since
+Android has built-in support for avoiding keyboard.
```
flochtililoch, line 116:
```suggestion
An attribute to solve the common problem of views that need to move out of the way of the virtual keyboard. It can automatically adjust the position of its children based on the keyboard height. This is useful when you want keyboard avoiding behavior in non-scrollable views. It is applied only in iOS since Android has built-in support for avoiding keyboard.
```

PR #200, opened 2020-10-30
Remove local state from <switch> element

adamstep:
Asana 1: https://app.asana.com/0/923014529014430/1179014325978272/f
Asana 2: https://app.asana.com/0/923014529014430/1144293982403313/f

This PR removes the use of local state from the `HvSwitch` component. I followed the same process as in #199 . Additionally, I addressed the issue of not being able to style the switch thumb, resulting in weird UI on Android. I'm using the `color` field from the applied styles to apply to the thumb.

iOS:
Screen Shot 2020-10-30 at 2 16 03 PM

Android:
Screen Shot 2020-10-30 at 2 34 00 PM
Screen Shot 2020-10-30 at 2 33 57 PM


flochtililoch:
> I'm using the `color` field from the applied styles to apply to the thumb.

Could you add a screenshot of how it looks like?

adamstep:
@flochtililoch added screenshot showing red color on the thumb of the switch.

PR #208, opened 2020-12-08
Fix: Picker field changes immediately

simhasrivatsa:
[Asana](https://app.asana.com/0/923014529014430/1189353069436699/f)
 [Video ios](https://drive.google.com/file/d/1TY_FnjlGPmcp2ubiBzqKr963S81rgtus/view?usp=sharing)
[Video android](https://drive.google.com/file/d/1IQ-0xBK7Xvr8Dyb5vuklchutH6dXUROk/view?usp=sharing)

The picker field from updating its state immediately. This change will stop that from happening

Screenshot 2020-12-08 at 10 00 36 PM

### DIFF
--- a/src/components/hv-picker-field/index.js
+++ b/src/components/hv-picker-field/index.js
@@ -157,7 +157,6 @@ export default class HvPickerField extends PureComponent<
     const props = {
       onValueChange: value => {
         this.setState({ pickerValue: value });
-        element.setAttribute('value', value || '');
       },
       selectedValue: this.state.pickerValue,
       style,


src/components/hv-picker-field/index.js
```diff
@@ -140,24 +148,23 @@ export default class HvPickerField extends PureComponent<
...
    * <picker-item> elements in the <picker-field> element.
    */
   renderPicker = (style: StyleSheetType<*>): ReactNode => {
-    const element: Element = this.props.element;
+    const element: Element = this.element;
     const props = {
       onValueChange: value => {
         this.setState({ pickerValue: value });
-        element.setAttribute('value', value || '');
       },
```
simhasrivatsa, line 168: @adamstep @flochtililoch removing this line is enough for the fix, rest are just state cleanups

src/components/hv-picker-field/index.js
```diff
@@ -52,23 +52,14 @@ export default class HvPickerField extends PureComponent<
   static localNameAliases = [];
   props: HvComponentProps;
   state: State;
+  element: Element;
 
   constructor(props: HvComponentProps) {
     super(props);
-    const element: Element = props.element;
-    const value: ?DOMString = element.getAttribute('value');
+    this.element = this.props.element.cloneNode(true);
```
adamstep, line 59: I'm a bit worried about storing `element` as an instance variable, rather than always referring to `props.element`.

We want this React component to re-render itself based on new props. For example, if some user action results in a change to the `<picker-field>` XML elemen, (for example, `set-value` behavior), then we want this component to use that new element to render. If we store a reference to the first element, then we won't use newer elements for updates.

If my understanding is correct, then the "set-value" example in the Hyperview demo app will not work with the picker field.

simhasrivatsa: the reason I created it as a class attr is because I didn't want to do `this.props.element = newElement`. would getDerivedStateFromProps help me here?

simhasrivatsa: I saw the set-value example and it does not work

adamstep (approved):

Issue #212, opened 2020-12-31
How to use SafeAreaView for hyperview


cengit:
The purpose of SafeAreaView is to render content within the safe area boundaries of a device, so in my code I just wrap Hyperview with the SafeAreaView component.

**actual result:**
the first page in Hyperview is good, it display all content in the safe area, but when navigate to next hyperview page, the SafeAreaView is not working

**expected result:**
content expected to render in the safe area for all page

 what should I do to make it work for all page?

in the demo it just set the Header height=72, I think this is not a best practices, so any solutions for this? thanks

```
<style id="Header"
             alignItems="center"
             backgroundColor="white"
             borderBottomColor="#eee"
             borderBottomWidth="1"
             flexDirection="row"
             height="72"
             paddingLeft="24"
             paddingRight="24"
             paddingTop="24" />
```

this is what I do for my code

```
import { SafeAreaView } from 'react-native-safe-area-view';

export default function MyAwesomeApp() {
  return (
    <SafeAreaView style={{ flex: 1 }}>
      <Hyperview
        entrypointUrl={url}
        fetch={fetch}
        back={this.back}
        closeModal={this.closeModal}
        openModal={this.openModal}
        navigate={this.navigate}
        push={this.push}
      />
    </SafeAreaView>
  );
}
```

cengit:
please see my screenshot as below.
first page: the header renders good, it renders in the safe area
<img width="452" alt="first-page" src="https://user-images.githubusercontent.com/3691806/103393143-3b5da780-4b5c-11eb-9a47-c6002283551e.png">

**second page: the header DID NOT renders in the safe area**
<img width="454" alt="second-page" src="https://user-images.githubusercontent.com/3691806/103393150-40baf200-4b5c-11eb-948b-ab1d264db15c.png">

adamstep:
Hi @cengit , sorry for the delay in response. One question: are you basing your example on the demo code in the Hyperview repo? Just want to understand which version of react-navigation you're using.

cengit:
@adamstep it's an old project, no on the demo code, we using "react-navigation": "2.18.0",

adamstep:
@cengit thanks for the info. I looked into the issue over the weekend. Unfortunately, there's no supported way to get a safe area view in the current version of Hyperview. In our own apps, we use header height like you saw in the demo. Of course, this is not ideal because different phones will need different heights.

I think we can support this feature in Hyperview by adding an attribute to the `<screen>` tag. Maybe something like
[...]

PR #217, opened 2021-01-20
Support safe area views

adamstep:
Addresses Issue #212 .

- Adds a `safe-area` attribute to view-like elements, most notably `<view>`, `<body>`, and `<header>`.
  - **NOTE**: `safe-area` will not have effect if the view also has `scroll="true"`.
- Updated the example app to use `safe-area` instead of padding through styling.
- Updated docs to explain the new attribute.
- Updated XML schema to accept the new attribute.

### Screenshots

The same XML with `safe-area="true"`, as viewed on iPhone 8 and 11 (with notch)
Screen Shot 2021-01-20 at 9 03 30 AM
Screen Shot 2021-01-20 at 9 13 35 AM


examples/case_studies/basic_form/index.xml
```diff
@@ -56,7 +56,7 @@
              marginTop="16" />
     </styles>
     <body scroll="true"
-          style="Body">
+          style="Body" >
```
flochtililoch, line 59:
```suggestion
          style="Body" safe-area="true">
```

examples/case_studies/lazy_load/index.xml
```diff
@@ -4,7 +4,7 @@
       <style id="Body"
              backgroundColor="white"
              flex="1"
-             paddingTop="48" />
+              />
```
flochtililoch, line 7: I think the `<body>` tag lower down this file should also set `safe-area` attribute

adamstep: The design of this screen is different from the others, it feature big colorful blocks that should not use `safe-area`.

adamstep:
@flochtililoch ready to re-review, Flow is passing now

flochtililoch (approved):

cengit:
awesome

PR #218, opened 2021-01-27
Style update: Add support for number of lines

reshavsingla:
Add support for number of lines for text


adamstep:
I thought we already supported this... what does the code in `services/index.js createProps()` do?

reshavsingla:
> I thought we already supported this... what does the code in `services/index.js createProps()` do?

Ahh..missed that just need to update the xsd to expose that for validation

flochtililoch (approved):

PR #219, opened 2021-01-27
Add `display` style prop to schema

flochtililoch:
Ref: https://reactnative.dev/docs/layout-props#display

### DIFF
--- a/schema/hyperview.xsd
+++ b/schema/hyperview.xsd
@@ -531,6 +531,16 @@
       <xs:attribute name="bottom" type="xs:integer"/>
 
       <xs:attribute name="color" type="hv:color"/>
+
+      <xs:attribute name="display">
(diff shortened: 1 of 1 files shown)


adamstep:
Should we update docs as well?

Wondering how this interacts with `hide="true"`. If a component is styled with `display="none"`, will hiding/showing the component as a load indicator work correctly?

flochtililoch:
> Should we update docs as well?

It's already part of the docs it looks like https://hyperview.org/docs/reference_style#display

> will hiding/showing the component as a load indicator work correctly?

If the component has `display="none"`, showing / hiding won't make a difference from a user standpoint. But the underlying view will be rendered when `display="none"` attribute is present but `hide="false"`, and it won't when `hide="true"`

adamstep (approved):

PR #215, opened 2021-01-09
Update dependencies

flochtililoch:
Update dependencies to match those in our apps, and update codebase to reflect new code requirements.
Note to reviewers: this PR is intended to be reviewed commit by commit - here's the breakdown:
- 93e2e71: Update dependencies and config files. Add patches for react-native and storybook, so that render tests don't fail
- 6943e13: Fix all flow errors: either resolve the actual issue, or add a `$FlowFixMe` comment in places where a fix is not straightforward
- a960489: Fix unit tests: importing `.` no longer work, so we import the full path instead
- 9b03b49: Run `eslint src --fix` to automatically fix auto-resolvable linting errors
- 06b0663: Remove circular dependencies by breaking up modules / moving code around
- 192f575: Fix `react/no-access-state-in-setstate` error - use callback when the current state needs to be referenced
- 5c9b8cf: Fix `sort-keys` and `react/destructuring-assignment` lint warnings
- 2522641: Enforce all previous warnings as errors

Issue #216, opened 2021-01-13
action="replace" doesn't take effect in some special cases


terryatgithub:
action="replace" doesn't take effect in some special cases in below cases as far I found: 
1. in Official Demo: 
   Behaviors -> Indicators -> Inline Indicator 
  (source file in hyperview/examples/behaviors/inline_button_indicator.xml)
   issue is :  
**when press the button 'Press me', after delay, the content of view which id="container" would not be replaced.**
   I checked above issue by expo client app installed in  emulator iPhone11(ios 14.3) and Android device(Android 10), both not work.

2. in real project as below code,
- in option 1: the loading effect will work, but replace action would NOT work;
- in option 2: the loading effect and replace aciton would work both;
the mere diffrence between two options is **put spinner tag out of option tag scope,** as you can find below:

```xml
   `<view>
          <text>select choice:</text>
          <select-single name="choice">
              <option value="1" trigger="select" action="replace"  delay="300" show-during-load="ID_Spinner1" target="modal-uninterested" href="/behaviors/_target_sibling_fragment.xml" >
                  <text>option 1</text>
                  <spinner id="ID_Spinner1" color="red" hide="true" />
              </option>
              <option value="2" trigger="select" action="replace" delay="300" show-during-load="ID_Spinner2"  target="modal-uninterested" href="/behaviors/_target_sibling_fragment.xml" >
                  <text>option 2</text>
              </option>
              <spinner id="ID_Spinner2" color="red" hide="true" />
          </select-single>
      </view>
`
```
I checked above issue in real project on Android device(Android 10);

please help check, feel free if need any more information, thanks a lot.

terryatgithub:
in case 2, the effect I want to realize is put spinner just after text in option tag, not outside of it.

PR #221, opened 2021-02-02
Remove `replace` from navigation actions

flochtililoch:
Somehow, the `replace` update action was also considered in the code to be a navigation action. This worked fined until #215, where our linter caught a new error where we would reference `NAV_ACTIONS.REPLACE` while undefined. The [introduced fix](https://github.com/Instawork/hyperview/commit/6943e13eee073f5206086b66a0c751e40534afc9#diff-e9726d469bba0bec8608c095ad5c361aeb95ea53afadd7ab44b462a2eb5e7467R379) caused the invalid/unexpected code to run, which would then start treating `replace` as a navigation action, and would prevent it to run as an update action.
This PR removes the confusion around it by removing the `replace` prop from the `HyperviewScreen` component (which reflects what is expected in the documentation), and removes the portion of unreachable code.


adamstep:
Does this address #216 ?

flochtililoch:
> Does this address #216 ?

no, this fix is only needed for the current work in progress branch with updated dependencies

adamstep (approved):

PR #227, opened 2021-02-24
translate id to different props depending on Platform.OS

linzhiyang85:
since there has conflict between properties testID and accessibilityLabel, they can not be set together.
need to specify a dedicated one depending on current rendering platform

### DIFF
--- a/src/services/index.js
+++ b/src/services/index.js
@@ -27,6 +27,7 @@ import { FORM_NAMES, LOCAL_NAME, NODE_TYPE } from 'hyperview/src/types';
 import HyperRef from 'hyperview/src/core/hyper-ref';
 import React from 'react';
 import type { StyleSheet } from 'react-native/Libraries/StyleSheet/StyleSheetTypes';
+import { Platform } from 'react-native';
 
 /**
  * This file is currently a dumping place for every functions used accross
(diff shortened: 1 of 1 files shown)


adamstep:
@flochtililoch I don't remember how we implemented this. Is there a conflict or difference in how test IDs work on iOS and Android?

flochtililoch:
We use both side by side on our code base - one is used by iOS while the other is used by Android. @linzhiyang85 can you describe what issue you're running into by having them both set at the same time?

linzhiyang85:
It has no problem for Android when both of the properties are given.
But for iOS, when we have a single testID, the attribute "name" will be assigned with the id's value and the other two attributes "value" and "label" will contains the actual literal text, for example the <text> tag. This is enough for finding by accessibility id. However if adding the accessibilityLabel, after rendering, the "value" and "label" attributes are assigned with the id's value, instead of the text which is displaying, then we have no way to retrieve the text value.

For your reference:
https://discuss.appium.io/t/react-native-app-on-ios-element-has-the-same-value-for-name-values-text-accessibilitylabel-etc/22237/7
https://github.com/tipsi/react-native-testid/blob/master/index.js

flochtililoch:
Makes sense, thanks for submitting that fix! I'll create a new release shortly

flochtililoch:
Released in 0.36.0

PR #243, opened 2021-03-12
Lint previously ignored files

flochtililoch:
- 8ba6a43 Lint `src/index.js` (linting errors in 87446a5)
- a2308fe Lint `src/VisibilityDetectingView.js` (linting errors in 29ad64a)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,5 @@
 flow-typed
 *.gen.js
 website/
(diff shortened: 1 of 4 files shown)


adamstep (approved):
When we incorporate the HV version with these changes, let's be sure to get a round of QA in case we missed any subtle changes.

PR #257, opened 2021-05-06
fix: refine flow when action=navigate

terryatgithub:
**Please note the only difference with #258 is that, #258 add some demo to verify the revises.**
please see which one to merge according to needs. and close the other one.

there are 2 revises in this PR:
1. Revise the variable `routeKeys` to global, or else it will be a local variable in every instance of Class HyperScreen.  
    The issue is when run `this.navigation.navigate()` to an intermediate page, it will always navigate back to the first page(component) in the navigation stack (in demo, it will be the examples/index.xml), and replace the first page with its own content.

2. Revise in the `navigation/index.js` to fix below issue: 
  When navigate to a new page (not in navigation stack yet), it will alse navigate back to the first page(component) in the navigation stack (in demo, it will be the examples/index.xml), and replace the first page with its own content.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -173,7 +173,6 @@ export default class HyperScreen extends React.Component {
       this.needsLoad = false;
     }
   }
-  
   /**
    * Performs a full load of the screen.
    */
--- a/src/index.js
+++ b/src/index.js
@@ -173,6 +173,7 @@ export default class HyperScreen extends React.Component {
       this.needsLoad = false;
     }
   }
+
   /**
    * Performs a full load of the screen.
    */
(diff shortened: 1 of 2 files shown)


terryatgithub:
If this PR is merged, could you pls help to make a new release ? thank you so much.

adamstep:
@terryatgithub We did not have time to look at this today, but we will review in the next day.

terryatgithub:
> @terryatgithub We did not have time to look at this today, but we will review in the next day.

hi Adam, it's ok. let me know if you need more informations. 
have a good day.

src/index.js
```diff
@@ -172,7 +173,6 @@ export default class HyperScreen extends React.Component {
       this.needsLoad = false;
     }
   }
```
adamstep, line 175:
```suggestion
  }

```

PR #258, opened 2021-05-06
fix: refine flow when action=navigate (With Demo)

terryatgithub:
Detail comment pls see #257.

The only difference with #257 is add some demo to verify the revises, please see which one to merge according to needs. and close the other one. 
Thank you very much.


terryatgithub:
duplicated with #257 , close it.

PR #262, opened 2021-05-07
fix error when destructure body from NodeList

terryatgithub:
to fix below error when debug by `yarn sync`:
image


adamstep (approved):
Good catch, thanks!

terryatgithub:
@adamstep @flochtililoch could you please help to build a new release ? thanks.

Issue #270, opened 2021-05-11
Feature: introduce new actions include and include-inner 


terryatgithub:
in order to implement some complicated UI like below two-layer cascading menu, it's found that current actions can not match this requirement. 

![image](https://user-images.githubusercontent.com/12246394/117808129-16f2a400-b28f-11eb-802a-c7949fb43ced.png)

> And below actions are needed in this case
- `include`
`include` means replace the target element with the `contents of source element`, instead of source element itself.
- `include-inner`
`include-inner` means replace the target element's children with `the contents of source element`. 

> The name and meaning of both new actions are referred from the [include syntax of Thymeleaf](https://www.thymeleaf.org/doc/tutorials/3.0/usingthymeleaf.html#difference-between-thinsert-and-threplace-and-thinclude).

I'm testing these new actions now, and want to hear your opinions before raising a PR.
the implement code is like below:

```javascript
// hyperview/src/services/behaviors/index.js
/**
 * Returns a new Document object where the given action was applied to the target element
 * with the new element.
 */
export const performUpdate = (
  action: UpdateAction,
  targetElement: Element,
  newElement: Element,
): Document => {
  if (action === ACTIONS.REPLACE) {
    const { parentNode } = targetElement;
    if (parentNode) {
      parentNode.replaceChild(newElement, targetElement);
      return shallowCloneToRoot((parentNode: any));
    }
  }

  if (action === ACTIONS.REPLACE_INNER) {
    let child: ?Node = targetElement.firstChild;
    // Remove the target's children
    while (child !== null && child !== undefined) {
      const nextChild: ?Node = child.nextSibling;
      targetElement.removeChild(child);
      child = nextChild;
    }
    targetElement.appendChild(newElement);
    return shallowCloneToRoot(targetElement);
  }

  // new action : include
  if (action === ACTIONS.INCLUDE) {
    const { parentNode } = targetElement;
    if (parentNode) {
      let child: ?Node = newElement.firstChild;
      // Insert the newElement's children
      while (child !== null && child !== undefined) {
        const nextChild: ?Node = child.nextSibling;
        parentNode.insertBefore(child, targetElement);
        child = nextChild;
      }
      parentNode.removeChild(targetElement);
      return shallowCloneToRoot((parentNode: any));
    }
  }

  // new action : include-inner
  if (action === ACTIONS.INCLUDE_INNER) {
    let child: ?Node = targetElement.firstChild;
    // Remove the target's children
[...]

PR #271, opened 2021-05-12
introduce two new actions: include, include-inner

terryatgithub:
This PR is to implement the requirement of #270 .
Please check it for detailed info. 

if any questions pls let me know.
thank you very much.

### DIFF
--- a/schema/hyperview.xsd
+++ b/schema/hyperview.xsd
@@ -70,6 +70,8 @@
       <xs:enumeration value="dispatch-event" />
       <xs:enumeration value="replace" />
       <xs:enumeration value="replace-inner" />
+      <xs:enumeration value="include" />
+      <xs:enumeration value="include-inner" />
       <xs:enumeration value="append" />
       <xs:enumeration value="prepend" />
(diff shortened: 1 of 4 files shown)


terryatgithub:
we implemented the complicated [cascader menu ](https://github.com/Instawork/hyperview/issues/270#issue-887003561)with the new actions `include` and `include-inner`.

adamstep:
Hi @terryatgithub . Thanks for the PR implementing these additions. I am not sure if we want to add them to the core Hyperview, I need to take more time to understand whether this is a useful action for general use.

In the meantime, is it possible to add `include` and `include-inner` to your app as a custom behavior? I think it should be possible to do it, and you won't be blocked on updates to the core HV library.

terryatgithub:
ok, Adam, thanks for reply. I can understand it. and i will try the method of custom-behavior. 
and maybe later I will also create a demo implemented by `include` and `include-inner` for your reference.
thanks so much.

terryatgithub:
Because of this wonderful realization in #309  by Adam, this PR is not needed anymore.

Issue #267, opened 2021-05-10
Feature: support customized ErrorPage


terryatgithub:
now, when hyperview detected xml href or format error, below error page will show.
but it can not satisfy our UX requirement. 
we need replace it with a customized error component. 
Do you consider support this feature in future? 
thank you very much.

![image](https://user-images.githubusercontent.com/12246394/117618992-4aef9b80-b1a1-11eb-9044-6e967f31bd09.png)

![image](https://user-images.githubusercontent.com/12246394/117619625-15977d80-b1a2-11eb-8a53-d1dc3f47dd40.png)

terryatgithub:
yes, when I tried this approach, I thought maybe a backup error xml version can be storaged in client locally. and only show when no Internet connection.

and this schedule is exciting, thank you so much.
`It's a fairly simple change, we can try to have it ready this week.`

PR #273, opened 2021-05-13
Support custom error screen

adamstep:
This PR addresses #267 . Hyperview clients can pass in a React component as an `errorScreen` prop to the `Hyperview` component. This component will be displayed if a screen fails to load due to bad XML parsing or network problems.

Example for `demo/screens/HyperviewScreen.js`:
```js
function ErrorScreen(props) {
  return (
    <View style={{flex: 1, justifyContent: 'center', alignItems: 'center'}}>
      <Text style={{fontSize: 24, fontWeight: 'bold', color: 'red'}}>Custom Error Screen</Text>
      <Text style={{color: 'blue'}}>{props.error.name}</Text>
      <Text>{props.error.message}</Text>

    </View>
  );
};

export default class HyperviewScreen extends React.PureComponent {
  // code ommited...

  render() {
    const navigation = this.props.navigation;
    const entrypointUrl = navigation.state.params.url;

    return (
      <Hyperview
        back={this.goBack}
        closeModal={this.closeModal}
        entrypointUrl={entrypointUrl}
        fetch={this.fetchWrapper}
        navigate={this.navigate}
        navigation={navigation}
        openModal={this.openModal}
        push={this.push}
        formatDate={this.formatDate}
        errorScreen={ErrorScreen}
      />
    );
  }
}
```

For a network error the result would look like this:

https://user-images.githubusercontent.com/1034502/118064694-5c2ada80-b350-11eb-986e-252edac6dd95.mov

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -225,13 +225,12 @@ export default class HyperScreen extends React.Component {
    */
   render() {
     if (this.state.error) {
-      return (
-        <LoadError
-          error={this.state.error}
-          onPressReload={() => this.reload()}
-          onPressViewDetails={(uri) => this.props.openModal({url: uri})}
-        />
-      );
+      const errorScreen = this.props.errorScreen || LoadError;
+      return React.createElement(errorScreen, {
+        error: this.state.error,
+        onPressReload: () => this.reload(),
+        onPressViewDetails: (uri) => this.props.openModal({url: uri}),
+      });
     }
     if (!this.state.doc) {
       return (
--- a/src/index.js
+++ b/src/index.js
@@ -228,7 +228,7 @@ export default class HyperScreen extends React.Component {
       const errorScreen = this.props.errorScreen || LoadError;
       return React.createElement(errorScreen, {
         error: this.state.error,
-        onPressReload: () => this.reload(),
+        onPressReload: this.reload,
         onPressViewDetails: (uri) => this.props.openModal({url: uri}),
       });
     }


adamstep:
@terryatgithub Note that the `errorScreen` prop can be any react component. That means it is possible to render Hyperview XML for this error screen. In your React Native code, you could fetch the XML for the error screen, cache locally, and render it for the error screen.

src/index.js
```diff
@@ -225,13 +225,12 @@ export default class HyperScreen extends React.Component {
...
-        <LoadError
-          error={this.state.error}
-          onPressReload={() => this.reload()}
-          onPressViewDetails={(uri) => this.props.openModal({url: uri})}
-        />
-      );
+      const errorScreen = this.props.errorScreen || LoadError;
+      return React.createElement(errorScreen, {
+        error: this.state.error,
+        onPressReload: () => this.reload(),
```
flochtililoch, line 231: This prevents creating a new function on every render, and would allow the custom error component to pass arguments to the original `reload` method (which accepts an optional URL)
```suggestion
        onPressReload: this.reload,
```

flochtililoch (approved):

terryatgithub:
@adamstep it's so nice!! thank you so much. we will try it .  @cengit pls note 
and will you release a new version including recent changes ?

cengit:
great enhancement, a question: how can i use the onPressReload method for the custom error screen? can you give me an example? thanks

adamstep:
> great enhancement, a question: how can i use the onPressReload method for the custom error screen? can you give me an example? thanks

Check out https://github.com/Instawork/hyperview/blob/master/src/core/components/load-error/index.js#L53

adamstep:
version v0.38.2 is available now.

cengit:
@adamstep 
cool!
I have a question: 
on custom error page, if i want to show different error for different actions like below array, how can I get the action that cause the error
```ts
const actions = ['push','new','navigate', 'reload', 'replace', 'replace-inner', 'append', 'prepend'];
```

cengit:
trying to get more detail of the error, how can I get the exact hv action ( like below array)when an error happen?
```ts
Actions = ['new','push','replace', 'replace-inner', 'append', 'prepend',...];
```
for example:
when below replace action failed on fetching /behaviors/_target_sibling_fragment.xml, how can I get this action when the fetch request response? 

```xml
 <view
        action="replace"
        delay="300"
        href="/behaviors/_target_sibling_fragment.xml"
        style="Button"
        target="container"
      >
        <text style="Button__Label">Press me</text>
      </view>
      <view id="container" style="Container" />
```

adamstep:
Hi @cengit . I believe the error screen will only appear on actions that load content in a new screen (`push`, `new`, `reload`). I believe it won't appear for update actions that replace content on a given screen, so your question may not be relevant.

Can you give an example of what you are trying to achieve? Maybe there's another way.

cengit:
> Hi @cengit . I believe the error screen will only appear on actions that load content in a new screen (`push`, `new`, `reload`). I believe it won't appear for update actions that replace content on a given screen, so your question may not be relevant.
> 
> Can you give an example of what you are trying to achieve? Maybe there's another way.

no, actually it will goes to the error screen for an update actions, see below example:

```xml
<view
        action="replace"
        delay="300"
        hide-during-load="myIndicator1 myIndicator2"
        href="/behaviors/error_fragment.xml"
        style="Button"
        target="container"
      >
        <text style="Button__Label">Press me</text>
 </view>
<view id="container" style="Container" />
```

error_fragment.xml
```xml
<doc xmlns="https://hyperview.org/hyperview">
          <screen><view id="container" style="Container">
          <text style="Container__Label">test rendering error!</text>
        </view></screen></doc>
```
we will get this error screen:
image

what I trying to achieve is: 

- for new screen action(push, new, reload), when catch an error, displaying an error-screen is fine
- but when an update action('replace', 'replace-inner', 'append', 'prepend'), we just display a text error in the target view(it's view#container as above example)

cengit:
now I'm using HyperScreen.needsLoad to determine current action is new or update.
```ts
<Hyperview
        ref={hyperviewRef}
        ...
      />

const curAction = hyperviewRef.current?.needsLoad === true ? 'new' : 'update';
```

adamstep:
@cengit I think this approach will work for now, but it may break if we refactor the internals of `Hyperview` component.

Another option would be to add more data to the error, or more local state about the type of action causing the load. For example, anywhere where we set the `error` state, we can also set the action (new screen vs fragment update): https://github.com/Instawork/hyperview/blob/master/src/index.js#L203

PR #274, opened 2021-05-17
filter null before remove routeKey

terryatgithub:
hi @adamstep , this PR fix below error:
`this.state.url` will be null in some case during development, so add a protection. 
pls check, thanks.

image


adamstep:
@flochtililoch can you review? thanks

flochtililoch (approved):

flochtililoch:
Thanks for the fix @terryatgithub !

PR #277, opened 2021-05-18
update editable attribute for text-field

Jasonwei1106:
#276 This pull request is used for solving this issue. If you have any other suggestions, please let me know. Thank you

PR #278, opened 2021-05-18
fix: text field attribute editable type transform

Jasonwei1106:
#277 This pull request is used for solving this issue. If you have any other suggestions, please let me know. Thank you


adamstep:
Hi @Jasonwei1106 , is there a reason this was closed? Did you find another solution?

Jasonwei1106:
Hi @adamstep, that 's the basically the same pull request. I was just trying to do the pull request in a formal way to see if it can pass all the tests. But it seems that wasn't the issues, so I closed it. Sorry for the confusion.

PR #280, opened 2021-05-20
Feature/introduce two new actions: include, include-inner with a demo

terryatgithub:
hi @adamstep , this PR provides a demo to present the usage of new action `include` `include-inner` (maybe you can rename it as needed).   #270 #271 
please refer, thank you. 

https://user-images.githubusercontent.com/12246394/118948745-278bd400-b98b-11eb-8697-8fa966d6af98.mov


terryatgithub:
the key points of this implementation is as below:
when select one item in a sublist, how to reset `selected` state in another sublist, and how to keep active state of current sublist in the future.

terryatgithub:
Because of this wonderful realization in #309  by Adam, this PR is not needed anymore.

PR #301, opened 2021-05-25
Set cors to `*` demo `test:xmlserver` command

enkidulan:
Following `Getting started` guide I stuck on demo-project refusing to load because of cors missing allow header:
 
Screenshot_20210524_231912

This PR sets cors to `*` for `http-server` so that the demo-app can be accessed without spending any extra time on investigating why it's not loading.


adamstep:
Thanks @enkidulan !

PR #304, opened 2021-05-25
Feature: front-end form validate tentative idea

terryatgithub:
This PR probably should not raised in the form of PR, it only  provided an experimental idea to validate form field in front-end.
demo-frontend-form-validate

The basic idea is:
1.  The code of form validate logic written by Javascript is located in server side: `examples/frontend_form_validate/verify.js `. and injected to `examples/frontend_form_validate/index.xml` when page loaded.
2. In client side, it's triggered by a custom behavior as below, to load the `form validate code`, and run it.
```xml
<behavior trigger="blur" action="verify" />
```

This is just an experimental method for reference. 
The biggest advantage is: 
-  form-validate is completely done in front-end, without any http request.

And the biggest disadvantages are:
- There are lots of hard code ID or file format in `xml` or `js` file.
- Potential security issues maybe injected by Javascript.

Please help check it, we'd love to hear your opinion or comment about it.
because in our app, we need do a few action about user info fill and submit, and the function of front-end form validation is vital to us.

thank you very much.

### DIFF
--- a/demo/behaviors/verify.js
+++ b/demo/behaviors/verify.js
@@ -1,0 +1,147 @@
+import type { DOMString, Document, Element } from 'hyperview/src/types';
+import { shallowCloneToRoot } from 'hyperview/src/services';
+
+const verifyBehavior = {
+  action: 'verify',
+  callback: (element: Element,
+    onUpdate: HvComponentOnUpdate,
+    getRoot: HvGetRoot,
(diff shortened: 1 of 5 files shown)


terryatgithub:
please close this PR when you think it's appropriate. 
thanks.

adamstep:
@terryatgithub I will close this PR but we can keep the conversation going on #305 .

PR #310, opened 2021-05-27
bug of select-single in cascader-demo

terryatgithub:
hi @adamstep, base on your great implementation, I did some test and find a little maybe bug as below:

> When there is a default value set, 
> What we expected is show `level1` and `level2` simultaneously.
> The fact is below three ways all have some abnormal issue:
1.  set attribute `value`  in both `select-single`, it will run error when select in `level1 list`:

https://user-images.githubusercontent.com/12246394/119785030-67116d80-bf01-11eb-9b82-15947a633051.mov

2. set attribute `selected`  in both `select-single`, can not show `level2` list when enter for the first time:

https://user-images.githubusercontent.com/12246394/119785049-6c6eb800-bf01-11eb-8df1-aa9fba929c4d.mov

3. set attribute `value selected`  separatedly in both `select-single`, need **double click** to show both level:

https://user-images.githubusercontent.com/12246394/119785039-6aa4f480-bf01-11eb-97d0-173b60fdb3a4.mov

### DIFF
--- a/examples/case_studies/cascader.xml
+++ b/examples/case_studies/cascader.xml
@@ -102,7 +102,13 @@
       </style>
     </styles>
     <body style="Body" safe-area="true">
-      <view scroll="true" style="Main" trigger="refresh" action="reload" href="#">
+      <view
+        scroll="true"
+        style="Main"
+        trigger="refresh"
+        action="reload"
+        href="#"
+      >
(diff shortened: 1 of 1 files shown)


flochtililoch:
@adamstep can you remember if this issue was fixed?

PR #312, opened 2021-05-27
`<list>` fix: Prevent rendering items from nested `<list>`

flochtililoch:
Render list items that are direct children of the `<list>` element, or nested under a `<items>` children node.

### DIFF
--- a/src/components/hv-list/index.js
+++ b/src/components/hv-list/index.js
@@ -17,6 +17,7 @@ import { FlatList } from 'react-native';
 import type { HvComponentProps } from 'hyperview/src/types';
 import { LOCAL_NAME } from 'hyperview/src/types';
 import type { State } from './types';
+import { getAncestorByTagName } from 'hyperview/src/services';
 
 export default class HvList extends PureComponent<HvComponentProps, State> {
   static namespaceURI = Namespaces.HYPERVIEW;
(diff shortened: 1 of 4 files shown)


src/components/hv-list/index.js
```diff
@@ -71,14 +71,28 @@ export default class HvList extends PureComponent<HvComponentProps, State> {
     const showScrollIndicator =
       this.props.element.getAttribute('shows-scroll-indicator') !== 'false';
 
+    const itemsContainer =
+      this.props.element.firstChild &&
```
adamstep, line 75: maybe I'm missing something, but `<items>` doesn't have to be the first child, right?

flochtililoch: You're thinking of this kind of markup, or did you have something else in mind?
```
<list>
  <item/>
  <item/>
  <items>
    <item/>
    <item/>
  </items>
</list>
```

flochtililoch: I've updated to account for this use case, let me know what you think

adamstep: Yes, there can also be many <items> elements if we’ve loaded multiple pages.

flochtililoch: > there can also be many elements if we’ve loaded multiple pages

Can you clarify this? Does the markup above not accounting for this case?

adamstep:
Would it relatively easy to add unit tests to account for all of these cases?

flochtililoch:
> Would it relatively easy to add unit tests to account for all of these cases?

I'll take a look. I think I'll have to refactor it to isolate the code so that I can mock elements

flochtililoch:
@adamstep I've added tests in 4b8cb95

adamstep (approved):
Thanks for the tests, it adds confidence to the functionality here in case we refactor in the future.

Issue #283, opened 2021-05-24
ran-once is set on triggering element rather than behavior element.


adamstep:
If a <behavior> element has once="true", then ran-once="true" should be set on that <behavior> element, not on the triggering element. Otherwise, we can block other behaviors from running.

adamstep:
This code adds `ran-once` to the `behavior` tag: https://github.com/Instawork/hyperview/blob/master/src/index.js#L331

But this code checks `ran-once` on the current element, not the behavior: https://github.com/Instawork/hyperview/blob/master/src/index.js#L391

PR #315, opened 2021-06-07
bug fix: set ran-once flag on the triggering behavior element

adamstep:
Addresses #283. `ran-once="true"` should be set on the same element as `once="true"`.

Added example showing the fix:
https://user-images.githubusercontent.com/1034502/121265518-c8c0c880-c86d-11eb-874c-dada75816116.mov

### DIFF
--- a/examples/behaviors/_once_1.xml
+++ b/examples/behaviors/_once_1.xml
@@ -1,0 +1,4 @@
+<text
+  style="Container__Label"
+  xmlns="https://hyperview.org/hyperview"
+>Added each time</text>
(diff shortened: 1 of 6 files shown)


flochtililoch (approved):

adamstep:
@flochtililoch Added one more change to properly support this. I found a case where `behaviorElement` was not being supplied to the `onUpdate` code, added in https://github.com/Instawork/hyperview/pull/315/commits/e367281c71db2a9349045b32d064186bd640401e#diff-0f25ff6f47414bfa28c4e330d1463d407afff4cf72443249a9f91b385d3ce7eeR142.

flochtililoch (approved):

PR #321, opened 2021-06-28
fix: date-picker supports locale

terryatgithub:
add attribute 'locale' to date-picker to support locale as below
image

### DIFF
--- a/src/components/hv-date-field/index.js
+++ b/src/components/hv-date-field/index.js
@@ -240,11 +240,13 @@ export default class HvDateField extends PureComponent<HvComponentProps> {
   renderPicker = (onChange: (evt: Event, date?: Date) => void): ReactNode => {
     const minValue: ?DOMString = this.props.element.getAttribute('min');
     const maxValue: ?DOMString = this.props.element.getAttribute('max');
+    const locale: ?DOMString = this.props.element.getAttribute('locale');
     const minDate: ?Date = HvDateField.createDateFromString(minValue);
     const maxDate: ?Date = HvDateField.createDateFromString(maxValue);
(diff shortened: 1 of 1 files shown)


adamstep:
Can locale be inferred from the request headers? I'm thinking something similar to HTML, where we use the device language and HTTP `Language` header to determine the locale to use for these components.

terryatgithub:
hi Adam, it sounds great. do you have some schedule to implement it? 
as we need it in the next release.
thanks.

adamstep:
@terryatgithub I looked at this some more: https://github.com/react-native-datetimepicker/datetimepicker#locale-optional-ios-only

> Allows changing of the locale of the component. By default it uses the device's locale.
> <RNDateTimePicker locale="es-ES" />

So, according to the docs, the date picker should already use the device locale. Is this not happening in your app? Or is there a use case for a different locale in the date picker than on the device itself?

terryatgithub:
> @terryatgithub I looked at this some more: https://github.com/react-native-datetimepicker/datetimepicker#locale-optional-ios-only
> 
> > Allows changing of the locale of the component. By default it uses the device's locale.
> 
> So, according to the docs, the date picker should already use the device locale. Is this not happening in your app? Or is there a use case for a different locale in the date picker than on the device itself?

yes, in practical use, we found that the 'UIDatePicker' component in iOS can not get the device(or simulator) locale as expected. 
so it needs to set attribute 'locale' manually. 
the below issue may help:
https://stackoverflow.com/questions/1522210/nslocale-currentlocale-always-returns-en-us-not-users-current-language

adamstep:
Got it, I'm ok to add this as a workaround for now.

adamstep:
Included in latest release: v0.39.2

PR #324, opened 2021-07-12
Fix: Set default color in all activity indicators

ashwinibm:
Apply default color for other `ActivityIndicators`

Reference: 
https://github.com/Instawork/hyperview/pull/275
https://github.com/Instawork/instawork/pull/13363/files#r667629321

### DIFF
--- a/src/components/hv-spinner/index.js
+++ b/src/components/hv-spinner/index.js
@@ -24,7 +24,7 @@ export default class HvSpinner extends PureComponent<HvComponentProps> {
   props: HvComponentProps;
 
   render() {
-    const color = this.props.element.getAttribute('color') || undefined;
+    const color = this.props.element.getAttribute('color') || "#8d9494";
     return <ActivityIndicator color={color} />;
   }
 }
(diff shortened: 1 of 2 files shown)


flochtililoch (approved):

adamstep (changes requested):
This is Instawork's default color. It shouldn't be hard-coded in HV.

adamstep:
Can we use the default color per-platform? Is that what the RN spinner does by default?

flochtililoch:
The version of RN we're using has a bug that prevents Android from using the default color on spinners. We've already taken the step to hardcode our color in #275, but that might not be the most elegant solution. Maybe we can undo #275 and instead write a patch for our RN app, so that default color is not ignored on Android.

adamstep:
I'm ok hard-coding the defaults in HV. But ideally we can work around the RN bug by hard-coding the original default color.

ashwinibm:
Closing in favor of using a patch in the client app instead https://github.com/Instawork/mobile/pull/3848

ashwinibm:
Closing in favor of using a patch in the client app instead https://github.com/Instawork/mobile/pull/3848

Issue #286, opened 2021-05-24
Support custom RN components for loading state


adamstep:
- When creating a Hyperview screen component, pass props to specify default for loading screens
- nuance: need to show close or back button depending on where the screen is created

We recently added this for error screen, loading can be similar.

PR #330, opened 2021-07-27
Proposal: Allow overriding of `Loading` / `LoadError` / `RefreshControl` components

flochtililoch:
Original intent is to allow customization of the pull to refresh UI - right now, it's not explicit at what point the user needs to pull in order to trigger the refresh. We could improve UX by indicating "pull some more" and "release now", or adding any custom animation.

This PR makes components used for rendering loading, load error and refresh states overridable by setting them as static properties of their parent components. The parent component can then be extended and used in place of the original component. It is aimed as a proposal to replace current limited support for custom error screen that uses `errorScreen` property.

I chose the approach of static properties / extensible classes over props, because the props from the root Hyperview component are not accessible to `HvList` and `HvSectionList` instances.

**Usage:**
For `Loading` / `LoadError`, it is as simple as extending the base `Hyperview` component, and redefining the static properties, then make sure to render the extended component instead of the original one.
i.e.

*before*
```js
import Hyperview from 'hyperview';

...

<Hyperview ... />
```

*after*
```js
import HyperviewCore from 'hyperview';
import Loading from './Loading';
import LoadError from './LoadError';

class Hyperview extends HyperviewCore {
  static LoadingComponent = Loading
  static LoadErrorComponent = LoadError;
}

...

<Hyperview ... />
```

For `RefreshControl`, this is done by extending the HvList / HvSectionList, and redefining the static property `RefreshComponent`, then re-register extended components (which will replace the default core component).
i.e.

```js
import { HvList as HvListCore, HvSectionList as HvSectionListCore } from 'hyperview';
import Refresh from './Refresh';

class HvList extends HvListCore {
  static RefreshComponent = Refresh;
}

class HvSectionList extends HvSectionListCore {
  static RefreshComponent = Refresh;
}

Hyperview.registerComponents([
  HvList,
  HvSectionList,
]);
```

TODO:
- [ ] Add support for replacing the default refresh control used in scroll views - this might need to be done by replacing the current `addHref` helper by an actual React component, that can also be overridden as a static property of `Hyperview`
- [ ] Update demo app to showcase overrides usage

Resolves https://github.com/Instawork/hyperview/issues/286

### DIFF
--- a/src/components/hv-list/index.js
+++ b/src/components/hv-list/index.js
@@ -11,9 +11,9 @@
 import * as Dom from 'hyperview/src/services/dom';
 import * as Namespaces from 'hyperview/src/services/namespaces';
 import * as Render from 'hyperview/src/services/render';
+import { FlatList, RefreshControl } from 'react-native';
 import React, { PureComponent } from 'react';
 import { DOMParser } from 'xmldom-instawork';
-import { FlatList } from 'react-native';
 import type { HvComponentProps } from 'hyperview/src/types';
 import { LOCAL_NAME } from 'hyperview/src/types';
 import type { State } from './types';
@@ -24,6 +24,8 @@ export default class HvList extends PureComponent<HvComponentProps, State> {
   static localName = LOCAL_NAME.LIST;
 
   static localNameAliases = [];
+
+  static RefreshComponent = RefreshControl;
 
   parser: DOMParser = new DOMParser();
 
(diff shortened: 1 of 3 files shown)


adamstep:
Regarding static properties vs component props: could we use React Context instead to make the customized components available to the lists? It would be nice if we can pass everything through props on the main `Hyperview` component for consistency, but I suppose this is an advanced case and maybe extending existing components is fine.

PR #337, opened 2021-08-10
feat(schema): add "resizeMode" support

DanielSchaffer:
# Summary
- add `resizeMode` enumeration type
- add `resizeMode` attribute to `<image>` and `<style>` elements

Verified locally via `be_valid_hyperview` expectation in a python unit test

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hyperview",
-  "version": "0.40.0",
+  "version": "0.40.1",
   "main": "lib/index.js",
   "description": "React Native client for Hyperview XML",
   "homepage": "https://hyperview.org",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hyperview",
-  "version": "0.40.0",
+  "version": "0.40.1",
```
flochtililoch, line 3: FWIW, we don't need to update the NPM module, the schema is not bundled as part of the module but rather published via the Hyperview website as soon as the master branch is updated.

DanielSchaffer: Oh shoot, just saw this after I merged. Alright, good to know for any future changes.

flochtililoch: To note also: version number should only be bumped with our versioning script, so that this ref and NPM don't get out of sync. In this present case, it's not too big of a deal, we'll just effectively skip 0.40.1 version when we patch release next.

DanielSchaffer: Ohhh fancy! I like it. I'm used to just doing it myself.

PR #342, opened 2021-08-13
Fix: undo hardcoded activity indicator color

ashwinibm:
Undoing the change in https://github.com/Instawork/hyperview/pull/275 since we wrote a patch in the client app instead.
[Ref 1](https://github.com/Instawork/hyperview/pull/324#issuecomment-878450708)

Context:
React Native 0.63 does not have a default color for ActivityIndicator on Android. See [reference](https://github.com/facebook/react-native/pull/30057) for creating a patch on the client app.

### DIFF
--- a/src/components/hv-spinner/index.js
+++ b/src/components/hv-spinner/index.js
@@ -24,7 +24,7 @@ export default class HvSpinner extends PureComponent<HvComponentProps> {
   props: HvComponentProps;
 
   render() {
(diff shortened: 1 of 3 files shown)


flochtililoch (approved):
Would be good if we could use the values from the `Colors` service instead of hard coded values. Looks good otherwise, thanks for the fix!

ashwinibm:
@flochtililoch I'm merging this since `Colors` service is not available on this repo.

Issue #355, opened 2021-10-08
Requirement: support custom input elements


terryatgithub:
hi @adamstep & all, 
Long time no see, hope you're all going well.

We need to support some custom input fields like below. 
We found it hard to implement based on current conditions. 

---
#### What do we want to implement?
As you can see in the below picture, there are some input fields for money amount. 
when it's focused, need to show the primitive number: '1234567890.00', 
and when it's blurred, need to show the formatted number:  '1,234,567,890.00'. 
and when the length of the amount exceeds a certain length, need to show '11.11M' or '11.11B'.
(the field also needs to support i18n, where may show '1.234.567.890,00' in some country.)

![image](https://user-images.githubusercontent.com/12246394/136492750-684480ed-5e14-4ac8-8b4a-24510bd4502f.png)

---
#### Proposal
The method we think of is:
add a 'custom namespace logic' through an extension mechanism similar to behaviors.
https://github.com/Instawork/hyperview/blob/d09230293919911731238c1e36c985e1260a2131/src/index.js#L78
so we can add a custom RN input field component, and register it to our custom namespace. 
and with some changes in the below code, the customized elements can be submitted together.

https://github.com/Instawork/hyperview/blob/d09230293919911731238c1e36c985e1260a2131/src/services/index.js#L276
```javascript
  // TODO: It would be more flexible to grab any element with a name and value.
  FORM_NAMES
    // Get all inputs in the form
    .reduce((acc: Array<Element>, tag: LocalName) => {
      const inputElements: NodeList<Element> = formElement.getElementsByTagNameNS(
        Namespaces.HYPERVIEW, // <--- new customised namespace will be used here.
        tag,
      );
```
How do you think of this proposal, or is there a better way to support this feature?

Looking forward to your prompt response.
Regards.

terryatgithub:
I've verified this solution initially, and now try to prepare a PR.

terryatgithub:
hi @flochtililoch, could you please help with this issue? thank you.

terryatgithub:
hi @adamstep , glad to see your reply, hope you are all going well.

Your attempt seems good. 
The logic of extracting the value is scattered into each input element, therefore it offers more flexibility.
But I didn't see where and how we can register such custom input elements. 
Looking forward to your implementation.

> I like your approach with registering namespaces. Perhaps it should be extended to be more precise. Instead of registering namespaces, we have a registry of input elements (namespace + element name). That way, we only serialized data from elements specifically identified as input elements.

[...]

PR #359, opened 2021-10-29
Support custom form input components

adamstep:
Addresses #355 . Each input component (standard or custom) defines a static `getFormInputValues(element: Element): Array<[string, string]>` method, which is used to gather the form data.

- All standard custom elements have been moved to use this new API.
- `getFormData()` now simply iterates through the component registry, finding all components with a `getFormInputValues` property, and calls the method on the elements within the form to create the final `FormData`.
- Added tests for each `getFormInputValues()` implementation.


adamstep:
@flochtililoch I've been struggling with the Flow types here. It seems Flow doesn't like that I'm adding an optional static property to HvComponent here: https://github.com/Instawork/hyperview/pull/359/files#diff-e9726d469bba0bec8608c095ad5c361aeb95ea53afadd7ab44b462a2eb5e7467R304

As far as I know this is the right syntax, any idea what's going on?

flochtililoch:
> any idea what's going on?

Not really, I remember running into the same issue when adding the attribute `localNameAliases` to the `HvComponentStatics`, and the only way I found to work around the problem, was to make the property a non-optional, and to define it in every `HvComponent` classes. Not the most elegant, but likely something to do with flow being outdated and buggy.

adamstep:
> the only way I found to work around the problem, was to make the property a non-optional, and to define it in every `HvComponent` classes. Not the most elegant, but likely something to do with flow being outdated and buggy.

I see. I guess this won't affect custom components, right? Since we're not checking types at that level?

flochtililoch:
> this won't affect custom components, right?

I think that should be fine, I don't see the `localNameAliases` on most of our custom components

Issue #298, opened 2021-05-24
add "change" trigger for form inputs


adamstep:
There is currently no way for the UI to react to change events when the user interacts with and changes values of form inputs.

For reference, see the "Adjusted" label on the Pro Hours Dispute form: https://www.figma.com/file/iWanSoicP538uJZ6W9H5Mw/Pro-Hours-Dispute?node-id=687%3A7808

The view needs to be able to show the "Adjusted" modifier when the user selects a "dirty" value, and hide it if the user changes it back to the original / "clean" value.

This may require implementing separate multiple for inputs - for example, `inputDirty` and `inputClean`. It would probably also be useful to have `inputChange` which would fire regardless of how the value compares to the original.

hgray-instawork:
Need to implement this for `select-single` and `select-multiple`.

PR #360, opened 2021-11-23
Add support for "change" trigger on text-area and text-field

baptistegrimaud:
In addition to the `blur` and `focus` triggers, add support for `change` so we can for instance, update a part of the UI when the user types in a text field or text area.

Partial implementation of https://github.com/Instawork/hyperview/issues/298 but doesn't handle the dirty/clean states.

### DIFF
--- a/docs/reference_behavior_attributes.md
+++ b/docs/reference_behavior_attributes.md
@@ -27,9 +27,9 @@ The rest of the document describes in-depth how these attributes work together.
 
 ## trigger
 
-| Type                                                                                                              | Required |
-| ----------------------------------------------------------------------------------------------------------------- | -------- |
-| **press** (default), longPress, pressIn, pressOut, visible, refresh, load, select, deselect, focus, blur, on-event | No       |
+| Type                                                                                                                       | Required |
+| -------------------------------------------------------------------------------------------------------------------------- | -------- |
+| **press** (default), longPress, pressIn, pressOut, visible, refresh, load, select, deselect, focus, blur, change, on-event | No       |
 
 ### `press`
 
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hyperview",
-  "version": "0.45.0",
```
adamstep, line 3: You don't need to update this version number, we do this automatically when pushing a release to npm (by running `yarn release:minor`). But that should wait after this code is approved and merged.

baptistegrimaud: reverted

adamstep:
I think the implementation generally looks good, but I'd like to see some [examples added](https://github.com/Instawork/hyperview/tree/master/examples/behaviors) to the repo so we can test a few interactions. 

In particular, I want to verify the interaction of the `change` trigger with other attributes on the `<text-field>`, including `auto-focus` (should not trigger change), `mask` (if the change is not allowed, should we trigger?), and `secure-text` (change still triggers).

baptistegrimaud:
> In particular, I want to verify the interaction of the `change` trigger with other attributes on the `<text-field>`, including `auto-focus` (should not trigger change), `mask` (if the change is not allowed, should we trigger?), and `secure-text` (change still triggers).

Thanks @adamstep for bringing up those cases.
I added some examples to cover them (see below), let me know if additional ones are needed!

https://user-images.githubusercontent.com/91907891/143160525-dbb3cfcf-669a-480e-aacc-b6b9c5d41e9b.mp4

adamstep:
> PS: when a `mask` is provided, the `change` is still triggered, even if a non allowed character is entered. Is that acceptable or should I fix it?

This is what I was wondering about. I think the existing behavior is ok, but we should mention this interaction between `mask` and `change` trigger in the website docs.

Otherwise, the changes look good!

baptistegrimaud:
> This is what I was wondering about. I think the existing behavior is ok, but we should mention this interaction between `mask` and `change` trigger in the website docs.

@adamstep , I just added [this commit](https://github.com/Instawork/hyperview/pull/360/commits/3b8a14f686b6e3e98d39a838a260d2f7f21a2ae0), let me know if I should word it differently :)
Thanks

adamstep:
@baptistegrimaud I think it's good to use the same language as in the `mask` docs, maybe something like this:

> Note: when using the change trigger on a `<text-field>` that include the `mask` attribute, the behavior will be triggered even if the mask prevents the pressed key from being set.

adamstep (approved):
Nice work!

PR #365, opened 2021-12-14
Fix picker-field cancel on iOS

adamstep:
There's a history of issues with `<picker-field>` due to different implementations on iOS vs Android:
- https://github.com/Instawork/hyperview/pull/173
- https://github.com/Instawork/hyperview/pull/208
- https://github.com/Instawork/hyperview/pull/224

On Android, changes to values from the picker should apply immediately, there is no "Cancel".
On iOS, changes should not apply immediately because we've implemented our own modal.

Previous attempts to fix the issue have not been platform-aware. By checking the platform, we can use the right approach to fix issues on both platforms.


baptistegrimaud (approved):
LGTM!

PR #367, opened 2021-12-28
feat: Render XML server errors

flochtililoch:
Before throwing a `Errors.ServerError`, which results in rendering a webview with the error content when in dev mode, or a generic error message, verify if the content-type is `application/xml`. When it is the case, render the XML document like any other screens.


DanielSchaffer (approved):
Would need to be done for 4xx errors too?

flochtililoch:
> Would need to be done for 4xx errors too?

No, it looks like we were only filtering for 5XX, 4XX are already rendering XML responses.

PR #368, opened 2022-01-04
Feature: platform specific modifier; switch-state color modifier

lenwhite:
The demo switch element on Android was rendered in two colors as only `backgroundColor` was set for the switch. This did not match the rendering on iOS, where both the background and the toggle shares the same color.
|Platform|Switch|
|-|-|
| Android | image|
| iOS | image|

Currently it is not possible to have the same appearance on both Android and iOS as setting `thumbColor` affects the color of the toggle in both active and inactive states. 

E.g. the following...
```xml
      <style id="switch" backgroundColor="#E1E1E1" color="#4778FF">
        <modifier selected="true">
          <style backgroundColor="#4778FF" />
        </modifier>
      </style>
```

...will produce:
image

...instead of the visually clearer:
image

--- 

This PR allows the `thumbColor` to be changed depending on the state of the switch, which allows the Android appearance to match iOS. It defaults to the 'off' style of the switch when no 'on' style is set (which is the same as the current behavior).

In addition, when no `thumbColor` is specified, the `thumbColor` on android will default to a darker shade of the `backgroundColor` for a visually consistent appearance.

image

### DIFF
--- a/src/components/hv-switch/index.js
+++ b/src/components/hv-switch/index.js
@@ -61,6 +61,7 @@ export default class HvSwitch extends PureComponent<HvComponentProps> {
         newElement.setAttribute('value', value ? 'on' : 'off');
         this.props.onUpdate(null, 'swap', this.props.element, { newElement });
       },
+      // iOS thumbColor default
       thumbColor: unselectedStyle?.color || selectedStyle?.color,
       trackColor: {
         false: unselectedStyle ? unselectedStyle.backgroundColor : null,
@@ -69,10 +70,12 @@ export default class HvSwitch extends PureComponent<HvComponentProps> {
       value: this.props.element.getAttribute('value') === 'on',
     };
 
+    // android thumbColor default
     if (Platform.OS === 'android' && !props.thumbColor) {
-      props.thumbColor = props.value ? props.trackColor.true : '#FFFFFF';
+      props.thumbColor = props.value ? '#406be4' : '#FFFFFF';
     }
 
+    // if thumbColors are explicitly specified, override defaults
     if (props.value && selectedStyle?.color) {
       props.thumbColor = selectedStyle.color;
     } else if (!props.value && unselectedStyle?.color) {
--- a/src/components/hv-switch/index.js
+++ b/src/components/hv-switch/index.js
@@ -11,8 +11,8 @@
 import * as Dom from 'hyperview/src/services/dom';
 import * as Namespaces from 'hyperview/src/services/namespaces';
 import type { Element, HvComponentProps } from 'hyperview/src/types';
+import { Platform, StyleSheet, Switch } from 'react-native';
 import React, { PureComponent } from 'react';
-import { StyleSheet, Switch } from 'react-native';
 import {
   createStyleProp,
   getNameValueFormInputValues,
@@ -68,6 +68,10 @@ export default class HvSwitch extends PureComponent<HvComponentProps> {
       },
       value: this.props.element.getAttribute('value') === 'on',
     };
+
+    if (Platform.OS === 'android' && !props.thumbColor) {
+      props.thumbColor = props.value ? props.trackColor.true : '#FFFFFF';
+    }
 
     if (props.value && selectedStyle?.color) {
       props.thumbColor = selectedStyle.color;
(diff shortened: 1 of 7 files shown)


tr4n2uil:
Nice work @lenwhite

flochtililoch (commented):
I feel this kind of logic could be handled on the server side. Since we know on the server what the platform is, we can generate the styles accordingly, making this client handling not necessary. @adamstep any thoughts?

adamstep:
> I feel this kind of logic could be handled on the server side. Since we know on the server what the platform is, we can generate the styles accordingly, making this client handling not necessary. @adamstep any thoughts?

I agree, I don't recommend introducing the `platform` attribute on modifiers. We handle this kind of thing when generating the styles in `bootstrap.xml`. Since the user will not be switching platforms dynamically, we don't need to include both.

The rest of the PR looks good, but it would be nice to update the documentation on `<switch>` with information on how to apply the styles.

lenwhite:
From a library perspective though, would it be a useful functionality to have, in a separate PR? If part of Hyperview's promise is "You can even deploy your app as a collection of static XML files if you want!", then device specific modifiers seem like it would be necessary, since there's no way to make even demo look good on both platforms otherwise. 

An alternative approach I can think of is to set some sane defaults so that basic `<switch>` styling can be done: defaulting the `thumbColor` to be `#FFFFFF` on iOS always, and set it to `backgroundColor` on Android when 'on' and `#FFFFFF` on Android when 'off'. That way the styling would be at least be relatively consistent across both platforms when only `backgroundColor` is styled; documentation could then note that styling `thumbColor` needs to be platform-specific.

src/components/hv-switch/index.js
```diff
@@ -61,14 +61,24 @@ export default class HvSwitch extends PureComponent<HvComponentProps> {
...
       trackColor: {
         false: unselectedStyle ? unselectedStyle.backgroundColor : null,
         true: selectedStyle ? selectedStyle.backgroundColor : null,
       },
       value: this.props.element.getAttribute('value') === 'on',
     };
 
+    if (Platform.OS === 'android' && !props.thumbColor) {
+      props.thumbColor = props.value ? props.trackColor.true : '#FFFFFF';
+    }
```
flochtililoch, line 74: Shouldn't this logic come after the two if conditions that set `props.thumbColor` below?

flochtililoch (approved):
one minor question for sanity check, looks good otherwise!

ratneshpkn (changes requested):
IMO, one issue (visually) with applying the same color is that it makes the whole switch component become a blob of color. If we're handling it in a platform specific way, it might be better to handle the visuals to match the platform components.

For reference:
**Android**
Android Switch

**iOS**
iOS Switch

src/components/hv-switch/index.js
lenwhite: * By default (above), on iOS, we leave `thumbColor` as `undefined` 
* Otherwise (here) we defaulting the `thumbColor` to the `trackColor` on android when selected, and white when not selected, to match the normal appearance of Android switches
* Then the two `if` conditions afterwards allows this two defaults to be overridden if set specifically.

Let me know if that makes sense.

flochtililoch: Got it, makes sense

lenwhite:
@ratneshpkn currently, we're only changing the default `thumbColor` on android if a `backgroundColor` is specified, otherwise it'll look identical to the platform default.

The issue is that Android switch can change color based on the theme/skin set (I'm not sure if this is true for iOS as well in some way). So there's no way to exactly match platform defaults. As far as I can tell, different theme/skins have different amounts of contrast between the `backgroundColor` and the `thumbColor`. 

I tried programmatically darkening the `backgroundColor` provided but if a dark `backgroundColor` is used (as ours is) and it sort of works but I'm not certain if it's a good idea to generalize to all colors, see this example:

image
image
image

Let me know if this is the approach we want to take

adamstep:
> From a library perspective though, would it be a useful functionality to have, in a separate PR? If part of Hyperview's promise is "You can even deploy your app as a collection of static XML files if you want!", then device specific modifiers seem like it would be necessary, since there's no way to make even demo look good on both platforms otherwise.

It's a good point, I can see adding this functionality in a separate PR for the reasons you stated. But if we add it, I want to make sure we think through several potential use cases to make sure the API is well-designed. For example, would we want to have platform specific checks for entire styles or only modifiers? Do we need to support platform checks in other parts of the UI?

ratneshpkn (approved):
lgtm

PR #376, opened 2022-01-22
fix: warning about `<items>` not registered

flochtililoch:
Lists can use the `<items>` element to wrap several items when paginating. This tag does not get rendered and is only used so that generated responses are valid XML. Because it is not registered as a component, it triggers a warning when inserted into the DOM. Add it as an alias for the `view` element to get rid of the warning.

Simulator Screen Shot - iPhone 8 - 2022-01-22 at 08 28 50

### DIFF
--- a/src/components/hv-view/index.js
+++ b/src/components/hv-view/index.js
@@ -35,6 +35,7 @@ export default class HvView extends PureComponent<HvComponentProps> {
     LOCAL_NAME.FORM,
     LOCAL_NAME.HEADER,
     LOCAL_NAME.ITEM,
+    LOCAL_NAME.ITEMS,
     LOCAL_NAME.SECTION_TITLE,
   ];
 
--- a/src/services/components/index.test.js
+++ b/src/services/components/index.test.js
@@ -33,6 +33,7 @@ const defaultRegistryContent = {
   header: HvView,
   image: HvImage,
   item: HvView,
+  items: HvView,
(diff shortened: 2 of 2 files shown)


adamstep (commented):
Does this mean that <items> will render as a view in some conditions? Because that would not be desired. Maybe there’s a more robust way to register certain elements to not be rendered?

flochtililoch:
> Does this mean that will render as a view in some conditions? Because that would not be desired.

It would render a view if the element `<items>` was used outside of the context of a list, but that's also the case for the element `<item>`

flochtililoch:
> Maybe there’s a more robust way to register certain elements to not be rendered?

Do you know if there are other candidate elements that would benefit from this system?

adamstep:
> > Maybe there’s a more robust way to register certain elements to not be rendered?
> 
> Do you know if there are other candidate elements that would benefit from this system?

None come to mind at the moment. But going back to my assumption, will something bad happen if we try to render `<items>` as a view within a list? Or does it not matter, since `<list>` doesn't render arbitrary children?

flochtililoch:
That's right, it won't matter as the list components are not rendering arbitrary children

adamstep (approved):

adamstep:
Ok, let's go with this solution for now. Maybe leave a comment explaining that this alias is a no-op, just to avoid a warning.

PR #378, opened 2022-02-05
feat: Do not crash when text string not wrapped in <text>

flochtililoch:
Prior to rendering a text node, verify its parent component is a `<text>`, and log a warning instead of lefting the RN app crash.

| Before | After |
|--------|------|
| Simulator Screen Shot - iPhone 8 - 2022-02-04 at 20 55 23 | Simulator Screen Shot - iPhone 8 - 2022-02-04 at 20 52 03 |

### DIFF
--- a/src/services/render/index.js
+++ b/src/services/render/index.js
@@ -96,9 +96,20 @@ export const renderElement = (
   }
 
   if (element.nodeType === NODE_TYPE.TEXT_NODE) {
-    // Render non-empty text nodes
-    if (element.nodeValue && element.nodeValue.trim().length > 0) {
-      return element.nodeValue.trim().replace(/\s+/g, ' ');
+    // Render non-empty text nodes, when wrapped inside a <text> element
+    if (element.nodeValue) {
+      const trimmedValue = element.nodeValue.trim();
+      if (trimmedValue.length > 0) {
+        if (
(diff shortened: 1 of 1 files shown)


src/services/render/index.js
```diff
@@ -96,9 +96,20 @@ export const renderElement = (
...
   if (element.nodeType === NODE_TYPE.TEXT_NODE) {
-    // Render non-empty text nodes
-    if (element.nodeValue && element.nodeValue.trim().length > 0) {
-      return element.nodeValue.trim().replace(/\s+/g, ' ');
+    // Render non-empty text nodes, when wrapped inside a <text> element
+    if (element.nodeValue) {
+      const trimmedValue = element.nodeValue.trim();
+      if (trimmedValue.length > 0) {
+        if (
+          element.parentNode &&
```
adamstep, line 104: To be explicit, we should also check the namespace in addition to the tag local name.

flochtililoch: Good catch - addressed in [24e5e7c](https://github.com/Instawork/hyperview/pull/378/commits/24e5e7cd08a268f9911c0db99f3b73739b9aade0. I'm also adding a check to allow components with namespaces other than Hyperview's to render text node.

Issue #380, opened 2022-02-07
Support sticky header for scroll view


palaparthi:
React native supports [stickyHeaderIndices](https://reactnative.dev/docs/scrollview#stickyheaderindices) for scroll view.

When `<view scroll="true">` we can support multiple `<sticky-header target="..." />` as children where `target` attribute references the view we want to make sticky.

Example
```
        <view scroll="true">
          <sticky-header target="view-1" />
          <sticky-header target="view-3" />
          <view id="view-0">
            <text style="Text">Non Sticky</text>
          </view>
          <view id="view-1">
            <text style="Text">Sticky</text>
          </view>
          <view id="view-2">
            <text style="Text">Non Sticky</text>
          </view>
          <view id="view-3">
            <text style="Text">Sticky</text>
          </view>
        </view>
```

palaparthi:
@flochtililoch @adamstep Any thoughts?

flochtililoch:
My immediate thought on the markup: it might be cleaner if the views would declare themselves as "sticky". Internally, the scroll view could re-compute the `stickyHeaderIndices` array as more child views are loaded.

```
        <view scroll="true">
          <view id="view-0">
            <text style="Text">Non Sticky</text>
          </view>
          <view id="view-1" sticky="true">
            <text style="Text">Sticky</text>
          </view>
          <view id="view-2">
            <text style="Text">Non Sticky</text>
          </view>
          <view id="view-3" sticky="true">
            <text style="Text">Sticky</text>
          </view>
        </view>
```

palaparthi:
@adamstep I tried that, for my usecase I don't think its ideal to use `<section-list>`.  If you look at the screenshot I want to make only the tabs part sticky. For this to make it work with `<section-list>` I would need to put the tabs in `<section-title>` and the content in contacts tab would be an `<item>` which again has a `<list>` of contacts. There is prob a way to refactor everything to make it work with `<section-list>`. But imo this solution will be cleaner.

@flochtililoch that makes sense, We can make it clear in the documentation that only direct children of `<view scroll="true">` can have ` sticky="true"`.

![Simulator Screen Shot - iPhone 11 - 2022-02-07 at 12 34 22](https://user-images.githubusercontent.com/28518512/152841446-cf3243c5-db0b-4c84-8008-191b2e62e7b5.png)

flochtililoch:
[...]

PR #381, opened 2022-02-08
feat(scroll-view): add sticky attribute

palaparthi:
Add support for sticky header in scroll view, addresses #380.

### Single Sticky header

https://user-images.githubusercontent.com/28518512/152913944-aeda5741-4120-4cb5-aa2a-73168cd5eb90.mp4

### Multiple sticky headers

https://user-images.githubusercontent.com/28518512/152913998-9e4a6039-ce6c-4c1b-846f-83583402e836.mp4

Note: Flow wasn't happy with `Array.from(`, so had to fix ignore them, lmk if there is a better way to handle this.

### DIFF
--- a/src/components/hv-view/index.js
+++ b/src/components/hv-view/index.js
@@ -52,9 +52,11 @@ export default class HvView extends PureComponent<HvComponentProps> {
   canRenderElement = (element: Element): boolean => {
     // ignore non rendering and hidden elements
     return (
-      this.props.options.componentRegistry &&
-      this.props.options.componentRegistry[element.namespaceURI] &&
-      this.props.options.componentRegistry[element.namespaceURI][
+      !!this.props.options.componentRegistry &&
+      !!element.namespaceURI &&
+      !!element.localName &&
+      !!this.props.options.componentRegistry[element.namespaceURI] &&
+      !!this.props.options.componentRegistry[element.namespaceURI][
         element.localName
       ] &&
       // $FlowFixMe
--- a/src/components/hv-view/index.js
+++ b/src/components/hv-view/index.js
@@ -42,7 +42,7 @@ export default class HvView extends PureComponent<HvComponentProps> {
 
   getStickyElementIndices = (elements: Element[]): number[] => {
     return elements.reduce((acc, element, index) => {
-      if (element.getAttribute('sticky') === 'true') {
+      if (element?.getAttribute('sticky') === 'true') {
         acc.push(index);
       }
       return acc;
@@ -140,6 +140,7 @@ export default class HvView extends PureComponent<HvComponentProps> {
 
     if (scrollable && scrollDirection !== 'horizontal') {
       // add sticky indicies
+      // $FlowFixMe
       const elements = children.map(child => child.props.element);
       const stickyIndicies = this.getStickyElementIndices(elements);
       if (stickyIndicies.length) {
--- a/src/components/hv-view/index.js
+++ b/src/components/hv-view/index.js
@@ -18,7 +18,7 @@ import {
   View,
 } from 'react-native';
 import React, { PureComponent } from 'react';
-import type { HvComponentProps } from 'hyperview/src/types';
+import type { HvComponentProps, NODE_TYPE } from 'hyperview/src/types';
 import type { InternalProps } from './types';
 import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scrollview';
 import { LOCAL_NAME } from 'hyperview/src/types';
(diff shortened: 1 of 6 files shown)


src/components/hv-view/index.js
```diff
@@ -40,6 +40,69 @@ export default class HvView extends PureComponent<HvComponentProps> {
...
+    for (let j = 0; j < attributes.length; j += 1) {
+      if (attributes[j].name === 'hide' && attributes[j].value === 'true') {
+        return true;
+      }
+    }
+    return false;
+  };
+
+  canRenderElement = (element: Element): boolean => {
+    const nonRenderingElements = [
```
adamstep, line 53: hard-coding the non-rendering elements here is not very future-proof. If someone adds support for a new non-rendering element, they won't necessarily know to update `HvView`.

A more future-proof approach would be to check for the rendering elements in the component registry. As part of the `options` prop passed to `HvView`, there is an attributed called `componentRegistry` ([link](componentRegistry)). You can look up elements in `componentRegistry` (similar to what we do [here](https://github.com/Instawork/hyperview/blob/master/src/services/render/index.js#L64). I have a future PR that cleans up this access with a helper method).

src/components/hv-view/index.js
```diff
@@ -40,6 +40,69 @@ export default class HvView extends PureComponent<HvComponentProps> {
...
+    );
+  };
+
+  getElementsToRender = (): Element[] => {
+    // do not include non rendering elements while calculating index
+    // $FlowFixMe
+    const childNodes: Element[] = Array.from(
+      this.props.element.childNodes ?? [],
+    );
+    return childNodes.reduce((acc, element) => {
```
adamstep, line 76: Does `filter()` make more sense than `reduce` here?

src/components/hv-view/index.js
```diff
@@ -40,6 +40,53 @@ export default class HvView extends PureComponent<HvComponentProps> {
 
   props: HvComponentProps;
 
+  isHiddenElement = (attributes: Attribute[]): boolean => {
```
adamstep, line 43: curious, why are you passing the attributes rather than the entire `Element`? Then you can do `return element.getAttribute("name") === "true"`

src/components/hv-view/index.js
```diff
@@ -40,6 +40,53 @@ export default class HvView extends PureComponent<HvComponentProps> {
...
+    );
+    return childNodes.filter(element => this.canRenderElement(element));
+  };
+
+  getStickyElementIndices = (): number[] => {
+    const elements = this.getElementsToRender();
+    return elements.reduce((acc, element, index) => {
+      // $FlowFixMe
+      const attributes: Attribute[] = Array.from(element.attributes ?? []);
+      attributes.forEach(attribute => {
```
adamstep, line 81: Rather than iterating through attributes, can you use `element.getAttribute("sticky") === "true"`?

src/components/hv-view/index.js
```diff
@@ -40,6 +40,53 @@ export default class HvView extends PureComponent<HvComponentProps> {
...
+  canRenderElement = (element: Element): boolean => {
+    // ignore non rendering and hidden elements
+    return (
+      !!this.props.options.componentRegistry &&
+      !!element.namespaceURI &&
+      !!element.localName &&
+      !!this.props.options.componentRegistry[element.namespaceURI] &&
+      !!this.props.options.componentRegistry[element.namespaceURI][
+        element.localName
+      ] &&
```
flochtililoch, line 61: Are we assuming that if the component is part of the registry, it is rendering some content? I'm not sure that's always true/verifiable without actually getting the component to render.
Is there a risk in assuming that every child of a scroll view that has the `sticky` property will be a header that we want to render? I think this could simplify the logic quite a bit here. I might be missing something though on why we need to check first for non rendering elements.

palaparthi: > Are we assuming that if the component is part of the registry, it is rendering some content

Yes, that is my understanding too. I manually tested it by adding non rendering elements like `<behavior>` as a child to sticky view to see if it's true. The only case where this is not true is when `hide="true"` hence I am specifically checking for hidden elements too.

> Is there a risk in assuming that every child of a scroll view that has the sticky property will be a header that we want to render?

I don't think so. If we add `sticky` to a non rendering element then it will be safely ignored.

> I might be missing something though on why we need to check first for non rendering elements

For example
```
  <view scroll="true">
    <view id="view-0">
      <text style="Text">Non Sticky</text>
    </view>
    <behavior />
    <view id="view-1" sticky="true">
      <text style="Text">Sticky</text>
    </view>
    <view id="view-2">
      <text style="Text">Non Sticky</text>
    </view>
  </view>
```
In this case when we want to pass `stickyHeaderIndices=[1]` as the prop to react, however if we don't filter out non rendering elements we would be passing `stickyHeaderIndices=[2]` because `<behavior />` will be counted too but when we render we ignore `<behavior />`. Thus instead of `view-1` being sticky `view-2` will become sticky which is not our intention.

src/components/hv-view/index.js
palaparthi: Same reason as below. Makes sense, updating to pass an element instead.

src/components/hv-view/index.js
```diff
@@ -112,6 +175,13 @@ export default class HvView extends PureComponent<HvComponentProps> {
       if (scrollDirection === 'horizontal') {
         props.horizontal = true;
       }
+
+      if (scrollDirection !== 'horizontal') {
```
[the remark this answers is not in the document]
palaparthi: My intention here in adding an additional check is to have separate branching block for each prop we add to make it more readable. lmk if you still want me to commit this.

src/components/hv-view/index.js
flochtililoch: I see - I wonder if instead of iterating over the elements in the Hyperview DOM, we should iterate over those in the React DOM, that is the list of components returned by the call to `Render.renderChildren` method in `render()`. This would give us a more accurate reflection of the indexes we need to mark as sticky.
Each React component returned by `Render.renderChildren` should in theory have a reference it's Hyperview element. Let me know what you think.

src/components/hv-view/index.js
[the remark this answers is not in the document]
flochtililoch: I think that's fine - I felt the if/else was more readable but I'll defer to your judgment here.

adamstep:
Looks good to me, I'll let @flochtililoch sign off once his questions are resolved.

src/components/hv-view/index.js
palaparthi: That's pretty neat, updated.

src/components/hv-view/index.js
```diff
@@ -122,17 +131,25 @@ export default class HvView extends PureComponent<HvComponentProps> {
...
+    const children = Render.renderChildren(
+      this.props.element,
+      this.props.stylesheets,
+      this.props.onUpdate,
+      viewOptions,
     );
+
+    if (scrollable && scrollDirection !== 'horizontal') {
+      // add sticky indicies
+      // $FlowFixMe
```
flochtililoch, line 143: `child` might be a `string` or `any` type, which is why you're getting a flow error - you need to check for the type and presence of `props` and `element`.
Looking back at the evolution of the change, it's now possible to simplify by iterating over the list of children and generating the indices in a single pass: the array reduce can be done over the `children` list directly. Note also the typo in "indices"

```js
const stickyIndices = children.reduce(
  (acc, element, index) =>
    typeof element !== 'string' &&
    element.props?.element?.getAttribute('sticky') === 'true'
      ? [...acc, index]
      : acc,
  [],
);

palaparthi: Gotcha. Curious, in which cases would a `child` be a `string`?

flochtililoch: If you drill down into `Render.renderChildren`, which internally calls `renderElement`, you can see the code path that could lead to rendering a string. Specifically, this would be these `return` statement: [1](https://github.com/Instawork/hyperview/blob/master/src/services/render/index.js#L108), [2](https://github.com/Instawork/hyperview/blob/master/src/services/render/index.js#L118)

PR #395, opened 2022-03-07
Switch example site to Eleventy (static site generator)

adamstep:
As I've been working on client-side validation (https://github.com/Instawork/hyperview/discussions/332), I'm finding the existing `examples/` directory to be quite limiting. Each file has hard-coded links, styles, etc. This results in a lot of boilerplate and inconsistencies. It's worth replacing it with a system that can let us re-use templates, reduce boilerplate, and generally be more efficient.

In this PR, I'm switching out the old `http-server` with [Eleventy](https://www.11ty.dev/), a static-site generator. Eleventy supports the features we need:
- works with many template formats, including [Nunjucks](https://mozilla.github.io/nunjucks/) (similar to Django templates)
- can build a static site, or run as a development server
- supports tagging/collections to automatically build up the directory screens in the example app

#### Implementation details:
- Command to run the server is the same, `yarn test:xmlserver`
- I've ported a few screens to use Nunjucks templates (main screen, top-level screens, and navigation screens)
- The rest of the screens work on passthrough, so all features and screens continue to work
- I've created a few base templates/partials to reduce duplication:
    - `examples/_includes/base.xml` has the basic screen structure and styles
    - `examples/_includes/styles.xml` has the shared styles across all screens
    - `examples/_includes/directory_list.xml` can automatically build up a list of links given a tag name.

For example, to generate the list of "Navigation" examples, the file now looks like this:
```xml
---
permalink: "/navigation/index.xml"
title: "Navigation"
tags: root
list_tag: "navigation"
---
{% extends 'directory_list.xml' %}
```

- `title` and `tags` are used to build up the parent screen.
- `list_tag` tells Eleventy to render a collection of links of screens that contain the "navigation" tag.
- When we add a new screen tagged as "navigation", it will immediately appear in this list.


examples/advanced_behaviors/index.xml.njk
```diff
@@ -1,3 +1,8 @@
+---
+title: "Advanced Behaviors"
+permalink: "/advanced_behaviors/index.xml"
+tags: root
+---
```
flochtililoch, line 5: Instead of renaming the xml files, could we just include them? That would allow us to continue linting the XML markup

```suggestion
---

{% include 'examples/advanced_behaviors/index.xml' %}
```

adamstep: But the xml files may no longer be valid, since they will contain template syntax.

A better approach may be to run linting/validation on the generated XML files. We could check those into the repo.

flochtililoch: Can templates have the `.xml` extension, or does the `.xml.njk` extension is required for parsing to work?

flochtililoch: Linting post-processing can become a headache depending on how complex templates becomes, particularly when nesting multiple template statements. That might be a non problem for this project but just wanted to raise this.

adamstep: > Can templates have the `.xml` extension, or does the `.xml.njk` extension is required for parsing to work?

The eleventy project detects which template handler to used based on the extension, so that's why `.njk` is included. I looked into it, it's possible to add a new extension, but I'm not sure it's worth it.

Regarding linting, it maybe matters less if we're not writing a lot of boilerplate.

flochtililoch (approved):

Issue #391, opened 2022-02-23
Demo does not work on latest master (v0.49.0)


benjaminoakes:
Hi all!

[I'm new here](https://github.com/Instawork/hyperview/pull/390), so apologies if I'm just dense, but it appears that the demo just doesn't work on the most recent revision (19fe26fb77d943205634002ddc0cbb09dea90cb4):

```
commit 19fe26fb77d943205634002ddc0cbb09dea90cb4 (HEAD -> master, tag: v0.49.0, origin/master, origin/HEAD)
Author: flochtililoch <flochtililoch@gmail.com>
Date:   Thu Feb 10 10:41:40 2022 -0800

    v0.49.0
```

I got `yarn start` running and connected to Expo, but I get the below errors when doing anything.  Any help would be appreciated.  I'd love to play around with Hyperview and see what it's like.  Thank you for your assistance!

On web:

```
Web Bundling complete 6738ms
./node_modules/react-native/Libraries/StyleSheet/normalizeColor.js:24:35
Module not found: Can't resolve './PlatformColorValueTypes'
  22 | ): ?ProcessedColorValue {
  23 |   if (typeof color === 'object' && color != null) {
> 24 |     const {normalizeColorObject} = require('./PlatformColorValueTypes');
     |                                   ^
  25 |     const normalizedColor = normalizeColorObject(color);
  26 |     if (normalizedColor != null) {
  27 |       return color;
```

On iOS:

```
iOS Bundling complete 1986ms
iOS Running app on iPhone 13

TypeError: undefined is not an object (evaluating '_this$getNavigationSt2.params')

This error is located at:
    in HyperScreen (created by HyperviewScreen)
    in Unknown (created by HyperviewScreen)
    in HyperviewScreen (created by SceneView)
    in StaticContainer
    in EnsureSingleNavigator (created by SceneView)
    in SceneView (created by CardContainer)
    in RCTView (created by View)
    in View (created by CardContainer)
    in RCTView (created by View)
    in View (created by CardContainer)
    in RCTView (created by View)
    in View
    in CardSheet (created by Card)
    in RCTView (created by View)
    in View (created by AnimatedComponent)
    in AnimatedComponent
    in AnimatedComponentWrapper (created by PanGestureHandler)
    in PanGestureHandler (created by PanGestureHandler)
    in PanGestureHandler (created by Card)
    in RCTView (created by View)
    in View (created by AnimatedComponent)
    in AnimatedComponent
[...]

PR #398, opened 2022-03-14
fix: Demo app Hyperview version

flochtililoch:
Follow up for #385: update the Hyperview version reference in the demo app.


adamstep (approved):

flochtililoch:
Fixes #391

PR #399, opened 2022-03-15
chore: Deprecate `<text-area>` element

flochtililoch:
`<text-area>` and `<text-field>` are very close in nature, in fact they shared mostly the same code and are backed by the same underlying `TextInput` RN component. This change centralizes the logic used into `components/hv-text-field` by making `text-area` an alias of `text-field`. When the local name used is `text-area`, the `multiline` property is set automatically.
It also formally deprecates usage of `<text-area>`.


adamstep:
@flochtililoch are there any features supported by `<text-field>` that weren't supported by `<text-area>`? I can think of one (`mask`), but maybe there's others. We should just test the combination of mask and multiline to make sure it works as expected.

flochtililoch:
> it works as expected

I would assume it's not expected to work, since it's not documented as formal usage. I think it should work though, I can verify a bit later.

adamstep:
> I would assume it's not expected to work, since it's not documented as formal usage. I think it should work though, I can verify a bit later.

`mask` is documented as formal usage: https://hyperview.org/docs/reference_textfield#mask

flochtililoch:
I meant, `mask` is not meant to be used along with `text-area`

adamstep:
Ah right, I think we didn't add it to `<text-area>` because there wasn't a clear use case like there is for things like SSNs, phone numbers, etc.

adamstep (approved):

flochtililoch:
@adamstep I tested the mask functionality with `<text-field multiline="true">`, it work well. In my last two commits, I updated the examples and fixed the schema, can you take another look?

adamstep (approved):
example updates look good

adamstep:
thanks @flochtililoch !

PR #414, opened 2022-04-17
feat(hv-view): `content-container-style` attribute

flochtililoch:
Add support for `content-container-style` attribute, that allows setting the property [`contentContainerStyle`](https://reactnative.dev/docs/scrollview#contentcontainerstyle) of RN `ScrollView`


adamstep (approved):
I thought we already had this, but I don't see any references in the code. Good addition!

PR #412, opened 2022-04-16
feat: Custom loading components

flochtililoch:
Allow passing props to define loading, load error and list refresh control components.

Implementation notes:
- each component is passed as optional props of `HyperScreen`
- `refreshControl` prop is exposed to `HvList` and `HvSectionList` via a React context

Follow up for #330
Resolves #286

PR #417, opened 2022-04-19
fix: Refresh control on Android

flochtililoch:
#412 introduced a bug with Android, where refresh control would not render correctly nor be properly removed once refreshing done. From my testing, it seems that nesting the `RefreshControl` under another component when setting `refreshControl` prop is causing this issue. Removing that extra nesting solves the issue.


adamstep (approved):
weird bug, but glad you found a fix.

PR #420, opened 2022-05-01
chore: Schema clean up

flochtililoch:
What this PR does:
- [39fe192](https://github.com/Instawork/hyperview/pull/420/commits/39fe1923540b641a32e9744a9773f2599b6e156a) removes Instawork custom schemas
- [48f7d5d](https://github.com/Instawork/hyperview/pull/420/commits/48f7d5d82f83eaf9b961d56fecc4061afa94dc69) renames existing `hyperview.xsd` to `core.xsd` and remove enumeration for `action` type, to allow redefinition in other schemas, as suggested [here](http://lists.xml.org/archives/xml-dev/200505/msg00003.html). That feels hacky, but I couldn't find a better way to redefine enumeration of a restricted type.
- [5efd04a](https://github.com/Instawork/hyperview/pull/420/commits/5efd04a34dacdf3f0e5f2d2e2dc0e66c8297ad77) re-creates `hyperview.xsd` that imports `core.xsd` and redefines the `action` type with allowed values 
- [c3fe3e3](https://github.com/Instawork/hyperview/pull/420/commits/c3fe3e3b109873b98723e9078ed68aa3a7fc496d) exclude examples that use Instawork's extensions from XML validation - should we remove these examples entirely?


adamstep:
Looks good, I think we probably should make the custom examples use base Hyperview with no extensions so we don't need to exclude those files. In any case, those examples use some outdated conventions, it would be nice to update them.

adamstep (approved):

PR #434, opened 2022-06-16
feat:Add set-value support to <select-multiple>

baptistegrimaud:
Add support for [set-value behavior attribute](https://hyperview.org/docs/reference_behavior_attributes#set-value) to the [`<select-multiple>` element](https://hyperview.org/docs/reference_selectmultiple).

### DIFF
--- a/src/behaviors/hv-set-value/index.js
+++ b/src/behaviors/hv-set-value/index.js
@@ -1,7 +1,6 @@
 // @flow
 
 import * as Behaviors from 'hyperview/src/services/behaviors';
-import * as Namespaces from 'hyperview/src/services/namespaces';
 import * as Xml from 'hyperview/src/services/xml';
 import type {
   DOMString,
@@ -12,7 +11,6 @@ import type {
   HvUpdateRoot,
 } from 'hyperview/src/types';
 import { later, shallowCloneToRoot } from 'hyperview/src/services';
-import { LOCAL_NAME } from 'hyperview/src/types';
 
 export default {
   action: 'set-value',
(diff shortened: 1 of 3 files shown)


src/behaviors/hv-set-value/index.js
```diff
@@ -48,6 +50,18 @@ export default {
 
       // Show the target
       targetElement.setAttribute('value', newValue);
+      // select the <option> matching the "newValue" and unselect the others
+      const options = targetElement.getElementsByTagNameNS(
+        Namespaces.HYPERVIEW,
+        LOCAL_NAME.OPTION,
+      );
+      for (let i = 0; i < options.length; i += 1) {
```
adamstep, line 58: If I understand correctly, this approach will result in just one option being selected, the other options will be deselected. Is that desired in the case of `<select-multiple>`?

adamstep: Also, over time we tried to detangle behaviors with specific implementations of other components. Introducing this code here means we are making some assumptions about how select single/multiple work under the hood.

I'd prefer if select-multiple knows how to handle the situation where `value` is set as an attr from this line:
`targetElement.setAttribute('value', newValue);`

You can see how select-single handles it here: https://github.com/Instawork/hyperview/blob/master/src/components/hv-select-single/index.js#L55

baptistegrimaud: Do you think I should add a new `"set-values"` attribute where multiple values can be specified (maybe comma-separated values)? Which would be available only for `<select-multiple>`.
Otherwise the `"set-value"` attribute of `<select-multiple>` would mostly be used with an empty string, to unselect all the options... which made me think: should we create a new `unselect` behavior action that would be supported by `<select-single>`, `<select-multiple>` and `<picker-field>`?

baptistegrimaud:
Replaced by https://github.com/Instawork/hyperview/pull/436

PR #436, opened 2022-06-20
feat:Create select-all/unselect-all behavior actions

baptistegrimaud:
Create `"select-all"` and `"unselect-all"` behavior actions.

https://user-images.githubusercontent.com/91907891/174645840-738bc378-fcfa-4fca-844f-e6576357a141.mp4

https://user-images.githubusercontent.com/91907891/174645831-6efcfd84-c78a-460c-b213-4b24588f0125.mp4

*NB*: while `"select-all"` is only supported by [`<select-multiple>`](https://hyperview.org/docs/reference_selectmultiple), `"unselect-all"` is supported by both [`<select-multiple>`](https://hyperview.org/docs/reference_selectmultiple) and [`<select-single>`](https://hyperview.org/docs/reference_selectsingle).

### DIFF
--- a/docs/reference_behavior_attributes.md
+++ b/docs/reference_behavior_attributes.md
@@ -342,9 +342,9 @@ The `verb` attribute defines the HTTP method used to request the content specifi
 
 ## action
 
-| Type                                                                                                                                                                                        | Required |
-| ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
-| **push** (default), new, back, close, navigate, deep-link, open-settings, replace, replace-inner, append, prepend, reload, dispatch-event, show, hide, toggle, set-value, copy-to-clipboard | No       |
+| Type                                                                                                                                                                                                                  | Required |
+| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------- |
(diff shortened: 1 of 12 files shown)


adamstep:
@baptistegrimaud does this still have the effect where it looks like the options are not all selected/unselected at the same time? The video makes it look like the change happens one-by-one, which is a bit surprising to me given your implementation. Since you're swapping in the new element after changing all the attributes, I expect the change to be reflected synchronously.

baptistegrimaud:
> @baptistegrimaud does this still have the effect where it looks like the options are not all selected/unselected at the same time? The video makes it look like the change happens one-by-one, which is a bit surprising to me given your implementation. Since you're swapping in the new element after changing all the attributes, I expect the change to be reflected synchronously.

Hi @adamstep,
It looks like the "flicker" issue only happens in the video, it looks normal to me in the simulator.
The select-all/unselect-all actions update all the options at once, when using multiple options (not just 2 like in my initial examples):

https://user-images.githubusercontent.com/91907891/174660413-b42f04c2-0b35-476b-b68b-50ca8b87f03b.mp4

https://user-images.githubusercontent.com/91907891/174660417-b9b25414-d034-494a-88e6-5ffee43e696d.mp4

adamstep (approved):

PR #441, opened 2022-06-24
refactor(schema): Make extending behavior schema possible

flochtililoch:
Make `behavior` type named, so that schema extensions can redefine it.

Same change as #438
Requires https://github.com/Instawork/instawork/pull/18568 before merging

### DIFF
--- a/schema/core.xsd
+++ b/schema/core.xsd
@@ -762,14 +762,13 @@
     </xs:complexType>
   </xs:element>
 
-  <xs:element name="behavior">
-    <xs:complexType>
-      <xs:sequence minOccurs="0" maxOccurs="unbounded">
-        <xs:element ref="alert:option" />
-      </xs:sequence>
-      <xs:attributeGroup ref="hv:behaviorAttributes" />
-    </xs:complexType>
-  </xs:element>
+  <xs:complexType name="behavior">
+    <xs:sequence>
+      <xs:element minOccurs="0" maxOccurs="unbounded" ref="alert:option" />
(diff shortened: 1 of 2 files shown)


schema/core.xsd
```diff
@@ -762,14 +762,13 @@
...
-      <xs:sequence minOccurs="0" maxOccurs="unbounded">
-        <xs:element ref="alert:option" />
-      </xs:sequence>
-      <xs:attributeGroup ref="hv:behaviorAttributes" />
-    </xs:complexType>
-  </xs:element>
+  <xs:complexType name="behavior">
+    <xs:sequence>
+      <xs:element minOccurs="0" maxOccurs="unbounded" ref="alert:option" />
+      <xs:element minOccurs="0" maxOccurs="unbounded" ref="hv:styles" />
```
baptistegrimaud, line 768: So that means `<behavior>` would now support `<styles>`? 🤔 
What would be the scenario exactly and what would be the expected behaviour in this case?

flochtililoch: Actually, thanks for catching this - I did not intend to include this line as part of the core schema (removed in [92a83fb](https://github.com/Instawork/hyperview/pull/441/commits/92a83fbaa30f1ec2a43038dc86edb06cf83bd22b))
The intention with this PR is only to make the `behavior` type extensible. Our custom extension then does allows the `styles` element to be nested under a `behavior` element. I'll tag you on relevant PRs for more details.

baptistegrimaud (approved):

PR #447, opened 2022-07-08
Revert "fix: Nested text spaces"

flochtililoch:
Reverts Instawork/hyperview#444

This change has a few issues:
- the warning isn't properly filtering out text nodes with empty strings (i.e. line breaks in the markup) and multiple warnings are being logged every time a screen renders
- the space handling is working fine for the given example (nested texts), but does not work as expected when there are multiple siblings `<text>` element nested under a common `<text>`.

I'm working on an alternative implementation to re-introduce this fix.

### DIFF
--- a/src/services/dom/helpers.js
+++ b/src/services/dom/helpers.js
@@ -15,9 +15,7 @@ import type {
   HvComponentOnUpdate,
   LocalName,
   NamespaceURI,
-  Node,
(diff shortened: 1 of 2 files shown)


adamstep (approved):
When you add this back, can you include unit tests to make sure the functionality works for all edge cases?

PR #448, opened 2022-07-10
fix: Nested text spaces

flochtililoch:
Create inline formatting context when rendering a `<text>` element. Traverse the element tree, create a list of text node leaves, ordered as they appear on screen, then apply logic to collide/trim spaces (explanation of logic [here](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace#how_does_css_process_whitespace#explanation)).

**Demo:**
| Before | After |
|-------|-------|
| Simulator Screen Shot - iPhone 13 mini - 2022-07-10 at 18 16 38 | Simulator Screen Shot - iPhone 13 mini - 2022-07-10 at 18 22 46 |

**Note:** Unlike HTML, siblings `<text>` elements that are not nested under a parent text won't apply the inline formatting context, hence won't be separated by space.

For example, the following markup:

```xml
<text>Hello</text> <text>World</text>
```

will render:
```
HelloWorld
```

to properly observe spaces, the `<text>` elements should be nested under another `<text>`, such as:
```xml
<text><text>Hello</text> <text>World</text></text>
```

will render:
```
Hello World
```

**TODO:**
- [X] add more demo cases to cover more use cases - done in #449
- [X] screenshots
- [x] Testing in Instawork app

Rework of #444

### DIFF
--- a/src/services/inline-context/helpers.js
+++ b/src/services/inline-context/helpers.js
@@ -19,15 +19,22 @@ export const convertLineBreaksIntoSpaces = (input: string): string =>
   input.replace(/\n/g, SPACE);
 
 export const ignoreSpacesFollowingSpace = (input: string[]): string[] => {
-  return input.map((value: string, index: number, values: string[]): string => {
-    if (index > 0) {
-      const previousEntry = values[index - 1];
-      const lastChar = previousEntry[previousEntry.length - 1];
-      if (lastChar === SPACE || previousEntry === EMPTY) {
-        return value.replace(/\s+/g, SPACE).trimStart();
-      }
+  let lastValueEndedWithSpace = false;
+  return input.map((value: string): string => {
+    // Replace every consecutive spaces with a single space
+    let trimmed = value.replace(/\s+/g, SPACE);
+
+    // If last string ended with a space, remove leading spaces
+    if (lastValueEndedWithSpace) {
+      trimmed = trimmed.trimStart();
     }
-    return value.replace(/\s+/g, SPACE);
+
+    // Set flag for next iteration, only when the current string is not empty
+    if (trimmed.length > 0) {
+      lastValueEndedWithSpace = trimmed[trimmed.length - 1] === SPACE;
+    }
+
+    return trimmed;
   });
 };
 
--- a/src/services/inline-context/helpers.js
+++ b/src/services/inline-context/helpers.js
@@ -12,12 +12,25 @@ export const LINE_BREAK = '\n';
 export const EMPTY = '';
 export const SPACE = ' ';
 
+/**
+ * Removes any space that follows a line break
+ */
 export const ignoreSpacesAfterLineBreak = (input: string): string =>
   input.replace(/\n\s+/g, LINE_BREAK);
 
+/**
+ * Replaces any line break by a space
+ */
 export const convertLineBreaksIntoSpaces = (input: string): string =>
   input.replace(/\n/g, SPACE);
 
+/**
+ * Provided a list of strings as input, removes any spaces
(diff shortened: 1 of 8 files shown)


src/services/inline-context/helpers.js
```diff
@@ -0,0 +1,72 @@
...
+  input.replace(/\n\s+/g, LINE_BREAK);
+
+export const convertLineBreaksIntoSpaces = (input: string): string =>
+  input.replace(/\n/g, SPACE);
+
+export const ignoreSpacesFollowingSpace = (input: string[]): string[] => {
+  return input.map((value: string, index: number, values: string[]): string => {
+    if (index > 0) {
+      const previousEntry = values[index - 1];
+      const lastChar = previousEntry[previousEntry.length - 1];
```
adamstep, line 25: if `previousEntry` is empty, won't this be an invalid index?

src/services/inline-context/helpers.js
```diff
@@ -0,0 +1,72 @@
...
+export const EMPTY = '';
+export const SPACE = ' ';
+
+export const ignoreSpacesAfterLineBreak = (input: string): string =>
+  input.replace(/\n\s+/g, LINE_BREAK);
+
+export const convertLineBreaksIntoSpaces = (input: string): string =>
+  input.replace(/\n/g, SPACE);
+
+export const ignoreSpacesFollowingSpace = (input: string[]): string[] => {
```
adamstep, line 21: This function would benefit from a comment showing an example input and output.

src/services/render/index.js
```diff
@@ -98,20 +115,28 @@ export const renderElement = (
...
-          return options.preformatted
-            ? trimmedValue
-            : trimmedValue.replace(/\s+/g, ' ');
+      if (
+        (element.parentNode?.namespaceURI === Namespaces.HYPERVIEW &&
+          element.parentNode?.localName === LOCAL_NAME.TEXT) ||
+        element.parentNode?.namespaceURI !== Namespaces.HYPERVIEW
+      ) {
+        if (options.preformatted) {
+          return element.nodeValue;
```
adamstep, line 124: So is it true that previously, we were (incorrectly) trimming a pre-formatted string?

flochtililoch: correct, the trimming was happening before the check for pre-formatted.

src/services/inline-context/helpers.js
flochtililoch: fixed in [f1aff6b](https://github.com/Instawork/hyperview/pull/448/commits/f1aff6b31e45855558f1cba2d9837ae462db896b)

src/services/inline-context/helpers.js
```diff
@@ -0,0 +1,72 @@
...
+      const lastChar = previousEntry[previousEntry.length - 1];
+      if (lastChar === SPACE || previousEntry === EMPTY) {
+        return value.replace(/\s+/g, SPACE).trimStart();
+      }
+    }
+    return value.replace(/\s+/g, SPACE);
+  });
+};
+
+export const trim = (input: string[]): string[] => {
```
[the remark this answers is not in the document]
flochtililoch: added in [0612ff9](https://github.com/Instawork/hyperview/pull/448/commits/0612ff9c7e70e32bbfa8335b0d45f23430992bf0)

src/services/inline-context/helpers.js
flochtililoch: added in [0612ff9](https://github.com/Instawork/hyperview/pull/448/commits/0612ff9c7e70e32bbfa8335b0d45f23430992bf0)

PR #449, opened 2022-07-11
chore: Break up text demo in multiple screens

flochtililoch:
Text demo screen is being crowded with examples, breaking it up to make it easier to add more examples in the future.

https://user-images.githubusercontent.com/309515/178171168-21d5dee0-85bf-40ed-8bf3-9d92743d606c.mp4


adamstep (approved):
Looks good, but this makes me think we should perhaps revisit https://github.com/Instawork/hyperview/pull/395 to make it easier to add new demo screens.

PR #453, opened 2022-07-17
chore(web): Event handlers stop propagation/prevent defaults

flochtililoch:
Wrap every press handler with helper that stop event propagation/prevent defaults. This does not affect UX on mobile, but allows for web interactions to perform as expected (so that the links added by behaviors are not "followed" by the browser).

### DIFF
--- a/src/core/hyper-ref/index.js
+++ b/src/core/hyper-ref/index.js
@@ -33,6 +33,16 @@ import VisibilityDetectingView from 'hyperview/src/VisibilityDetectingView';
 import { XMLSerializer } from 'xmldom-instawork';
 import { createTestProps } from 'hyperview/src/services';
 
+export const createEventHandler = (
+  handler: () => void,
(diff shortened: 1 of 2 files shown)


src/core/hyper-ref/index.js
```diff
@@ -33,6 +33,16 @@ import VisibilityDetectingView from 'hyperview/src/VisibilityDetectingView';
 import { XMLSerializer } from 'xmldom-instawork';
 import { createTestProps } from 'hyperview/src/services';
 
+export const createEventHandler = (
```
adamstep, line 36: Can you add a comment why we're using this wrapper?

flochtililoch: Added in [77fda83](https://github.com/Instawork/hyperview/pull/453/commits/77fda838f90af11274ea8fb25d47f5bad93738f3)

PR #457, opened 2022-07-17
fix: `hv-switch` imports

flochtililoch:
Wrap RN only imports to prevent build failures

### DIFF
--- a/src/components/hv-switch/index.js
+++ b/src/components/hv-switch/index.js
@@ -17,9 +17,9 @@ import {
   createStyleProp,
   getNameValueFormInputValues,
 } from 'hyperview/src/services';
-import type { ColorValue } from 'react-native/Libraries/StyleSheet/StyleSheetTypes';
+import type { ColorValue } from './style-sheet';
 import { LOCAL_NAME } from 'hyperview/src/types';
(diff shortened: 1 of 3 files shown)


adamstep:
Since `normalizeColor` doesn't do anything on the web, will this affect the UI in any way that matters?

flochtililoch:
> Since `normalizeColor` doesn't do anything on the web, will this affect the UI in any way that matters?

It might, I honestly am not sure what we use it for. This workaround is mostly to get around build failures as the import does not exists in `react-native-web`. This might be something we revisit in the future.

adamstep:
Got it, can we create issues for things like this where we need to add web support later?

adamstep (approved):

lenwhite:
The what `normalizeColor` was used to solve in that PR is color-consistency in the switch. 

As seen in the screenshot above, the **Switch on** state had a green color (the system-default `thumbColor`) with a blue background (that was styled). https://github.com/Instawork/hyperview/pull/368 allowed the setting of the `thumbColor`, with it defaulting to a darker shade of the `backgroundColor` if not set to maintain consistency in appearance, as per the screenshots in the PR:

image


image

`normalizeColor` was used to auto-generate the background color on android devices. This functionality was limited to android only (`Platform.OS === 'android'`) though we could have it for web as well.

PR #460, opened 2022-07-17
fix(examples): List layouts

flochtililoch:
- Remove scrollview container from base layout, and create a separate scrollview layout. Create `container` block, and update related screens that need to extend it. (Solves RN warning about nested scrollable views)
- Update `directory_list.xml.njk` to not extend a base template and instead just render the list. Screens entry point now extend the default base template. In parallel, each screen that rely on a directory list now creates a separate `list.xml` template that extends `directory_list.xml.njk`. (Solves list pull to refresh mechanism)

https://user-images.githubusercontent.com/309515/179424708-80457e0c-18ab-4fa2-9fa8-813ba9188724.mp4

### DIFF
--- a/examples/_includes/directory_list.xml.njk
+++ b/examples/_includes/directory_list.xml.njk
@@ -1,13 +1,13 @@
 {# Standard template for showing a list of links to other screens. List is generated from a collection of screens with the given "list_tag" #}
-{% extends 'base.xml.njk' %}
 
-{% block styles %}
-
-{% endblock %}
-
-{% block content %}
-  <behavior trigger="refresh" action="reload" href="#" />
-  <list>
+<list
+  xmlns="https://hyperview.org/hyperview"
+  trigger="refresh"
+  action="replace"
+  target="directory-list"
+  id="directory-list"
+  href="{% block list_href %}{% endblock %}"
+>
   {% for screen in collections[list_tag] %}
     <item
       href="{{ screen.data.permalink }}"
@@ -21,5 +21,4 @@
       <text style="Item__Chevron">&gt;</text>
     </item>
   {% endfor %}
-  </list>
-{% endblock %}
+</list>
--- a/examples/_includes/directory_list.xml.njk
+++ b/examples/_includes/directory_list.xml.njk
@@ -6,7 +6,7 @@
   action="replace"
   target="directory-list"
   id="directory-list"
-  href="{% block list_href %}{% endblock %}"
+  href="{% if list_href %}{{ list_href }}{% else %}{{ permalink }}{% endif %}"
 >
   {% for screen in collections[list_tag] %}
     <item
--- a/examples/_includes/base.xml.njk
+++ b/examples/_includes/base.xml.njk
@@ -12,9 +12,7 @@
         {% endif %}
         <text style="Header__Title">{{ title }}</text>
       </header>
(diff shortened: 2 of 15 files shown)


examples/_includes/directory_list.xml.njk
```diff
@@ -1,13 +1,13 @@
...
-
-{% endblock %}
-
-{% block content %}
-  <behavior trigger="refresh" action="reload" href="#" />
-  <list>
+<list
+  xmlns="https://hyperview.org/hyperview"
+  trigger="refresh"
+  action="replace"
```
adamstep, line 6: I think the original screens were written before we had the `reload` action. With the addition of this action, I think we don't need to override `list_href` for each directory, or have a separate list template, right? That would simplify this work quote a bit.

flochtililoch: The `reload` action on lists (paired with `href="#"`) does not work expected - the loading state remains stuck:

https://user-images.githubusercontent.com/309515/179427537-bfbf6adc-1e42-4926-9702-67e4ed254689.mp4

adamstep: I see. You can also try to use the eleventy metadata, something like:
`href="{{ permalink }}"` instead of `href="{% block list_href %}{% endblock %}"`.

flochtililoch: I think the issue will be that this will generate a URL to the whole screen, not the URL of the list partial

flochtililoch: I see I can define another variable but it looks like it would needs to be defined in both `index.xml.njk` and `list.xml.njk`. This is how the code change looks like:
```diff
diff --git a/examples/_includes/directory_list.xml.njk b/examples/_includes/directory_list.xml.njk
index eb4a793..6e42e2b 100644
--- a/examples/_includes/directory_list.xml.njk
+++ b/examples/_includes/directory_list.xml.njk
@@ -6,7 +6,7 @@
   action="replace"
   target="directory-list"
   id="directory-list"
-  href="{% block list_href %}{% endblock %}"
+  href="{{ list_href }}"
 >
   {% for screen in collections[list_tag] %}
     <item
diff --git a/examples/index.xml.njk b/examples/index.xml.njk
index e4a1f38..8e5cec0 100644
--- a/examples/index.xml.njk
+++ b/examples/index.xml.njk
@@ -2,7 +2,8 @@
 title: "Hyperview"
 permalink: "index.xml"
 list_tag: "root"
-hide_back_button: true,
+hide_back_button: true
+list_href: "list.xml"
 ---
 {% extends 'base.xml.njk' %}

diff --git a/examples/list.xml.njk b/examples/list.xml.njk
index c36f88e..7b98f5b 100644
--- a/examples/list.xml.njk
+++ b/examples/list.xml.njk
@@ -1,7 +1,6 @@
 ---
 permalink: "list.xml"
 list_tag: "root"
+list_href: "list.xml"
 ---
 {% extends "directory_list.xml.njk" %}
-
-{% block list_href %}list.xml{% endblock %}
```

flochtililoch: Oh yeah that could work, I was under the impression if statements did not exist in njk templates. Will try that.

adamstep (approved):

Issue #307, opened 2021-05-25
Support for inline links


adamstep:
See screenshot attached - implementation of such UI isn't quite possible at the moment.

A typical React Native implementation for this UI would be:

```
<Text>
  Includes deduction to cover Occupational Accident Insurance.
  <TouchableOpacity onPress={onPressMoreInfo}>
    <Text style={link}>More Info here</Text>
  </TouchableOpacity>
</Text>
```

which we could try to translate in HXML to

```
<text>
  Includes deduction to cover Occupational Accident Insurance.
  <text style="link" href="/more-info" action="push" trigger="press">More Info here</text>
</text>
```

However, this triggers a React Native error, as the added "href" internally wraps the inner link text component inside a view, and React Native requires <View> nested within <Text> to have a width and height defined.

Since the text needs to flow based on its content, we cannot infer the width and height, so we should refactor the `addHref` helper in hyperview to support this use case.

This somewhat relates to #288

terryatgithub:
we've encountered this issue now as below, looking forward to an early resolution.
![image](https://user-images.githubusercontent.com/12246394/130736694-33f8b0df-8d7f-4582-ae8f-a4b06f41ab6d.png)

flochtililoch:
Fixed by #462

PR #462, opened 2022-07-18
fix: Inline links

flochtililoch:
When rendering `hyper-ref` component within a text element, use `Text` component instead of `TouchableOpacity`.

Note: I tried using more modern `Pressable` component as a replacement of `TouchableOpacity`, but it did not allow keeping inline layout on nested `<text>`.

Solves #307

https://user-images.githubusercontent.com/309515/180471486-f5ba11fa-d502-4eed-8cb9-b1d59e0a5d9b.mp4

https://user-images.githubusercontent.com/309515/183530120-271e4c5d-dac2-4953-a000-f73dd5dd2f3d.mov

### DIFF
--- a/src/core/hyper-ref/index.js
+++ b/src/core/hyper-ref/index.js
@@ -10,9 +10,11 @@
 
 import * as Dom from 'hyperview/src/services/dom';
 import * as Events from 'hyperview/src/services/events';
+import * as Namespaces from 'hyperview/src/services/namespaces';
 import * as Render from 'hyperview/src/services/render';
 import {
   ACTIONS,
+  LOCAL_NAME,
   NAV_ACTIONS,
   PRESS_TRIGGERS,
   TRIGGERS,
@@ -30,7 +32,12 @@ import type {
 } from 'hyperview/src/types';
 import type { PressHandlers, Props, State } from './types';
 import React, { PureComponent } from 'react';
-import { RefreshControl, ScrollView, TouchableOpacity } from 'react-native';
+import {
+  RefreshControl,
+  ScrollView,
+  Text,
+  TouchableOpacity,
+} from 'react-native';
 import type { Node } from 'react';
 import VisibilityDetectingView from 'hyperview/src/VisibilityDetectingView';
 import { XMLSerializer } from 'xmldom-instawork';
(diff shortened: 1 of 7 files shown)


src/core/hyper-ref/index.js
```diff
@@ -305,10 +304,51 @@ export default class HyperRef extends PureComponent<Props, State> {
...
+      this.props.element.parentNode?.localName === LOCAL_NAME.TEXT;
+
+    if (isNestedUnderText) {
+      const noop = () => {};
+      return (
+        <Text
+          accessibilityLabel={accessibilityLabel}
+          accessible={false}
+          onLongPress={onLongPress}
+          // when no press handler set, we still need an empty handler for pressIn or pressOut handlers to work
```
adamstep, line 323: How did you figure out this issue and the one below for `onResponderRelease`/`onResponderRelease`? Is there an SO of GH issue we can link to?

src/core/hyper-ref/index.js
```diff
@@ -305,10 +304,51 @@ export default class HyperRef extends PureComponent<Props, State> {
...
+          onPress={onPress || (onPressIn || onPressOut ? noop : undefined)}
```
flochtililoch, line 324:
```suggestion
          onPress={onPress || ((onPressIn || onPressOut) ? noop : undefined)}
```

src/core/hyper-ref/index.js
flochtililoch: I just ran into the bug myself, I can check if there's any document out there for this

src/core/hyper-ref/index.js
flochtililoch: hmm I guess prettier does not want these extra parenthesis and `?` is interpreted on the whole expression, not just the last bit, so change is not needed.

src/core/hyper-ref/index.js
flochtililoch: I haven't found anything of value in GH/SO so far. I might pause this one a little bit and do a bit more testing on it including Android

adamstep (approved):

abdelav (approved):

src/core/hyper-ref/index.js
flochtililoch: Tested on Android 12, it's working well.

PR #464, opened 2022-07-22
refactor: `HyperRef` component

flochtililoch:
Break up the `HyperRef.render` method into smaller, more manageable chunks.
Note to reviewers: it'll be easier to take a look first at individual commits to understand the thought process behind this refactor.

Verified that demo app continues to work as expected

https://user-images.githubusercontent.com/309515/180355495-92c85252-9cd4-4b87-9f0b-7a7674a28695.mp4

https://user-images.githubusercontent.com/309515/180359323-7310790a-a9dd-4411-b44e-d5c9238586eb.mp4

### DIFF
--- a/src/core/hyper-ref/index.js
+++ b/src/core/hyper-ref/index.js
@@ -26,6 +26,7 @@ import type {
   PressTrigger,
   StyleSheet,
   StyleSheets,
+  Trigger,
 } from 'hyperview/src/types';
 import type { PressHandlers, Props, State } from './types';
 import React, { PureComponent } from 'react';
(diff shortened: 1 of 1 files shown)


src/core/hyper-ref/index.js
```diff
@@ -62,6 +62,8 @@ export default class HyperRef extends PureComponent<Props, State> {
     refreshing: false,
   };
 
+  behaviorElements: Element[];
```
adamstep, line 67: This makes me a little nervous, I don't remember when all the React component lifecycle methods fire. Will `componentDidUpdate` always get called, including when we first render the component?

flochtililoch: `componentDidUpdate` only gets called when the state/props change (see [React doc](https://reactjs.org/docs/react-component.html#componentdidupdate)). For the initial rendering, we rely on setting these in the component's constructor

adamstep: Ah I see that was added in a later commit, thanks.

adamstep (approved):
Nice work, definitely cleaner!

PR #465, opened 2022-07-22
fix: Custom loading screen

flochtililoch:
Restore functionality of custom loading screens, which got broken in #69.

Implementation notes:
The dictionary that keeps track of loading screen was improperly moved to the `navigation` instance, which is re-created on each screen. Move this back to a global variable so new pushed screens can access it.

https://user-images.githubusercontent.com/309515/180581726-9c5ebd87-7f19-4321-98f3-ac68c363c3fb.mp4


adamstep (approved):
Good find! I guess the default loading screen was too similar to the custom one, so we missed this!

PR #467, opened 2022-07-26
feat(11ty): Sort template collections

flochtililoch:
Add filter that allows sorting collections of templates in alphabetical order.


adamstep (approved):
nice find

PR #470, opened 2022-08-22
chore(schema): Allow `key` attribute

flochtililoch:
`key` attribute is used by React when rendering component in the context of loop. We need to allow it on every elements for XML validation to pass.

### DIFF
--- a/schema/core.xsd
+++ b/schema/core.xsd
@@ -287,6 +287,10 @@
     <xs:restriction base="xs:token" />
   </xs:simpleType>
 
+  <xs:simpleType name="KEY">
+    <xs:restriction base="xs:token" />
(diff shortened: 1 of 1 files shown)


adamstep:
I don't understand how React rendering affects the Hyperview XML. Do you have an example where this is causing issues?

flochtililoch:
Key is extracted [here](https://github.com/Instawork/hyperview/blob/master/src/services/render/index.js#L91-L93) during rendering, we need it [here](https://github.com/Instawork/instawork/pull/19672) to remove these pesky warnings thrown by React.

adamstep:
Got it, thanks for the clarification.

adamstep (approved):

PR #471, opened 2022-08-27
fix(installation): change instructions for physical

Atem18:
Files paths have changed for physical devices


flochtililoch (approved):
Thanks for the update!

PR #475, opened 2022-09-07
refactor: `hv-view` component

flochtililoch:
Reorganize code to ease future improvements, fix a bug, improve typing and documentation.
Note to reviewers: check commits in the history order, which should considerably help understanding the path taken for this refactor, and should ease the review process.

- d884a870a63656665d7bb5807fc5cebe8bdc5d9c fix(docs): add missing view attributes
- b00d8637671f80d5cdbef619b9770bafaf36ab33 fix: `if` condition not reading the correct param (introduced in #414)
- 76a367a660683645c6819e04987f423103573581 refactor: explicitly defined / typed attributes
- e55c3d88bee1fd10e8b218753214ad848ba9ea6a refactor: keeping track of text input refs
- 2b83003cddbe5916e7625e0afec1083a05adfcc4 refactor: remove unneeded denormalized variables
- 6e7d91bcd5adf723b5d4eb8b4e268412864dc59b refactor: move local variables closer to where they are used
- 061d8bc66dc2b5482baf74086893c3315e011347 refactor: remove unneeded mutation
- 4378170fa7c1379eff52bf74cb3f73da9be14bdb chore: remove remains of `react-native-keyboard-aware-scroll-view`
- 2b7da176bbc3d2c3014c047111c9d3a49c6867ce refactor: render `children` earlier
- ff81c67bc595f3a961821c76b1742d36a5aa966a refactor: handle safe area view earlier
- 529cb96e76f9dc903b904495d0cd91dd3afe8183 refactor: combine if conditions
- cc8f7305c14cdb2b6c2036193fd54f48ea6c0285 refactor: remove implicit props
- 042c3889b5423fff1695a31a59a322ddc142111b refactor: isolate scroll view props
- c44d02732fad3d6acfe040598516dd17008331b9 refactor: isolate keyword aware scroll view props
- 97b80d7e0019aaed1512a7b3b32d5e8160395682 refactor: isolate rendering component from render method
- a9b34852ac2a16cda6db7959a54b17b2ab9fca17 refactor: use JSX to render content
- e3c1271a8a0c25db950be67ebab911f3cf641f3c refactor: isolate common props and remove props mutation
- 0100831cf3a171c2b80ae831078f7b3d48df11cc fix: types

https://user-images.githubusercontent.com/309515/188980937-c8472867-074c-4894-967b-d00d94fe317c.mp4


src/components/hv-view/index.js
```diff
@@ -41,142 +47,213 @@ export default class HvView extends PureComponent<HvComponentProps> {
...
+  componentDidUpdate(prevProps: HvComponentProps) {
+    if (prevProps.element === this.props.element) {
+      return;
+    }
+
+    this.updateAttributes();
+  }
+
+  updateAttributes = () => {
+    // $FlowFixMe: reduce returns a mixed type, not Attributes
```
yozuliangInsta, line 66: It seems to only retrieve fields defined in ATTRIBUTES, so why is the return type not Attributes? s it throwing a type error?

src/components/hv-view/index.js
```diff
@@ -41,142 +47,213 @@ export default class HvView extends PureComponent<HvComponentProps> {
...
     );
-    const scrollable = !!this.props.element.getAttribute('scroll');
+    const id = this.props.element.getAttribute('id');
+    if (!id) {
+      return { style };
+    }
+    if (Platform.OS === 'ios') {
+      return { style, testID: id };
+    }
+    return { accessibilityLabel: id, style };
```
yozuliangInsta, line 101: Why does the id need different name for iOS vs. Android?

src/components/hv-view/index.js
```diff
@@ -41,142 +47,213 @@ export default class HvView extends PureComponent<HvComponentProps> {
...
-        props.horizontal = true;
-      }
+  getScrollToInputAdditionalOffsetProp = (): number => {
+    const defaultOffset = 120;
+    if (this.attributes[ATTRIBUTES.SCROLL_TO_INPUT_OFFSET]) {
+      const offset = parseInt(
+        this.attributes[ATTRIBUTES.SCROLL_TO_INPUT_OFFSET],
+        10,
+      );
+      return Number.isNaN(offset) ? 0 : defaultOffset;
```
yozuliangInsta, line 153: do we not care about the actual value of offset?

```suggestion
      );
      return Number.isNaN(offset) ? defaultOffset : offset;

File: src/components/hv-view/index.js [hunk 4]
```diff
@@ -41,142 +47,213 @@ export default class HvView extends PureComponent<HvComponentProps> {
...
+            {children}
+          </KeyboardAwareScrollView>
+        );
       }
+      return (
+        <ScrollView
+          {...{
+            ...this.getCommonProps(),
+            ...this.getScrollViewProps(children),
+          }}
```
yozuliangInsta, line 225: this spread usage looks a little funny, would the following return the same structure?
```suggestion
          {
            ...this.getCommonProps(),
            ...this.getScrollViewProps(children),
          }
```

src/components/hv-view/index.js
flochtililoch: I don't remember the exact details about this, but here's some context
- https://github.com/Instawork/hyperview/pull/227
- https://github.com/Instawork/hyperview/pull/93

src/components/hv-view/index.js
flochtililoch: the surrounding curly brackets here are not representing a dictionary, but rather a [placeholder in a JSX template](https://reactjs.org/docs/introducing-jsx.html#embedding-expressions-in-jsx). Here, we spread a dictionary that's created on the fly with two other dictionaries (returned by the two function calls)

src/components/hv-view/index.js
flochtililoch: Yes, it's throwing a typing error. The problem appear to be with the `reduce` operation, which contains an inner function and might be too complicated to infer from Flow. There might be ways to solve that kind of typing error but I haven't found one yet.

src/components/hv-view/index.js
flochtililoch: hmm, you're right, this code seems odd. I rewrote it without changing the actual logic (see [here](https://github.com/Instawork/hyperview/pull/475/files#diff-d20a18a04fc2288d54bd648885b73443cbd01e36ba101bdc8331e4bdf9283437L93-L96)), but this might have been an existing bug.

yozuliangInsta (approved):

PR #484, opened 2022-11-07
feat(header): support stale reason header and stale error trigger

palaparthi:
### Summary
1. Add support for `X-Response-Stale-Reason` header, this header is set when content fetched is stale, typically served from a cache.
2. Implement new trigger `load-stale-error`, so that the UI can be modified when the content served is stale.

### DIFF
--- a/src/services/dom/index.test.js
+++ b/src/services/dom/index.test.js
@@ -11,6 +11,7 @@
 import * as Dom from 'hyperview/src/services/dom';
 import * as UrlService from 'hyperview/src/services/url';
 import { version } from 'hyperview/package.json';
+import { X_RESPONSE_STALE_REASON } from './types';
 
 // Mock instawork-xmldom module
 const mockExpectedDocument = { foo: 'bar' };
@@ -25,12 +26,8 @@ jest.mock('xmldom-instawork', () => {
 // Mock other dependencies
 const fetchMock = jest.fn();
 const responseTextMock = jest.fn();
-const headers = new Map();
-headers.set(Dom.HTTP_HEADERS.X_RESPONSE_STALE_REASON, 'stale-if-error');
 // $FlowFixMe
 fetchMock.mockResolvedValue({ text: responseTextMock });
-// $FlowFixMe
-fetchMock.mockResolvedValueOnce({ headers, text: responseTextMock });
 const beforeParseMock = jest.fn();
 const afterParseMock = jest.fn();
 const urlServiceAddFormDataToUrlMock = jest.spyOn(
@@ -51,7 +48,7 @@ describe('Parser', () => {
 
   describe('load', () => {
     describe('offline GET', () => {
-      it('sets the right isStale value', async () => {
+      it('sets the right staleHeaderType', async () => {
         const url = 'http://foo/bar';
         const expectedOptions = {
           body: undefined,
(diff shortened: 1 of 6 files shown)


src/services/dom/index.test.js
```diff
@@ -25,8 +25,12 @@ jest.mock('xmldom-instawork', () => {
 // Mock other dependencies
 const fetchMock = jest.fn();
 const responseTextMock = jest.fn();
+const headers = new Map();
+headers.set(Dom.HTTP_HEADERS.X_RESPONSE_STALE_REASON, 'stale-if-error');
 // $FlowFixMe
 fetchMock.mockResolvedValue({ text: responseTextMock });
+// $FlowFixMe
+fetchMock.mockResolvedValueOnce({ headers, text: responseTextMock });
```
flochtililoch, line 33: If I understand correctly, we mock once because the first test is the one expecting to use this mock? I think this is a bit brittle. Can this mock be created in the context of the test that needs it?

flochtililoch (commented):
The "stale" flag seem to be set as a global state for the current screen, but a screen can trigger multiple fragments loads, also I wonder if the storage strategy is correct. Would we be better off maintaining a map of URLs and their stale state instead?

palaparthi:
> The "stale" flag seem to be set as a global state for the current screen, but a screen can trigger multiple fragments loads, also I wonder if the storage strategy is correct. Would we be better off maintaining a map of URLs and their stale state instead?

@adamstep and I briefly discussed regarding this. The assumption behind this is `X-Response-Stale-Reason` is only set when the request itself is unable to finish due to network failure or if server is unreachable. So it's ok to consider the entire screen is stale in these cases. I agree that tracking the stale state individually per URL is the right approach but might not be worth it for our initial usecase.

src/services/dom/parser.js
```diff
@@ -58,7 +58,7 @@ export class Parser {
     data: ?FormData,
     httpMethod: ?HttpMethod,
     acceptContentType: string = CONTENT_TYPE.APPLICATION_VND_HYPERVIEW_XML,
-  ): Promise<Document> => {
+  ): Promise<{ doc: Document, isStale: boolean }> => {
```
adamstep, line 61: just thinking ahead a bit, how will we distinguish between stale due to error vs stale while revalidating? Maybe instead of a boolean, should we return an enum?

src/services/dom/parser.js
```diff
@@ -87,8 +87,12 @@ export class Parser {
     const contentType: string = response.headers?.get(
       HTTP_HEADERS.CONTENT_TYPE,
     );
+    const staleHeader: string = response.headers?.get(
```
adamstep, line 90: To future-proof this code once we introduce `stale-white-revalidate`, we shouldn't just check for the presence of the header, but also for the value. https://github.com/Instawork/mobile/pull/4697/files#diff-0ff31a656ba253b8f63cd07921f9daf4606bce49c587ccbf5f2b877765af55baR82

palaparthi:
> I spoke more with @flochtililoch about this. He brought up an interesting case. What if the screen is loaded stale (due to user being offline). Then they go online, and we load a fragment on the screen. In that case, the screen's state will be "fresh", even though the bulk of the content is stale.
> 
> 
> 
> Perhaps it makes sense that once a screen is marked "stale", we keep it stale until a `reload` happens?

> Yeah that's a good point. We can update the `isStale` state in `fetchElement`(called only when loading partial) only when the content is "stale" and no-op if it's fresh. I think this way we can keep the screen stale until a reload happens.

src/services/dom/parser.js
palaparthi: addressed in https://github.com/Instawork/hyperview/pull/484/commits/664aefe39ad93cc1f0085b0e14ab5175a65e39bb

src/services/dom/parser.js
palaparthi: addressed in https://github.com/Instawork/hyperview/pull/484/commits/664aefe39ad93cc1f0085b0e14ab5175a65e39bb

src/services/dom/index.test.js
palaparthi: addressed in https://github.com/Instawork/hyperview/pull/484/commits/664aefe39ad93cc1f0085b0e14ab5175a65e39bb

adamstep (approved):

PR #504, opened 2023-02-09
feat: Handle element error gracefully

palaparthi:
### Summary
When the document is already loaded on to the screen but the subsequent requests fail due to an error while scrolling, currently entire screen shows an error message, and we lose the previous content.
This PR aims to handle HTTP errors gracefully by showing a banner at the bottom indicating there was an error but still let the user browse the content which was already loaded.

### Videos
| Before | After |
| --- | --- |
| <video src="https://user-images.githubusercontent.com/28518512/217921739-5aed32a3-fd03-4d93-b1fc-8cfe406da5f5.mp4" /> | <video src="https://user-images.githubusercontent.com/28518512/217847216-56d2e123-e7f9-4933-89c0-d3b28cf345b9.mp4" /> |

### DIFF
--- a/src/core/components/load-element-error/index.js
+++ b/src/core/components/load-element-error/index.js
@@ -13,7 +13,6 @@ import { SafeAreaView, Text, View } from 'react-native';
 import type { Props } from './types';
 import styles from './styles';
 
-
 const LoadElementError = (props: Props) => {
   const getError = () => {
     if (!props.error) {
@@ -22,7 +21,10 @@ const LoadElementError = (props: Props) => {
     if (__DEV__) {
       return `${props.error.name}: ${props.error.message}`;
     }
-    if (props.error.name == 'TypeError' && props.error.message === 'Network request failed') {
+    if (
+      props.error.name == 'TypeError' &&
+      props.error.message === 'Network request failed'
+    ) {
       return 'You seem to be offline, check your connection';
     }
     return 'An error occured';
--- a/src/core/components/load-element-error/index.js
+++ b/src/core/components/load-element-error/index.js
@@ -1,0 +1,40 @@
+// @flow
+
+/**
+ * Copyright (c) Garuda Labs, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import React from 'react';
+import { SafeAreaView, Text, View } from 'react-native';
+import type { Props } from './types';
+import styles from './styles';
+
+
+const LoadElementError = (props: Props) => {
+  const getError = () => {
+    if (!props.error) {
+      return 'An error occured';
+    }
+    if (__DEV__) {
+      return `${props.error.name}: ${props.error.message}`;
+    }
+    if (props.error.name == 'TypeError' && props.error.message === 'Network request failed') {
+      return 'You seem to be offline, check your connection';
+    }
+    return 'An error occured';
+  };
+
+  return (
+    <SafeAreaView>
+      <View style={styles.container}>
+        <Text style={styles.title}>{getError()}</Text>
+      </View>
+    </SafeAreaView>
+  );
+};
+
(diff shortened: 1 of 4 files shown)


adamstep:
Thanks @palaparthi . Just to make the impact of this change clearer, can you include a video of the experience before this change included (entire screen taken over with load error)

palaparthi:
> Just to make the impact of this change clearer, can you include a video of the experience before this change included (entire screen taken over with load error)

@adamstep Added the old experience.

src/index.js
```diff
@@ -305,13 +314,12 @@ export default class HyperScreen extends React.Component {
         // We are doing this to ensure that we keep the screen stale until a `reload` happens
         this.setState({ staleHeaderType });
       }
+      if (this.state.elementError) {
+        this.setState({ elementError: null });
+      }
```
flochtililoch, line 319: shouldn't the error stated cleared before the request is emitted?

src/index.js
```diff
@@ -431,7 +439,9 @@ export default class HyperScreen extends React.Component {
           targetElement = currentElement;
         }
 
-        newRoot = Behaviors.performUpdate(action, targetElement, newElement);
+        if (newElement) {
+          newRoot = Behaviors.performUpdate(action, targetElement, newElement);
+        }
```
flochtililoch, line 444: this is an unexpected change with regards to the feature this PR is introducing, is this solving a bug?

src/core/components/load-element-error/index.js
```diff
@@ -0,0 +1,39 @@
...
+    ) {
+      return 'You seem to be offline, check your connection';
+    }
+    return 'An error occured';
+  };
+
+  return (
+    <SafeAreaView>
+      <View style={styles.container}>
+        <Text style={styles.title}>{getError()}</Text>
```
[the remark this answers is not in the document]
adamstep: It's a little trickier for fragments. For full screen, we can reload it. But for fragments, we don't know which fragment to retry if multiple requests fail. The structure of the page can also change between the time the request fails and the retry happens.

src/core/components/load-element-error/index.js
```diff
@@ -0,0 +1,39 @@
...
+import styles from './styles';
+
+const LoadElementError = (props: Props) => {
+  const getError = () => {
+    if (__DEV__) {
+      return `${props.error.name}: ${props.error.message}`;
+    }
+    if (
+      props.error.name === 'TypeError' &&
+      props.error.message === 'Network request failed'
```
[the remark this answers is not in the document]
adamstep: Maybe we can use the official `fetch` API definition of a network error, not sure if this implementation follows the spec exactly: https://fetch.spec.whatwg.org/#concept-network-error

src/index.js
palaparthi: If we clear the error state before the request was made the screen will re-render and the banner would flicker in the case where the error still persists. This might be a good user experience imo, what do you think?

https://user-images.githubusercontent.com/28518512/217935762-2aac5024-557e-4203-a7bc-5f3516c7a292.mp4

src/index.js
palaparthi: Yes, we are returning [null](https://github.com/Instawork/hyperview/blob/sree/handle-fragment-network-error/src/index.js#L324) from `fetchElement` when there is an error, so this tries to perform update on null and breaks.

src/index.js
flochtililoch: Got it, I think you're right. I was looking at this in an optimistically update point of view, assuming that the next call would succeed but that may not be true.

flochtililoch (approved):

src/core/components/load-element-error/index.js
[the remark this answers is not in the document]
palaparthi: @adamstep when there is a network error the only way to detect is via message as far as I can tell, I think RN implementation takes care following the spec and hides the details .

PR #510, opened 2023-03-03
docs: Add `avoid-keyboard` note

tecnoloco:
Add a quick note about `avoid-keyboard` . I found that the direct parent element should have the same screen height to work properly


docs/reference_view.md
```diff
@@ -146,6 +146,8 @@ If `hide="true"`, the element will not be rendered on screen. If the element or
 
 An attribute to solve the common problem of views that need to move out of the way of the virtual keyboard. It can automatically adjust the position of its children based on the keyboard height. This is useful when you want keyboard avoiding behavior in non-scrollable views. It is applied only in iOS since Android has built-in support for avoiding keyboard.
 
+> For `avoid-keyboard` to work properly the direct parent element should have same screen height.
```
adamstep, line 149: This could be clarified. "the same screen height" as what? As the element with the `avoid-keyboard` attribute?

tecnoloco: [8aea958](https://github.com/Instawork/hyperview/pull/510/commits/8aea958755f2608f4f521ed16132c535085ae666)

adamstep (approved):

PR #514, opened 2023-03-07
feat(section-list): remove 'section' requirement

### LINKED ISSUE #294: Rethink `<section-list>` syntax to better support pagination
### Overview
Currently, `<section-list>` elements contain `<section>` elements, which contain `<section-title>` and `<item>` elements. See [reference](https://hyperview.org/docs/reference_sectionlist) for details. This structure doesn't work well to implement infinite-scroll behavior, which may need to both add items to the last section, and add new sections.

To solve this, we should remove the requirement for `<section>` elements altogether. `<section-list>` would just contain `<section-title>` and `<item>` elements as direct children. Grouping would be implied based on the ordering. With this approach, new responses could add items to the last section and add new sections.

### Current Implementation
Take for example, a section list grouping items by date. We want to implement "infinite-scroll": when the user scrolls to the bottom of the section list, we make a request to the server for the next page of data. The response contains XML markup that gets appended to a part of the section list. See the example below, with two options for where we can append the data.

```xml
<section-list id="dates">
  <section>
    <section-title>
      <text>Feb 2023</text>
    </section-title>

    <item key="item-1">
      <text>Feb 1</text>
    </item>
  </section>

  <section>
    <section-title>
      <text>Jan 2023</text>
    </section-title>

    <item key="item-2">
      <behavior trigger="visible" action="append" target="dates" href="/dates?page=2" />
      <text>Jan 1</text>
    </item>
  </section>

  <!-- response from /dates?page=2 will be appended here -->
</section-list>
```

When "Jan 1" scrolls into view, the `<behavior>` element will be triggered, and Hyperview will
- request XML from the server (`GET /dates?page=2`)
- append it to the `<section-list>` element (`id="dates"`)
- re-render the UI to show the new content

What if the response from the server contains more dates that should be grouped in Jan 2023? The response may look something like this:

```xml
<items>
  <section>
    <section-title>
      <text>Jan 2023</text>
    </section-title>

    <item key="item-3">
      <text>Jan 3</text>
    </item>
  </section>
  <section>
    <section-title>
      <text>Feb 2023</text>
    </section-title>

    <item key="item-4">
      <text>Feb 2</text>
    </item>
  </section>
</items>
```

After appending the response XML, the rendered UI will show two sections with the same title (Jan 2023), which is not ideal and could confuse users:

```xml
<section-list id="dates">
  <section>
    <section-title>
      <text>Feb 2023</text>
    </section-title>

    <item key="item-1">
      <text>Feb 1</text>
    </item>
  </section>

  <!-- First section labeled Jan 2023 -->
  <section>
    <section-title>
      <text>Jan 2023</text>
    </section-title>

    <item key="item-2">
      <behavior trigger="visible" action="append" target="dates" href="/dates?page=2" />
      <text>Jan 1</text>
    </item>
  </section>

  <!-- Second section labeled Jan 2023 -->
  <section>
    <section-title>
      <text>Jan 2023</text>
    </section-title>

    <item key="item-3">
      <text>Jan 3</text>
    </item>
  </section>

  <section>
    <section-title>
      <text>Feb 2023</text>
    </section-title>

    <item key="item-4">
      <text>Feb 2</text>
    </item>
  </section></section-list>
```

### Proposed solution
We can remove the requirement for `<section>` elements in `<section-list>`. The only purpose it serves is to group the items and section title together. However, this grouping can be implied by the ordering of `<section-title>` and `<item>` elements:

```xml
<section-list id="dates">
    <section-title>
      <text>Feb 1, 2023</text>
    </section-title>

    <item key="item-1">
      <text>Item 1</text>
    </item>

    <section-title>
      <text>Feb 2, 2023</text>
    </section-title>

    <item key="item-2">
      <behavior trigger="visible" action="append" target="dates" href="/dates?page=2" />
      <text>Item 2</text>
    </item>

    <!-- response from /dates?page=2 will be appended here -->
</section-list>
```

The server response for `/dates?page=2` will also only contain `<item>` and `<section-title>` elements.

```xml
<items>
    <item key="item-3">
      <text>Jan 3</text>
    </item>

    <section-title>
      <text>Feb 2023</text>
    </section-title>

    <item key="item-4">
      <text>Feb 2</text>
    </item>
</items>
```

When appended, the "Jan 3" will be grouped with the existing "Jan 2023" section, and "Feb 2" will be grouped in the "Feb 2023" section:

```xml
<section-list id="dates">
    <section-title>
      <text>Feb 1, 2023</text>
    </section-title>

    <item key="item-1">
      <text>Item 1</text>
    </item>

    <section-title>
      <text>Feb 2, 2023</text>
    </section-title>

    <item key="item-2">
      <behavior trigger="visible" action="append" target="dates" href="/dates?page=2" />
      <text>Item 2</text>
    </item>

    <item key="item-3">
      <text>Jan 3</text>
    </item>

    <section-title>
      <text>Feb 2023</text>
    </section-title>

    <item key="item-4">
      <text>Feb 2</text>
    </item>
</section-list>
```

For backwards compatibility, The XML should still work if `<section>` elements are present, we just simply ignore them and look at the ordering of `<section-title>` and `<item>` elements within the `<section>`.

### Implementation
- The code to modify is in the [HvSectionList](https://github.com/Instawork/hyperview/blob/master/src/components/hv-section-list/index.js) component, specifically the [render()](https://github.com/Instawork/hyperview/blob/master/src/components/hv-section-list/index.js#L91) method. This code walks through the XML DOM of the `<section-list>` element, extracting the items and section titles and building up a `sections` array, which gets passed as a prop to the RN [SectionList](https://reactnative.dev/docs/0.67/sectionlist) component. This code will need to be modified to create the `sections` array, without relying on the presence of `<section>` elements.
- The stories for HvSectionList should be enhanced, with a new story for the new syntax
- The [examples](https://github.com/Instawork/hyperview/tree/master/examples/ui_elements/sectionlist) in the HV demo app should be updated to use the new syntax
- The [XML schema](https://github.com/Instawork/hyperview/blob/master/schema/core.xsd#L835) for `<section-list>` will need to be updated, right now `<section-list>` is only allowed to contain `<behavior>` or `<section>` elements, but now it will need to support `<section-title>` and `<item>` as well.
- The [XML schema](https://github.com/Instawork/hyperview/blob/master/schema/core.xsd#L880) for `<items>` will also need to be updated to support `<section-title>` elements.

hgray-instawork:
Update section-list to not require grouping via \<section\> children

Added a new 'deprecated' example which displays backwards compatibility for including \<section\> elements within a \<section-list\> and as the top element in a reload/refresh page.

Resolves #294

### DIFF
--- a/docs/reference_index.md
+++ b/docs/reference_index.md
@@ -28,9 +28,9 @@ Display elements in HXML can be combined to define the layout of a screen.
 - [`<image>`](/docs/reference_image): Any image content on a screen.
 - [`<list>`](/docs/reference_list): Efficient layout of repeated items on a screen. Supports list-specific interactions like pull-to-refresh.
 - [`<section-list>`](/docs/reference_sectionlist): Efficient layout of sectioned (grouped) repeated items on a screen. Supports list-specific interactions like pull-to-refresh.
-- [`<section>`](/docs/reference_sectionlist): Groups `<item>` elements in a `<section-list>`.
-- [`<section-title>`](/docs/reference_sectiontitle): The header of a `<section>` in a `<section-list>`.
-- [`<item>`](/docs/reference_sectiontitle): An individual item in a `<list>` or `<section-list>`.
+- [`<section>`](/docs/reference_sectionlist): [Deprecated] Groups `<item>` and `<section-title>` elements in a `<section-list>`.
+- [`<section-title>`](/docs/reference_sectiontitle): The header of a group of `<item>` in a `<section-list>`.
+- [`<item>`](/docs/reference_item): An individual item in a `<list>` or `<section-list>`.
 - [`<spinner>`](/docs/reference_spinner): Activity indicator element.
 
 #### Input Elements
--- a/docs/reference_item.md
+++ b/docs/reference_item.md
@@ -4,7 +4,7 @@ title: <item>
 sidebar_label: <item>
 ---
 
-The `<item>` element represents an item in a `<list>` or `<section>`. The structure and attributes of an `<item>` match the `<view>` element.
+The `<item>` element represents an item in a `<list>` or `<section-list>`. The structure and attributes of an `<item>` match the `<view>` element.
 
 Here's a single item in a list:
 
@@ -26,21 +26,19 @@ Here's a single item in a section list:
 
 ```xml
 <section-list style="List">
-  <section>
-    <section-title style="Header">
-      <text>Section 1</text>
-    </section-title>
+  <section-title style="Header">
+    <text>Section 1</text>
+  </section-title>
 
-    <item key="a" style="Item">
-      <view style="Left">
-        <image source="avatar.png" style="Avatar">
-      </view>
-      <view style="Right">
-        <text>Line 1</text>
-        <text>Line 2</text>
-      </view>
-    </item>
-  </section>
+  <item key="a" style="Item">
+    <view style="Left">
+      <image source="avatar.png" style="Avatar">
+    </view>
+    <view style="Right">
+      <text>Line 1</text>
+      <text>Line 2</text>
+    </view>
+  </item>
 </section-list>
 ```
 
@@ -48,7 +46,7 @@ Note that the `key` attribute is required and must be unique among items in the 
 
 ## Structure
 
-An `<item>` element can only apear as a direct child of a `<list>` or `<section>` element. It can contain any non-list element as a child.
+An `<item>` element can only apear as a direct child of a `<list>` or `<section-list>` element. It can contain any non-list element as a child.
 
 ## Attributes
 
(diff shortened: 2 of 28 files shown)


flochtililoch (approved):
Code looks good, thanks for integrating this first change, excited to start using it on our app!
Good to merge once you've verified that the schema still validates the old deprecated behavior. Since the transition period can take a bit of time, we'll want to preserve support for a period of time both behaviors.

hgray-instawork:
Including or omitting 'section' elements within the 'section-list' elements was working as expected but a last-minute removal of a previous change to the core.xsd was causing validation errors when receiving a page of new items which started with a `section`.

flochtililoch (approved):
Latest addition to preserve deprecated behavior in schema (and ensuring CI test it with the example) looks good, thanks for adding those!

PR #518, opened 2023-03-09
feat(section-list): add 'items' to list of 'section-list' child types

hgray-instawork:
`items` is required as an element of `section-list` to allow grouped elements like a loading indicator


flochtililoch (approved):
thanks for the fix!

PR #525, opened 2023-03-28
feat(items): support nested items

hgray-instawork:
- `items` element must support `items` as a child element to allow fully replacing `section`


hgray-instawork:
Converted to draft for now until we determine if we need this functionality

hgray-instawork:
As a replacement for the former <section>, it makes sense. @flochtililoch had some additional thoughts so we decided to hold off and found another way to solve a structural issue we ran into.

PR #527, opened 2023-04-10
Updates xmldom to latest

lenwhite:
This updates xmldom to the latest version, ignoring Instawork's modifications to the library. 

The original context for the fork https://app.asana.com/0/244065166232575/1129162278755512/f is no longer necessary since the error is no longer reproducible on newer versions of webkit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,9 +45,9 @@
     "test": "yarn generate test && yarn test:flow && yarn test:lint && yarn test:render && yarn test:unit && yarn test:validate-xml"
   },
   "dependencies": {
+    "@xmldom/xmldom": "^0.8.7",
     "tiny-emitter": "2.1.0",
-    "url-parse": "1.4.3",
-    "xmldom-instawork": "0.0.1"
+    "url-parse": "1.4.3"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


lenwhite:
I think I misunderstood the original bug then. Was this not a browser-only issue? When would native apps use webkit to render content?

flochtililoch:
The issue had to do with the JSCore runtime (provided by webkit) used by React Native on Android. At instawork we have now used Hermes for a while so the bug might not be reproducible anymore, though it might be worth running some regression tests to ensure it indeed is gone.

adamstep:
> At instawork we have now used Hermes for a while so the bug might not be reproducible anymore, though it might be worth running some regression tests to ensure it indeed is gone.

We are

adamstep:
> The issue had to do with the JSCore runtime (provided by webkit) used by React Native on Android. At instawork we have now used Hermes for a while so the bug might not be reproducible anymore, though it might be worth running some regression tests to ensure it indeed is gone.

Keep in mind other users may not be using Hermes yet, since it's not a requirement. This is making me think though that perhaps we should have a minimum-supported version of RN that is clearly documented.

PR #530, opened 2023-04-20
fix: partial document error state

palaparthi:
PR https://github.com/Instawork/hyperview/pull/504 changed partial document error state which removed reload button, adding the reload button back.

Edit: updated screenshots based on feedback
| error state 1 | error state 2 |
| --- | --- |
| Simulator Screenshot - iPhone 13 - 2023-04-19 at 23 15 57 | Simulator Screenshot - iPhone 13 - 2023-04-19 at 23 30 27 |

### DIFF
--- a/src/core/components/load-element-error/index.js
+++ b/src/core/components/load-element-error/index.js
@@ -29,12 +29,16 @@ const LoadElementError = (props: Props) => {
 
   return (
     <SafeAreaView style={styles.safeArea}>
-      <View style={styles.container}>
-        <Text style={styles.title}>{getError()}</Text>
-        <TouchableOpacity onPress={props.onPressReload}>
-          <Text style={styles.button}>Reload</Text>
-        </TouchableOpacity>
-      </View>
+      <TouchableOpacity onPress={props.onPressReload}>
+        <View style={styles.container}>
+          <View style={styles.textWrapper}>
+            <Text style={styles.title}>{getError()}</Text>
+          </View>
+          <View style={styles.btnWrapper}>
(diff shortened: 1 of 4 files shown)


flochtililoch:
A couple of questions/suggestions:
- Could we adjust the layout a bit so that there's more room around the reload link? I can see this being really hard to tap on the link in both of your screenshots
- Is there a specific reason to show the URL related tot he server error? Does this happen only on dev env? I think we probably shouldn't render this in prod apps
- is the yellow background choice adequate? i.e. why not keeping it white background?

palaparthi:
> * Could we adjust the layout a bit so that there's more room around the reload link? I can see this being really hard to tap on the link in both of your screenshots

Sounds good, I will add some padding and make the entire banner tappable as @adamstep suggested above.
> * Is there a specific reason to show the URL related tot he server error? Does this happen only on dev env? I think we probably shouldn't render this in prod apps

Yes, this is only for dev
> * is the yellow background choice adequate? i.e. why not keeping it white background?

The intention was to highlight the error specifically for the network error when the content is cached (2nd error state above), white background could feel like content rather than a warning. That being said I don't have a strong preference, I can switch it back to white if you think that's better @flochtililoch.

palaparthi:
Screenshot with white background, what do you think @flochtililoch?
simulator_screenshot_5F89C0A6-E26E-44FD-8735-E7553B3F5B36

flochtililoch (approved):
I think latest changes are an improvement over what existed before, so let's get this in, thanks for the updates!

PR #532, opened 2023-04-21
Update example_instawork.md

codetalcott:
Suggest linking to iOS version as it has many more reviews, more informative reviews, and much higher rating.


adamstep (approved):
thanks!

PR #533, opened 2023-04-25
Working typescript branch

lenwhite:
On this branch, `yarn tsc` successfully generates type declaration files without errors. Mostly works.


src/types.ts
```diff
@@ -0,0 +1,438 @@
...
+};
+
+export type NavigationProps = {
+  back: (routeParams?: NavigationRouteParams | null | undefined) => void;
+  closeModal: (routeParams?: NavigationRouteParams | null | undefined) => void;
+  navigate: (
+    routeParams: NavigationRouteParams,
+    key?: string | null | undefined,
+  ) => void;
+  openModal: (routeParams?: NavigationRouteParams | null | undefined) => void;
```
ashwinibm, line 434: Same question here - why are they both optional and maybe types?

.eslintrc
```diff
@@ -1,6 +1,7 @@
 {
   "extends": ["plugin:instawork/recommended"],
   "plugins": ["instawork"],
+  "ignorePatterns": ["**/*.ts", "**/*.tsx"],
```
ashwinibm, line 4: Is this temporary? 
We should not be ignoring since we have added eslint rules for TS also.

flochtililoch (commented):
is there a way to gradually integrate this into master? Ideally we can start leveraging some of the common configuration when we write new code, and I'll also make things easier to review changes if we migrate little by little. I can see this PR being already too big to review.

Issue #290, opened 2021-05-24
Custom behaviors don't observe global attributes like "once", "delay"


adamstep:
This causes issues in particular with amplitude behavior, as some events are logged multiple times while they should be logged only once.

https://github.com/Instawork/hyperview/blob/master/src/index.js#L402

PR #534, opened 2023-04-28
fix(behavior): respect `once` for custom update

palaparthi:
Currently `once` is only respected for [update actions](https://github.com/Instawork/hyperview/blob/master/src/types.js#L390). All the other actions such as `load`, `load-stale-error` etc. ignore `once`.
This causes an infinite render of `HyperScreen` and thereby causing a crash in some situations. The same is true with custom behaviors (https://github.com/Instawork/hyperview/issues/290).

| Before | After |
| --- | --- |
| <video src="https://user-images.githubusercontent.com/28518512/235128036-23e453ad-79e4-4c44-9b02-2e19910552fc.mp4" /> | <video src="https://user-images.githubusercontent.com/28518512/235133854-4cc711f3-ff0d-4255-90de-60085085e493.mp4" /> |

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -289,13 +289,8 @@ export default class HyperScreen extends React.Component {
     return false;
   }
 
-  /**
-   * Sets `ran-once` if applicable.
-   */
-  setOnceIfApplicable = (behaviorElement) => {
-    if (!this.isOncePreviouslyApplied(behaviorElement)) {
-      behaviorElement.setAttribute('ran-once', 'true');
-    }
+  setRanOnce = (behaviorElement) => {
+    behaviorElement.setAttribute('ran-once', 'true');
   }
 
   /**
@@ -371,7 +366,7 @@ export default class HyperScreen extends React.Component {
         return;
       }
 
-      this.setOnceIfApplicable(behaviorElement);
+      this.setRanOnce(behaviorElement);
 
       // Check for event loop formation
       if (trigger === 'on-event') {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -277,6 +277,23 @@ export default class HyperScreen extends React.Component {
...
+
+    if (once === 'true' && ranOnce === 'true') {
+      return true;
+    } 
+    if (once === 'true') {
+      behaviorElement.setAttribute('ran-once', 'true');
+    }
+
+    return false;
+  }
```
hgray-instawork, line 295: Maybe:
```
if (once === 'true') {
    if (ranOnce === 'true') {
        return true;
    }
    behaviorElement.setAttribute('ran-once', 'true');
}
return false;
```

hgray-instawork (approved):

src/index.js
```diff
@@ -277,6 +277,23 @@ export default class HyperScreen extends React.Component {
     );
   }
 
+  /**
+   * Checks if `once` is previously applied and sets `ran-once` if required.
+   */
+  isOncePreviouslyApplied = (behaviorElement) => {
```
flochtililoch, line 283: I think the method name could better describe that it does mutate the element.

adamstep: Difficulty with naming and the fact that this function is both checking and mutating state is a strong sign it should be two separate functions.

palaparthi: Addressed in https://github.com/Instawork/hyperview/pull/534/commits/92114e701835884b5b07b862056652cef402c106

PR #543, opened 2023-05-18
feat(error) - adding an HvBaseError into Hyperview

hgray-instawork:
- Created new HvBaseError class
- Created tests for base error

Borrowed heavily from https://github.com/Instawork/mobile/blob/master/core-mobile/src/services/errors/index.test.js

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -54,7 +54,20 @@
     },
     {
       "files": ["**/*.ts", "**/*.tsx"],
-      "extends": ["plugin:instawork/recommended-typescript"]
+      "extends": ["plugin:instawork/recommended-typescript"],
+      "rules": {
+        "instawork/flow-annotate": "off",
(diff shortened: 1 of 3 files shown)


.eslintrc
```diff
@@ -54,7 +54,20 @@
     },
     {
       "files": ["**/*.ts", "**/*.tsx"],
-      "extends": ["plugin:instawork/recommended-typescript"]
+      "extends": ["plugin:instawork/recommended-typescript"],
```
flochtililoch, line 57: does this need a rebase with master?

src/services/error/index.ts
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable instawork/error-object */
```
flochtililoch, line 1: I missed this on my first review - I think we'll want this error to be disabled for the entire repo (in the .eslintrc file)

flochtililoch (approved):

PR #545, opened 2023-05-19
feat(schema) - Updated schema for <navigator> and <nav-route>

hgray-instawork:
Updated schema for new and elements with the following rules:

- `<doc>` may contain either `<screen>` or `<navigator>` but not both
- `<navigator>` must contain one or more `<nav-route>`
  - must have an 'id' attribute
  - must have a 'type' attribute with a value of 'stack, bottom-tab, top-tab'
- `<nav-route>` may contain zero or one `<navigator>`
  - must have an 'id' ID attribute
  - must have an 'href' string attribute
  - may have an 'initial' boolean attribute

https://app.asana.com/0/0/1204637347494053/f

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -48,7 +48,8 @@
     {
       "files": ["**/types.js"],
       "rules": {
-        "no-use-before-define": "off"
+        "no-use-before-define": "off",
+        "max-len": "off"
       }
     },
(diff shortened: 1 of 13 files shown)


adamstep (approved):
looks good!

PR #553, opened 2023-05-24
fix(flow) - Ignore ts/x files in flow

hgray-instawork:
Updated the config to ignore .ts and .tsx files


hgray-instawork:
This seems to not be necessary and is causing some other issues.

PR #556, opened 2023-05-25
feat(navigation) - Required packages

hgray-instawork:
Imported required packages into Hyperview and /demo

https://app.asana.com/0/0/1204660853298346/f


package.json
```diff
@@ -45,7 +45,14 @@
     "test": "yarn generate test && yarn test:flow && yarn test:lint && yarn test:render && yarn test:unit && yarn test:validate-xml"
   },
   "dependencies": {
+    "@react-navigation/bottom-tabs": "6.5.7",
```
flochtililoch, line 48: I think we might want to make these peer-dependencies in case we don't get to inject these before we release

flochtililoch (approved):

PR #557, opened 2023-05-25
feat(navigation) - Add target support

hgray-instawork:
Adding support to pass target values into navigation actions

https://app.asana.com/0/0/1204660853298351/f

### DIFF
--- a/src/core/components/hv-screen/index.js
+++ b/src/core/components/hv-screen/index.js
@@ -344,7 +344,7 @@ export default class HvScreen extends React.Component {
   /**
    *
    */
-  onUpdate = (href, action, currentElement, opts) => {
+  onUpdate = (href, action, currentElement, opts, target) => {
     if (action === ACTIONS.RELOAD) {
       this.reload(href);
     } else if (action === ACTIONS.DEEP_LINK) {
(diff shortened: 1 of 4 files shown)


src/core/components/hv-screen/index.js
```diff
@@ -344,15 +344,15 @@ export default class HvScreen extends React.Component {
   /**
    *
    */
-  onUpdate = (href, action, currentElement, opts) => {
+  onUpdate = (href, action, currentElement, opts, target) => {
```
flochtililoch, line 347: I think this function is called all over the place by custom behaviors, so I think we might want to make it a key of the `opts` object instead, so that we don't break compatibility

src/types.js
```diff
@@ -404,12 +404,14 @@ export type BehaviorOptions = {|
   behaviorElement: Element,
   showIndicatorId?: string,
   delay?: number,
+  targetId: ?string,
```
flochtililoch, line 407: the presence of the key should be made optional, but when present, it's value should be set - you can do that by moving the `?` before `:`:
```suggestion
  targetId?: string,
```

flochtililoch (approved):

PR #558, opened 2023-05-25
feat(navigation) - Demo examples

hgray-instawork:
These are not the final examples but are a WIP to allow testing complex nested navigators and navigation actions

- Updated starting point for demo to use a version with a root level navigator
- Updated sync file to change name to the new 'index.tsx'

https://app.asana.com/0/0/1204660853298348/f

### DIFF
--- a/demo/src/Root.tsx
+++ b/demo/src/Root.tsx
@@ -6,16 +6,19 @@
  *
  */
 
-import { PureComponent, useState } from 'react';
-import { Button, SafeAreaView, Text, View } from 'react-native';
+import { PureComponent } from 'react';
+import { Button, SafeAreaView, View } from 'react-native';
 import NavigatorLegacy from './NavigatorLegacy';
 import Navigator from './Navigator';
+
+type Props = {};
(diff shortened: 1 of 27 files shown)


schema/core.xsd
hgray-instawork, line 1: I don't know why this is showing as changed. I rebased from the integration branch after this was merged.

hgray-instawork:
I'm going to hold off merging until the rest of the components are in

demo/src/Root.tsx
```diff
@@ -6,16 +6,19 @@
...
+type Props = {};
+type State = { useLegacy: boolean };
+
 /**
  * Branch the demo/example to use new navigation or old navigation
  */
-export default class Root extends PureComponent {
-  constructor(props) {
+export default class Root extends PureComponent<Props, State> {
+  constructor(props: Props | Readonly<Props>) {
```
flochtililoch, line 21: I think props should always be read-only. Is this how React recommends typing props in TypeScript?

flochtililoch (approved):

PR #561, opened 2023-05-30
fix(fetch-element): Don't update the state when fetch fails

palaparthi:
When fetching a partial fails then don't update the state with new doc.

### Videos
| Before | After |
| --- | --- |
| <video src="https://github.com/Instawork/hyperview/assets/28518512/8c885640-2c7c-4741-ab9f-7fcbce12cf91" /> | <video src="https://github.com/Instawork/hyperview/assets/28518512/47ee48c5-590f-4746-bb66-0386c062a71a" /> |

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -447,6 +447,10 @@ export default class HyperScreen extends React.Component {
     // Fetch the resource, then perform the action on the target and undo indicators.
     const fetchAndUpdate = () => this.fetchElement(href, verb, newRoot, formData)
       .then((newElement) => {
+        // If the fetch failed, don't update the doc state.
+        if (!newElement) {
+          return;
+        }
         // If a target is specified and exists, use it. Otherwise, the action target defaults
         // to the element triggering the action.
         let targetElement = targetId ? this.doc.getElementById(targetId) : currentElement;
@@ -454,9 +458,7 @@ export default class HyperScreen extends React.Component {
           targetElement = currentElement;
         }
 
-        if (newElement) {
-          newRoot = Behaviors.performUpdate(action, targetElement, newElement);
-        }
+        newRoot = Behaviors.performUpdate(action, targetElement, newElement);
         newRoot = Behaviors.setIndicatorsAfterLoad(showIndicatorIdList, hideIndicatorIdList, newRoot);
         // Re-render the modifications
         this.setState({


palaparthi:
> @palaparthi was this a regression introduced at some point? Since it seemed like this functionality was working a few weeks ago.

I think we got lucky with the order of requests made. When there are multiple `setState` happening simultaneously, there seems to be a race condition where if the last request fails to fetch any data the app crashes. In the video I posted above the last request is `notifications/bell` which fails and causes the app to crash. This request wasn't happening on local so we weren't able to catch this before.

Regardless I think this fix makes sense imo because when a partial fetch fails there is no point trying to update the doc with the same root.

adamstep (approved):

src/index.js
```diff
@@ -447,16 +447,18 @@ export default class HyperScreen extends React.Component {
     // Fetch the resource, then perform the action on the target and undo indicators.
     const fetchAndUpdate = () => this.fetchElement(href, verb, newRoot, formData)
       .then((newElement) => {
+        // If the fetch failed, don't update the doc state.
+        if (!newElement) {
```
flochtililoch, line 451: if we return at this point, doesn't this cause an issue with loading indicators not being cleared properly?

palaparthi: Since the fetch was unsuccessful maybe it is ok to not clear the indicators?

flochtililoch: I don't think that's right, indicators are meant to represent some activity is pending, but we're not auto-retrying calls as I understand it

palaparthi: Agreed, this is a known limitation (showing loading state when request fails) for now. Do you have any recommendation for this?

flochtililoch: I'm not sure what you mean - I think we do hide the loading state (and custom loading indicators) when the request fail right now, don't we?

palaparthi: Not in all cases, for example when loading a sectionlist if the next page fails to load we keep showing the spinner. 

https://github.com/Instawork/hyperview/assets/28518512/534b18d7-7bec-4214-9807-b50681b91293

flochtililoch: Right, I think that's a bug in the component that wraps the section list, which assumes the request will be succesful (the spinner is always visible instead of being shown when the request kicks off). This is not related to the way Hyperview handles loading indicators.

palaparthi: Thanks for clarifying. I will put up a follow up pr to clear the loading indicators once I figure out and fix the root cause. But for now I think this is ok because with the previous implementation the app was crashing.

PR #562, opened 2023-05-30
feat(navigation) - Contexts and types

hgray-instawork:
- Cleaned up the package imports
- Ported required contexts into new `navigation-legacy.ts` file
- Created additional required contexts
  - navigation: provides fetch, entry-point url, as well as behaviors and components
  - navigator: provides runtime storage of urls and nested navigators which could not be easily assigned to route params
-  created /types.ts files for all of the new components which strong-type their expected props. The props are broken into smaller chunks and use composition to allow fine control over which data is expected at each point
- Updated src/index.tsx with new contexts
- Placeholder <nav-route>


src/contexts/navigator.tsx
```diff
@@ -0,0 +1,106 @@
...
+      'useMapContext must be used within a NavigatorMapContext',
+    );
+  }
+  return context;
+}
+
+/**
+ * Cache the initial route name for the <navigator>
+ */
+function SetInitialRouteName(name: string) {
```
flochtililoch, line 41: by convention, function names are `camelCase` - we reserve capitalization for the React functional components (also apply to other functions in this file)
```suggestion
function setInitialRouteName(name: string) {
```

src/contexts/navigator.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+function GetRouteNavigator(key: string): Element | undefined {
+  const context = useMapContext();
+  return context?.elementMap?.get(key);
+}
+
+/**
+ * Encapsulated context provider
+ */
+// eslint-disable-next-line react/destructuring-assignment
```
flochtililoch, line 86: let's not disable this rule, it's intention is to ensure we make if obvious what is a prop vs other variables from the local scope

src/contexts/navigator.tsx
```diff
@@ -25,7 +25,7 @@ export type NavigatorCache = {
  */
 const NavigatorMapContext = createContext<NavigatorCache | null>(null);
 
-function useMapContext() {
+function UseMapContext() {
```
hgray-instawork, line 28: Note that eslint is requiring the use of uppercase for the `UseMapContext` due to its inclusion in the other functions.

eslint[react-hooks/rules-of-hooks](https://reactjs.org/docs/hooks-rules.html)

flochtililoch (approved):
one question about imports, rest looks good

src/services/navigation/index.js
```diff
@@ -72,9 +72,12 @@ export default class Navigation {
     const { showIndicatorId, delay, targetId } = opts;
     const formData: ?FormData = getFormData(element, formComponents);
 
-    // Serialize form data as query params, if present.
-    const baseUrl = UrlService.getUrlFromHref(href, this.url);
-    const url = UrlService.addFormDataToUrl(baseUrl, formData);
```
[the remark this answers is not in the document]
hgray-instawork: Correct, the fragments are only for pointing to routes within the navigation hierarchy. Should I throw a warning if formData != null?

src/index.tsx
```diff
@@ -6,21 +6,20 @@
...
-// import * as Namespaces from 'hyperview/src/services/namespaces';
+import * as HvScreenProps from './core/components/hv-screen/types';
+import { DateFormatContext, RefreshControlComponentContext } from './contexts';
 import React, { PureComponent } from 'react';
-import HvNavigator from 'hyperview/src/core/components/hv-navigator';
-import HvScreen from 'hyperview/src/core/components/hv-screen';
-import { Props } from 'hyperview/src/core/components/hv-navigator/types';
+import HvRoute from './core/components/hv-route';
+import HvScreen from './core/components/hv-screen';
+import { NavigationContext } from './contexts/navigation';
```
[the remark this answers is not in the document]
hgray-instawork: Yup, we can, I'll switch it over

PR #564, opened 2023-05-31
feat(navigation) - Implementations

hgray-instawork:
- Implementations of `hv-route`, `hv-navigator`
- Support files for navigator including navigation logic, errors, rendering the correct route type


src/core/components/hv-navigator/index.tsx
```diff
@@ -6,43 +6,264 @@
...
+            component={HvRoute}
+            getId={({ params }) => params?.url}
+            // empty object required because hv-screen doesn't check for undefined param
+            initialParams={{}}
+            name={NavigatorService.ID_DYNAMIC}
+          />,
+        );
+
+        // Modal is used to display all routes in stack which are presented as modals
+        screens.push(
```
flochtililoch, line 150: are we missing a `case` statement here?

src/core/components/hv-navigator/index.tsx
```diff
@@ -6,43 +6,264 @@
...
 
+    // Add the dynamic screens
+    switch (type) {
+      case NavigatorService.NAVIGATOR_TYPE.STACK:
+        // Dynamic is used to display all routes in stack which are presented as cards
+        screens.push(
+          <Stack.Screen
+            key={NavigatorService.ID_DYNAMIC}
+            component={HvRoute}
+            getId={({ params }) => params?.url}
```
flochtililoch, line 142: are there any risk for `param.url` to be undefined, in which case the id would also be undefined?

src/core/components/hv-route/index.tsx
```diff
@@ -6,10 +6,152 @@
...
+    }
+    if (this.state.doc) {
+      return <ContentView />;
+    }
+    return <LoadingView />;
+  }
+}
 
 /**
  * Functional component wrapper around HvRouteInner
```
flochtililoch, line 157: is there a particular reason why we need a inner component + a wrapper, instead of a single component?

src/core/components/hv-route/index.tsx
```diff
@@ -18,10 +160,59 @@ import React, { Text, View } from 'react-native';
...
+    element = navCache.elementMap?.get(props.route?.params?.id);
+  }
+
   return (
-    <View>
-      <Text>ROUTE</Text>
-    </View>
+    <HvRouteInner
+      // eslint-disable-next-line react/jsx-props-no-spreading
+      {...{ ...props, ...navProps, ...navCache }}
```
flochtililoch, line 206: ideally we should avoid prop spreading as it obfuscate what we pass from one component to another. But maybe here if we combine both component we might not need this

src/core/components/hv-screen/types.ts
```diff
@@ -60,19 +63,26 @@ export type DataProps = {
  * Props used by legacy external navigation components
  */
 export type ActionProps = {
-  back?: () => void;
-  closeModal?: () => void;
+  back?: (params: object) => void;
+  closeModal?: (params: object) => void;
```
flochtililoch, line 67: don't we know the exact types of params we pass to back/closeModal?

src/services/dom.d.ts
```diff
@@ -0,0 +1,29 @@
...
+import type { Document } from 'hyperview/src/types-legacy';
+
+/**
+ * TS definition of the Parser class to provide an alternate Document response
+ */
+export class Parser {
+  constructor(
+    fetch: Fetch,
+    onBeforeParse: ?BeforeAfterParseHandler,
+    onAfterParse: ?BeforeAfterParseHandler,
```
flochtililoch, line 23: are these the right way to define optional params? I thought the leading `?` was flow-type specific.

src/services/navigator/helpers.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (element?.childNodes?.length) {
+    for (let i = 0; i < element?.childNodes?.length; i += 1) {
+      const child: Types.Element = element.childNodes[i];
+      if (child.nodeType === Types.NODE_TYPE.ELEMENT_NODE) {
+        elements.push(child);
+      }
+    }
+  }
+  return elements;
+};
```
flochtililoch, line 27: more succinct version of this code:
```suggestion
export const getChildElements = (element: Types.Element): Types.Element[] =>
  (element.childNodes || []).filter(
    (child: Types.Element) => child.nodeType === Types.NODE_TYPE.ELEMENT_NODE,
  );

```

src/services/navigator/index.ts
```diff
@@ -0,0 +1,319 @@
...
+    switch (navAction) {
+      case TypesLegacy.NAV_ACTIONS.BACK:
+        if (routeParams) {
+          this.routeBackRequest(navigation, routeParams);
+        } else {
+          navigation.goBack();
+        }
+        break;
+      case TypesLegacy.NAV_ACTIONS.CLOSE:
+        navigation.goBack();
```
flochtililoch, line 261: we should implement the routeParams stuff for the close action too

src/services/navigator/index.ts
```diff
@@ -0,0 +1,319 @@
...
+      case TypesLegacy.NAV_ACTIONS.NAVIGATE:
+        if (routeId) {
+          navigation.dispatch(Imports.CommonActions.navigate(routeId, params));
+        }
+        break;
+      case TypesLegacy.NAV_ACTIONS.NEW:
+        if (routeId) {
+          navigation.dispatch(Imports.CommonActions.navigate(routeId, params));
+        }
+        break;
```
flochtililoch, line 272: we can combine these two actions

src/services/navigator/index.ts
```diff
@@ -0,0 +1,319 @@
...
+  back = (params: TypesLegacy.NavigationRouteParams) => {
+    this.sendRequest(TypesLegacy.NAV_ACTIONS.BACK, params);
+  };
+
+  closeModal = (params: TypesLegacy.NavigationRouteParams) => {
+    this.sendRequest(TypesLegacy.NAV_ACTIONS.CLOSE, params);
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars, no-unused-vars
+  navigate = (params: TypesLegacy.NavigationRouteParams, _: string) => {
```
flochtililoch, line 291: do we need to define the second argument if it's not used?

src/services/navigator/render.tsx
```diff
@@ -0,0 +1,144 @@
...
+          push={props.navLogic.push}
+          route={route}
+          url={props.url || undefined}
+        />
+      )}
+    </Contexts.DateFormatContext.Consumer>
+  );
+};
+
+export const Route = (props: RouteRenderProps): React.ReactElement => {
```
flochtililoch, line 76: I don't know if the code in this file belong to services, since both functions are React components. They would be better suited in src/core/components or src/components

flochtililoch (approved):
I added a couple issues that I think are bugs or missing functionality, that we should address in priority, then also a couple suggested refactors that are less urgent, though I think we should put them in our backlog of things to address. Ideally, we should revisit the new code in general to do less variable mutations, and be more functional - I think this should help a great deal with readability and the task of adding unit tests.

src/services/navigator/index.ts
```diff
@@ -0,0 +1,297 @@
...
+    url: string | undefined,
+    isStatic: boolean,
+  ): string => {
+    switch (action) {
+      case NAV_ACTIONS.PUSH:
+        return ID_DYNAMIC;
+      case NAV_ACTIONS.NEW:
+        return ID_MODAL;
+      default:
+    }
```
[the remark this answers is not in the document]
hgray-instawork: This is fixed in https://github.com/Instawork/hyperview/tree/hardin/navigator-implementation-cleanup

src/core/components/hv-route/index.tsx
hgray-instawork: This was part of the rapid development where we decided against rewriting the whole thing. I can work on removing the class for this implementation

src/core/components/hv-screen/types.ts
hgray-instawork: We do but I wanted to be consistent with the other implementations. If you'd prefer I strong type them, I'll switch them back. This seemed the best way to continue to support legacy functionality.

src/services/dom.d.ts
hgray-instawork: You'd have to ask the author of that particular file ;).

I'll change it.

src/services/navigator/index.ts
hgray-instawork: Done in the new branch

src/services/navigator/render.tsx
hgray-instawork: Fair point, I'll look at moving it

src/core/components/hv-navigator/index.tsx
hgray-instawork: No missing case statement (now if clause in updated code). The two screens (dynamic and modal) are added to every navigator of type STACK

src/core/components/hv-navigator/index.tsx
hgray-instawork: If the value is null, it's the same as there being no 'getId' assigned, which means it will be referenced by its name. This means the initial two stack screens will be 'dynamic' and 'modal'. I will add additional checks against a null url when navigating/pushing additional screens onto the stack.

src/services/navigator/helpers.ts
hgray-instawork: While this is prettier, childNodes is of type NodeList<T> which doesn't support iteration or filtering.

https://github.com/Instawork/hyperview/blob/d4077db6950abac8fad28ae30a19991b97b62a5a/src/types.js#L224-L227

src/services/navigator/index.ts
```diff
@@ -0,0 +1,297 @@
...
+  props: Props;
+
+  constructor(props: Props) {
+    this.props = props;
+  }
+
+  /**
+   * Recursively search down the navigation path for the target in state and build a path to it
+   */
+  findPath = (state: NavigationState, targetId: string, path: string[]) => {
```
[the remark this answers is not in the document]
hgray-instawork: Rewritten without mutations

src/core/components/hv-navigator/index.tsx
```diff
@@ -6,43 +6,264 @@
...
-    if (!this.state.doc) {
-      const loadingScreen = this.context.loadingScreen || Loading;
-      return React.createElement(loadingScreen);
+
+    const elements: TypesLegacy.Element[] = NavigatorService.getChildElements(
+      element,
+    );
+    const { buildTabScreen } = this;
+    for (let i = 0; i < elements.length; i += 1) {
+      const child: TypesLegacy.Element = elements[i];
```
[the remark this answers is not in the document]
hgray-instawork: Fixed

src/core/components/hv-navigator/index.tsx
hgray-instawork: Updated to enforce urls in params and to check for empty urls

src/core/components/hv-route/index.tsx
hgray-instawork: The function provides access to multiple contexts. We had to wrap the class implementation inside the function to allow passing the multiple contexts' information in.

src/core/components/hv-route/index.tsx
hgray-instawork: We'll keep for now as it's a requirement of passing the props along with the context data into the class

src/core/components/hv-screen/types.ts
hgray-instawork: Updated to support strong or weak typed

src/services/dom.d.ts
hgray-instawork: Fixed.

src/services/navigator/helpers.ts
hgray-instawork: Fixed with Array.From()

src/services/navigator/index.ts
hgray-instawork: Fixed.

src/services/navigator/render.tsx
hgray-instawork: Reworked to move them into hv-route

PR #569, opened 2023-06-02
feat(navigator) - improvements

hgray-instawork:
Improvements and bug fixes based on comments in previous PR.

https://github.com/Instawork/hyperview/pull/564


src/core/components/hv-navigator/index.tsx
```diff
@@ -129,38 +132,34 @@ export default class HvNavigator extends PureComponent<Props> {
...
-            component={HvRoute}
-            getId={({ params }) => params?.url}
-            // empty object required because hv-screen doesn't check for undefined param
-            initialParams={{}}
-            name={NavigatorService.ID_DYNAMIC}
-          />,
-        );
-
-        // Modal is used to display all routes in stack which are presented as modals
-        screens.push(
```
flochtililoch, line 150: I had commented about this in your previous PR, it look like we were missing a "case" statement here - the refactor doesn't seem to fix that

src/core/components/hv-route/index.tsx
```diff
@@ -137,73 +277,78 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, State> {
...
     if (this.state.error) {
-      return <ErrorView error={this.state.error} />;
+      return <ErrorComponent error={this.state.error} />;
     }
     if (this.state.doc) {
-      return <ContentView />;
+      return <ContentComponent />;
     }
-    return <LoadingView />;
+    return <LoadingComponent />;
```
flochtililoch, line 292: minor: components typically don't use the "Component" suffix in their name

src/services/navigator/helpers.ts
```diff
@@ -92,3 +93,132 @@ export const getUrlFromHref = (
...
+ */
+export const findPath = (
+  state: Imports.NavigationState,
+  targetId: string,
+): string[] => {
+  let path: string[] = [];
+  if (!state) {
+    return path;
+  }
+  console.log('state', state);
```
flochtililoch, line 126: to do: remove

src/core/components/hv-navigator/index.tsx
hgray-instawork: Both screens are required for stack navigator, so both are added. No case statement needed here.

src/services/navigator/helpers.ts
```diff
@@ -92,3 +93,131 @@ export const getUrlFromHref = (
...
+    }
+    param = { screen };
+    param.params = buildParams(routeId, path, routeParams, index + 1);
+  } else {
+    param = { screen: routeId };
+    // The last screen in the path receives the route params
+    // example: { url: 'someurl.xml' }
+    param.params = routeParams;
+  }
+  return param;
```
flochtililoch, line 204: minor: remove mutation

```suggestion
  if (path.length && index < path.length) {
    const screen = path[index];

    if (!screen) {
      throw new Errors.HvNavigatorError('screen is undefined');
    }
    return {
      screen,
      params: buildParams(routeId, path, routeParams, index + 1);
    }
  } else {
    return {
      screen: routeId,
      // The last screen in the path receives the route params
      // example: { url: 'someurl.xml' }
      params: routeParams,
    };
  }
```

flochtililoch (approved):

src/services/navigator/helpers.ts
hgray-instawork: I've replaced this and several other spots where mutation was occurring

demo/App.js
```diff
@@ -6,6 +6,7 @@
  *
  */
 
+import 'react-native-gesture-handler';
```
flochtililoch, line 9: Checkout the module `src/gesture-handler` which does that same import, but does have an escape hatch for web. This reminds me we should test the demo on web too.

flochtililoch (approved):

PR #572, opened 2023-06-09
chore(examples) - fix document formatting

hgray-instawork:
Formatted all *.njk files using `Nunjucks Template`: https://marketplace.visualstudio.com/items?itemName=eseom.nunjucks-template#overview

No other changes made

### DIFF
--- a/examples/_includes/directory_list.xml.njk
+++ b/examples/_includes/directory_list.xml.njk
@@ -1,20 +1,7 @@
 {# Standard template for showing a list of links to other screens. List is generated from a collection of screens with the given "list_tag" #}
-
-<list
-  xmlns="https://hyperview.org/hyperview"
-  trigger="refresh"
-  action="replace"
-  target="directory-list"
-  id="directory-list"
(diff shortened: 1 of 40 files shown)


flochtililoch:
I think if we're going to have some linting on these files, we should have them linted on CI too. Doesn't Prettier support formatting for njk files?

hgray-instawork:
Prettier is not doing any formatting for me. I looked around and found a few node modules that purported to add formatting support for prettier but they seemed half-baked.

I added a local document override to my VSCode setup:
"[njk]": {
    "editor.defaultFormatter": "eseom.nunjucks-template"
  }

Adam and I discussed this would be a starting point to make the document consistent as we start making changes. Let me know if you have a better suggestion.

adamstep:
Agreed, linting checks through CI would be ideal, but @hgray-instawork does formatting locally for now, I'm personally ok with that as a short-term solution.

flochtililoch (approved):

PR #576, opened 2023-06-14
chore(navigator) - moving dependencies

hgray-instawork:
Moved the @react-navigation/native dependency into peer and dev

https://app.asana.com/0/0/1204801986148220/f

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "@react-navigation/stack": "6.3.16",
     "react": ">= 16.13.1",
     "react-native": ">= 0.63.3",
-    "react-native-gesture-handler": "2.10.1",
+    "react-native-gesture-handler": "2.11.0",
     "react-native-keyboard-aware-scrollview": ">= 2.1.0",
     "react-native-safe-area-context": "4.2.4",
     "react-native-webview": ">= 10.10.2"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -55,11 +54,13 @@
     "@react-native-picker/picker": ">= 2.2.1",
     "@react-navigation/bottom-tabs": "6.5.7",
     "@react-navigation/core": "6.4.8",
+    "@react-navigation/native": "6.1.6",
     "@react-navigation/stack": "6.3.16",
     "react": ">= 16.13.1",
     "react-native": ">= 0.63.3",
     "react-native-gesture-handler": "2.10.1",
     "react-native-keyboard-aware-scrollview": ">= 2.1.0",
+    "react-native-safe-area-context": "4.2.4",
```
flochtililoch, line 63: Does this need to be added since it wasn't part of `dependencies`? if so, should it also be added to `devDependencies`?

flochtililoch (commented):
after addressing the comment, can you run `yarn` so that we also include the updated yarn.lock file in the change?

package.json
hgray-instawork: I don't use it directly in Hyperview so I didn't have it as a dependency or dev dependency. The navigators need it. It's in the demo package so I figured it was better to explicitly mark it as a peer dependency.

flochtililoch (approved):

PR #579, opened 2023-06-15
feat: alert button style

flochtililoch:
Allow alert option button style to be defined.

Simulator Screenshot - iPhone 14 Pro - 2023-06-15 at 18 38 23


adamstep (commented):
Should you update the schema?

PR #580, opened 2023-06-16
fix: Validate generated XML against schema

flochtililoch:
Use Eleventy generated files to validate XML instead of template files.


adamstep (approved):
Thanks!

PR #581, opened 2023-06-16
fix: scrollbar bug with iOS 13+

flochtililoch:
Follow up for #413

Apply the same fix for `list` and `section-list` that was originally applied to scrollable `view`.

|  | Before | After |
|--|--------|-------|
| `list` | before-list | after-list |
| `section-list` | before-section-list | after-section-list |


adamstep:
🎉 🎉

palaparthi (approved):

PR #582, opened 2023-06-16
fix: ignore Eleventy generated files from Prettier

flochtililoch:
Before this change, one would need to delete the `_examples_site` directory before being able to run Prettier tests. This change also removes the trailing slash in the command that validates XML (doesnt change the test behavior, but makes the output of validated file paths correct).

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,6 @@
 flow-typed
 *.gen.js
+_examples_site/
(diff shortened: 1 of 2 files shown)


hgray-instawork (approved):
Thank you for this! This tripped me up when I first got going in the repo

PR #585, opened 2023-06-22
fix(navigator) - Stack navigator with defined sub-navigator

hgray-instawork:
Resolve an issue where subsequent dynamic screens were reloading the defined sub-navigator rather than using the passed url.

Asana: https://app.asana.com/0/0/1204865667381531/f

### DIFF
--- a/src/core/components/hv-route/index.tsx
+++ b/src/core/components/hv-route/index.tsx
@@ -362,9 +362,7 @@ export default function HvRoute(props: Types.Props) {
   const id: string | undefined =
     props.route?.params?.id || navigatorContext.initialRouteName || undefined;
 
-  const state = props.navigation?.getState();
-  const index = state?.index;
-  const type = state?.type;
+  const { index, type } = props.navigation?.getState() || {};
   // The nested element is only used when there is an id
   //    and the navigator is not a stack or is the first screen in a stack
   //    other stack screens will use a url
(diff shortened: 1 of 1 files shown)


src/core/components/hv-route/index.tsx
```diff
@@ -362,8 +362,17 @@ export default function HvRoute(props: Types.Props) {
   const id: string | undefined =
     props.route?.params?.id || navigatorContext.initialRouteName || undefined;
 
+  const state = props.navigation?.getState();
+  const index = state?.index;
+  const type = state?.type;
```
flochtililoch, line 367: minor/more concise way to write this:
```suggestion
  const { index, type } = props.navigation?.getState() || {};
```

src/core/components/hv-route/index.tsx
```diff
@@ -362,8 +362,17 @@ export default function HvRoute(props: Types.Props) {
...
+  // The nested element is only used when there is an id
+  //    and the navigator is not a stack or is the first screen in a stack
+  //    other stack screens will use a url
+  const includeElement =
+    id && (type !== NavigatorService.NAVIGATOR_TYPE.STACK || index === 0);
```
flochtililoch, line 372: nit: it look like includeElement is used as a boolean, so we might want to explicitly type it, which in turns should require the expression to be reworked as such:
```suggestion
  const includeElement: boolean =
    !!id && (type !== NavigatorService.NAVIGATOR_TYPE.STACK || index === 0);
```

flochtililoch (approved):

src/core/components/hv-route/index.tsx
hgray-instawork: I've reworked this bit. I had to move the id out to allow the .get(id) to evaluate properly

flochtililoch (approved):

PR #594, opened 2023-06-27
fix(navigation) - Remove cyclic dependency between HvRoute and HvNavigator

hgray-instawork:
- Updated HvNavigator to receive HvRoute through a getter prop

Asana: https://app.asana.com/0/0/1204845198486199/f


hgray-instawork:
I guess we don't actually need to do this through a getter function. It could just be passed as a param, but I thought this might be more correct.

flochtililoch (approved):

PR #596, opened 2023-06-28
fix(navigation) - Allow injecting a doc into HvScreen

hgray-instawork:
- Provide a document to HvScreen when it is created from HvRoute
- Use the document once in lieu of loading from url
- Remove the initial document so it won't be used after initialization completes

Asana: https://app.asana.com/0/1204008699308084/1204911836776291/f


flochtililoch (approved):

flochtililoch (changes requested):
Looks like this is causing a regression:

https://github.com/Instawork/hyperview/assets/309515/d0387bf2-314c-4431-a05a-bb8f72d4160b

hgray-instawork:
Current resolution is to pass a clone of the document to remove the binding between HvRoute and HvScreen

flochtililoch (approved):

PR #598, opened 2023-07-03
feature(navigation) - Add preload functionality to HvRoute

hgray-instawork:
Adding preload screen functionality to HvRoute since it is now handling loading
- Added global preload screen registry to navigator context
  - In legacy code, this is local to the screen as each screen instantiates its own navigation service. This change allows the preload registry to continue to live in the local navigation service and also in a global context available to all HvRoutes in the hierarchy.
  - provided callback to allow HvScreen+local navigation service to also register in the global context registry
- Added preloadScreen to route params
  - previous requirement included only id and url
- Added preload screen rendering into HvRoute
  - HvRoute now performs the loading which HvScreen previously performed
- Created description files for services required for preload render
  - Components and Stylesheet now have *.d.ts files

Asana: https://app.asana.com/0/0/1204952399375930/f

| With preload  | No preload |
| ------------- | ------------- |
|  with  | without  |

### DIFF
--- a/src/services/navigation/index.js
+++ b/src/services/navigation/index.js
@@ -68,6 +68,7 @@ export default class Navigation {
     element: Element,
     formComponents: ComponentRegistry,
     opts: BehaviorOptions,
+    registerPreload?: (id: number, element: Element) => void,
   ): void => {
     const { showIndicatorId, delay, targetId } = opts;
     const formData: ?FormData = getFormData(element, formComponents);
@@ -91,6 +92,9 @@ export default class Navigation {
       if (loadingScreen) {
         preloadScreen = Date.now(); // Not trully unique but sufficient for our use-case
         this.setPreloadScreen(preloadScreen, loadingScreen);
+        if (registerPreload) {
+          registerPreload(preloadScreen, loadingScreen);
+        }
       }
     }
 
--- a/src/contexts/navigator.tsx
+++ b/src/contexts/navigator.tsx
@@ -13,6 +13,7 @@ export type NavigatorContextProps = {
   routeMap?: Map<string, string>;
   elementMap?: Map<string, TypesLegacy.Element>;
   initialRouteName?: string;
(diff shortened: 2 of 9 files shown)


adamstep:
Can you include an example in the description of how this functionality would be used (from the XML)? I'm having a slightly hard time visualizing it.

hgray-instawork:
Assigning a preload screen works as it always has. See this example: https://github.com/Instawork/instawork/blob/48cedc24a85ee66a876e53175889f65e857b051f/apps/mobile/templates/mobile/components/screen/stories.xml#L12C1-L32C15

The change made in this code is to globalize the lookup map. Previously this was hosted in the individual Notification service which was owned by each screen. It still lives there for legacy functionality, but is now also held in a context at the top of the navigation hierarchy. HvRoute was updated with render capabilities borrowed from HvScreen to allow it to render the preload screen.

src/services/navigation/index.js
```diff
@@ -68,6 +68,7 @@ export default class Navigation {
     element: Element,
     formComponents: ComponentRegistry,
     opts: BehaviorOptions,
+    registerPreload: ?(id: number, element: Element) => void,
```
hgray-instawork, line 71: Does this need a question mark before AND after the colon? I've seen it three different ways in the Javascript

adamstep: Before the colon means the key (`registerPreload`) is optional, it doesn't need to be included in the object. But if the key is included, the value must be a callback:
```js
// registerPreload?: (id: number, element: Element) => void

{ } // valid
{ registerPreload: (id, element) => {} } // valid
{ registerPreload: undefined } // invalid
```

After the colon means the key must always be present, but the value can be undefined:
```js
// registerPreload: ?(id: number, element: Element) => void

{ } // invalid
{ registerPreload: (id, element) => {} } // valid
{ registerPreload: undefined } // valid
```

adamstep (approved):

PR #599, opened 2023-07-09
Port "UI Elements" in demo to Eleventy

adamstep:
Ported all screens under "UI Elements" to Nunjucks templates. This was done mostly automated with a script. Many files were also moved so that each page has its own directory.


hgray-instawork (commented):
Looks like we need to apply formatting to all the new njk files

hgray-instawork (approved):

PR #604, opened 2023-07-17
feature(navigation) - Implementation of navigation in Hyperview

hgray-instawork:
- Updated initial index to provide a/b functionality for legacy vs. updated navigation
  - Selection is based on presence or absence of a navigation system wrapping the screen (presence will use legacy)
- Support for nesting of navigators of type 'stack' and 'tab'
- Support for advanced navigation using an expanded version of Hyperview's navigation actions


demo/src/HyperviewScreen.js
```diff
@@ -6,62 +6,59 @@
  *
  */
 
+import * as Constants from './constants';
 import React, { PureComponent } from 'react';
 import HandleBack from './HandleBack';
-import Hyperview from 'hyperview';
+import Hyperview from 'hyperview/src';
```
flochtililoch, line 12: is this change needed? look at apply only needed changes (without Prettier / renaming)

demo/App.js
```diff
@@ -7,7 +7,6 @@
  */
 
 import './src/gesture-handler';
-import React from 'react';
```
flochtililoch, line 10: is this change needed?

flochtililoch (approved):
Nice work and achievement! Looking forward to what this evolution enables for us and other Hyperview devs

PR #609, opened 2023-07-21
chore(dependencies) - Reduce dependency on react-navigation

hgray-instawork:
In an effort to reduce dependency on react-navigation within Hyperview, these types were easy to replace with generic types.
- Replaced react-navigation types with local types which represent the minimal required data and functionality
- Removed package dependency on react-navigation/core

Tested in business app using new navigation system


flochtililoch (approved):
LGTM

PR #613, opened 2023-07-28
chore(cleanup) - avoid xml lookups

hgray-instawork:
Storing root, navigator, and screen into state to avoid having to look them up in XML every render


hgray-instawork:
Closing and reverting these changes. We don't want to destructure the document into state. We will look into creating a more performant document search method.

PR #614, opened 2023-07-28
fix(navigation) - ensure correct element index

hgray-instawork:
The original implementation assumed the first route was the only one with a nested element:
\<navigator\>
   \<nav-route ...\>
      \<navigator ...\/>
   \</nav-route\>
\</navigator\>

The index was assumed based on the document being read and built only once (index=0 on first run).

These changes allow for the correct element to be injected into the correct route regardless of its position or of the current index of the navigator state. They are necessary to properly handle deep link state updates.

Asana: https://app.asana.com/0/0/1205157207660925/f


hgray-instawork:
Closing to rethink solution

PR #615, opened 2023-07-28
fix(navigator) - improved handling of nested navigators

hgray-instawork:
The original implementation for handling `<navigator>` elements nested inside `<nav-route>` elements relied on storing Dom elements in a map by the id of the route which contained them.

This had several problems:
- A cache had to be created
- The items in the cache could become orphaned from their parent doc
  - Mutations to the doc would not be reflected in the cache
- The cache was never cleared
- The id or index of the implementation could change

This new solution uses a context to wrap every `hv-route` which contains its own document. This allows child routes to retrieve their Dom element and retrieve any nested navigator direction from the doc.

Improvements are also made to the initial route by creating all routes into the stack navigator.

### DIFF
--- a/src/contexts/route-doc.tsx
+++ b/src/contexts/route-doc.tsx
@@ -1,0 +1,18 @@
+/**
+ * Copyright (c) Garuda Labs, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import * as TypesLegacy from 'hyperview/src/types-legacy';
+import { createContext } from 'react';
+
+export type RouteDocContextProps = {
+  doc?: TypesLegacy.Document;
+};
+
+export const Context = createContext<TypesLegacy.Document | undefined>(
+  undefined,
+);
--- a/src/core/components/hv-route/index.tsx
+++ b/src/core/components/hv-route/index.tsx
@@ -15,6 +15,7 @@ import * as NavigationContext from 'hyperview/src/contexts/navigation';
 import * as NavigatorMapContext from 'hyperview/src/contexts/navigator-map';
 import * as NavigatorService from 'hyperview/src/services/navigator';
 import * as Render from 'hyperview/src/services/render';
+import * as RouteDocContext from 'hyperview/src/contexts/route-doc';
 import * as Stylesheets from 'hyperview/src/services/stylesheets';
 import * as Types from './types';
 import * as TypesLegacy from 'hyperview/src/types-legacy';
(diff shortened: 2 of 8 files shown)


src/core/components/hv-route/index.tsx
```diff
@@ -313,6 +323,36 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, Types.State> {
...
+
+  const routes = doc
+    .getElementsByTagNameNS(
+      Namespaces.HYPERVIEW,
+      TypesLegacy.LOCAL_NAME.NAV_ROUTE,
+    )
+    .filter((n: TypesLegacy.Element) => {
+      return n.getAttribute('id') === id;
+    });
+  const route = routes && routes.length > 0 ? routes[0] : undefined;
```
flochtililoch, line 345: I think this is the same with a little less iterations:
```suggestion
  const route = doc.getElementById(id)
```

src/core/components/hv-route/index.tsx
```diff
@@ -313,6 +323,36 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, Types.State> {
...
+  if (route) {
+    const navigators = route.getElementsByTagNameNS(
+      Namespaces.HYPERVIEW,
+      TypesLegacy.LOCAL_NAME.NAVIGATOR,
+    );
+    return navigators && navigators.length > 0 ? navigators[0] : undefined;
```
flochtililoch, line 351: I assume here, by extracting the first navigator, we want to retrieve the first immediate `<navigator>` descendant of `route`?
if so we might want to restrict the selection to `route.childNodes`?

hgray-instawork: Thanks. I had the same thought and had the fix done but didn't push before I stepped away. It's updated now.

src/core/components/hv-route/index.tsx
hgray-instawork: This is what I first tried but this created a difference between document and element which made a bunch of other TypeScript calls (like getFirstTag) no longer work. This was the quickest way to move forward.

flochtililoch (approved):

src/contexts/route-doc.tsx
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+import * as TypesLegacy from 'hyperview/src/types-legacy';
+import { createContext } from 'react';
+
+export type RouteDocContextProps = {
+  doc?: TypesLegacy.Document;
+};
+
+export const Context = createContext<TypesLegacy.Document | undefined>(
```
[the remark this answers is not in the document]
hgray-instawork: Confirmed nested contexts work as expected

PR #616, opened 2023-07-28
Updates `metro-react-native-babel-preset`, `chokidar`, and `react`

lenwhite:
Updates the `metro-react-native-babel-preset`, `chokidar`, and `react` packages, then regenerates the `yarn.lock` file.

This removes the need to build the problematic `fsevent` package. Works fine otherwise as far as I can tell. 

This changes min. node version to 18 though.


flochtililoch (commented):
The diff in the yarn.lock seems huge, I suspect it was deleted then recreated? Could we instead preserve the existing, and let `yarn` just apply the diff with new dependencies?

lenwhite:
Unfortunately, that didn't work -- it required recreating the file to get the dependency to not install.

There probably is a way to get around it by manually editing the `yarn.lock` file to find the problematic dependency chain, but that'll be a lot of effort. Let me know if you think we should attempt that instead.

flochtililoch (commented):
I wouldn't recommend merging this change, as the entire yarn.lock was regenerated, which obfuscate what dependencies have changed.

After poking around, I noticed (by running the command `yarn why fsevent`) that the old version of fsevent is currently brought by `react-native` and `@storybook/react-native`. While the version of RN that we use at instawork (0.67.4) seem to have remove the dependency, it is still present on the latest version of `@storybook/react-native` (6.5.4, albeit as optional, which the command `yarn --no-optional` should help with).

I think we should instead update the two dependencies listed above (and their peer dependencies that require updating). Though from my experience, despite the error we see when installing the packages, that error does not prevent other packages from being installed.

Another stop-gap solution would be to "manually" doctor the yarn.lock file and remove the references to fsevent v1.

PR #617, opened 2023-07-31
Migrate remaining JS files to TypeScript

lenwhite:


schema/core.xsd
```diff
@@ -1195,7 +1195,7 @@
       </xs:sequence>
       <xs:attribute name="id" use="required" type="hv:ID" />
       <xs:attribute name="href" use="required" type="xs:string" />
-      <xs:attribute name="initial" type="xs:boolean" />
+      <xs:attribute name="selected" type="xs:boolean" />
```
DanielSchaffer, line 1198: Intentional?

src/VisibilityDetectingView.tsx
```diff
@@ -1,35 +1,33 @@
...
-type Props = {|
-  children: any,
-  onInvisible: ?() => void,
-  onVisible: ?() => void,
-  style: ?ViewStyleProp,
-|};
+type Props = {
+  children: any;
+  onInvisible: () => void | null | undefined;
+  onVisible: () => void | null | undefined;
```
DanielSchaffer, line 11: Do these actually need to be able to return `null` or `undefined`? `on*` event handlers are pretty much always `void`

src/VisibilityDetectingView.tsx
```diff
@@ -41,37 +39,44 @@ export default class VisibilityDetectingView extends PureComponent<Props> {
...
+  };
+
+  onMeasure = (
+    x: number,
+    y: number,
+    width: number,
+    height: number,
+    pageX: number,
+    pageY: number,
+  ) => {
```
DanielSchaffer, line 51: return type?

src/behaviors/hv-alert/index.ts
```diff
@@ -88,6 +91,7 @@ export default {
     // On Android, alerts don't have a default button when unspecified, so we need to set one.
     if (!options.length && Platform.OS === 'android') {
       options.push({
+        // eslint-disable-next-line @typescript-eslint/no-empty-function
```
DanielSchaffer, line 94: We can define a `noop` function in a utility module somewhere, disable the lint rule onece, and reuse it instead of having to include the lint disable comment everywhere

```
# util.ts
// eslint-disable-next-line @typescript-eslint/no-empty-function
export const noop = () => {};

# ...hv-alert/index.ts
import { noop } from './util';

...

onPress: noop,
```

src/VisibilityDetectingView.tsx
lenwhite: No, but it makes no difference in practice. Returning `void` alone would work, but we're leaving these sort of slightly-overly-broad things for a later cleanup step.

src/VisibilityDetectingView.tsx
```diff
@@ -1,35 +1,33 @@
...
-  tickInterval: ?IntervalID;
+  tickInterval: number | null | undefined;
 
   unmounted = false;
 
-  view: ?ElementRef<typeof View>;
+  view: ElementRef<typeof View> | null | undefined;
 
-  onRef = (view: ?ElementRef<typeof View>) => {
+  onRef = (view?: ElementRef<typeof View> | null) => {
```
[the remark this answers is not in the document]
lenwhite: There's a practical difference here: you can call `onRef(null)` with this signature, which otherwise won't be possible. 

If there are any such calls made, we'd have to find and change them, again in a subsequent cleanup step.

src/VisibilityDetectingView.tsx
lenwhite: You don't have to specify a return type explicitly, and it's not convention that return types always needs to be specified. 

In this case the compiler can infer that it returns `void`.

src/behaviors/hv-alert/index.ts
lenwhite: The cleaner way would be to make `onPress` optional IMO instead of making it mandatory and requiring passing a noop.

src/VisibilityDetectingView.tsx
DanielSchaffer: Sure, but if it's part of a public API, we should err on the side of being explicit.

src/behaviors/hv-alert/index.ts
DanielSchaffer: Ah, I didn't realize that was an option. That would indeed be better!

src/components/hv-list/index.tsx
```diff
@@ -69,7 +66,7 @@ export default class HvList extends PureComponent<HvComponentProps, State> {
   };
 
   getItems = () => {
-    const isOwnedBySelf = item => {
+    const isOwnedBySelf = (item: undefined) => {
```
DanielSchaffer, line 69: This doesn't seem right - I'm not sure why the `item.parentNode` property access below wouldn't cause a type checking error

src/components/hv-section-list/index.tsx
```diff
@@ -27,7 +24,7 @@ import type { State } from './types';
 
 export default class HvSectionList extends PureComponent<
   HvComponentProps,
-  State,
+  State
```
DanielSchaffer, line 27: Aren't trailing commas allowed here too now?

src/components/hv-view/index.tsx
```diff
@@ -71,11 +69,12 @@ export default class HvView extends PureComponent<HvComponentProps> {
   };
 
   getCommonProps = (): CommonProps => {
-    const style = createStyleProp(
+    // TODO: fix type
```
DanielSchaffer, line 72: To be addressed in this PR or a followup?

src/components/hv-web-view/index.tsx
```diff
@@ -26,7 +24,13 @@ export default class HvWebView extends PureComponent<HvComponentProps> {
 
   props: HvComponentProps;
 
-  onMessage = (event: ?{ nativeEvent: { data: string } }) => {
+  onMessage = (
+    event?: {
+      nativeEvent: {
+        data: string;
+      };
+    } | null,
```
DanielSchaffer, line 32: Can we create a type for this instead of defining it in the function signature?

src/core/components/hv-navigator/index.tsx
```diff
@@ -174,20 +174,20 @@ export default class HvNavigator extends PureComponent<Types.Props> {
...
-    const initial:
+    const selected:
       | TypesLegacy.Element
-      | undefined = NavigatorService.getInitialNavRouteElement(props.element);
-    if (!initial) {
+      | undefined = NavigatorService.getSelectedNavRouteElement(props.element);
+    if (!selected) {
       throw new NavigatorService.HvNavigatorError(
-        `No initial route defined for '${id}'`,
+        `No selected route defined for '${id}'`,
```
DanielSchaffer, line 182: Is the `selected` <-> `initial` change intentional?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hyperview",
-  "version": "0.68.0",
+  "version": "0.68.1",
```
tecnoloco, line 3: should we consider a major version bump (1.0.0)? or at least a feature version bump here (0.69.0)?
Also isn't this automatically updated by npm tooling?

scripts/generate/stories.js
tecnoloco, line 1: will scripts be kept in js?

src/behaviors/hv-alert/index.ts
```diff
@@ -62,8 +64,8 @@ export default {
             // With multiple behaviors for the same trigger, we need to stagger
             // the updates a bit so that each update operates on the latest DOM.
             // Ideally, we could apply multiple DOM updates at a time.
-            later(i).then(
-              () =>
+            later(i).then(() => {
+              return (
```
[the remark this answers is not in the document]
tecnoloco: +1

src/components/hv-section-list/index.tsx
```diff
@@ -111,9 +108,15 @@ export default class HvSectionList extends PureComponent<
 
     addNodes(this.props.element);
 
-    let items = [];
+    let items: Array<any> = [];
```
tecnoloco, line 111: Couldn't find where this is being used

src/VisibilityDetectingView.tsx
lenwhite: That's true, but I don't think this particular class is part of Hyperview's public API.

schema/core.xsd
hgray-instawork: Yes, this was a change we made in the navigation code

src/core/components/hv-navigator/index.tsx
hgray-instawork: Yup!

package.json
hgray-instawork: It will be versioned when released. This change should be undone here.

scripts/generate/stories.js
hgray-instawork: I think it's fine for these to be kept in js

PR #619, opened 2023-07-31
feat: `onError` hook

flochtililoch:
Add `onError` optional prop to main component, and call it on caught errors.

https://app.asana.com/0/1204323441804469/1205174837597406/f


hgray-instawork (approved):

src/core/components/hv-route/index.tsx
hgray-instawork, line 1: I think you forgot to pass the onError to <HvScreen> in this block: https://github.com/Instawork/hyperview/blob/8a6b638a931ffabb1eaca052dab02231d8c0d8d8/src/core/components/hv-route/index.tsx#L224-L245

PR #623, opened 2023-08-02
feature(selection) - Track the current selection state of each nav-route

hgray-instawork:
Reflect the user's selections in the document

- Use the 'focus' listener to allow each hv-route to set its selected state within the document
- De-select any sibling routes to ensure the current is the only selected

Note: I did not use 'blur' to have each route deselect itself as that would cause routes in the background to lose their state.

In the video, as the user selects between the different tabs, the various \<nav-route\> children of `tabs-navigator` are shown to reflect current selection. When the `company` modal is popped up, the `company` \<nav-route\> shows selected while its sibling `tabs-route` becomes deselected. Note that the `tabs-navigator` children retain their correct selection state.

Asana: https://app.asana.com/0/0/1205197138955014/f

https://github.com/Instawork/hyperview/assets/127122858/0687558f-c500-419a-bd69-bbd0843ade72

### DIFF
--- a/src/core/components/hv-route/index.tsx
+++ b/src/core/components/hv-route/index.tsx
@@ -386,15 +386,7 @@ const getNestedNavigator = (
     return undefined;
   }
 
-  const routes = doc
-    .getElementsByTagNameNS(
-      Namespaces.HYPERVIEW,
-      TypesLegacy.LOCAL_NAME.NAV_ROUTE,
-    )
-    .filter((n: TypesLegacy.Element) => {
-      return n.getAttribute('id') === id;
-    });
-  const route = routes && routes.length > 0 ? routes[0] : undefined;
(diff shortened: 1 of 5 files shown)


src/core/components/hv-route/index.tsx
```diff
@@ -438,6 +430,17 @@ export default function HvRoute(props: Types.Props) {
     routeDocContext,
   );
 
+  React.useEffect(() => {
+    if (props.navigation) {
+      const unsubscribe = props.navigation.addListener('focus', () => {
+        NavigatorService.setSelected(routeDocContext, props.route?.params?.id);
+      });
+
+      return unsubscribe;
```
flochtililoch, line 439: I'm not too familiar with the useEffect API, but does it automatically remove the listener on unmount?

flochtililoch (approved):

PR #626, opened 2023-08-04
chore(modal) - Change iOS presentation to use fullscreen

hgray-instawork:
Using the `cardStyleInterpolator` option to force iOS to render modals at full screen while retaining their intended animation.

| Default | Updated iOS | Updated Android |
| ------ | ----- | ----- |
| default | iOS | android |

Asana: https://app.asana.com/0/0/1205197143810575/f


flochtililoch (commented):
I see the flags we're using seem to be iOS specific, how do these work in Android?

hgray-instawork:
The presets only affect the target platform. Android is not affected by this config. See updated videos above.

flochtililoch (approved):

PR #633, opened 2023-08-07
refactor: Pickers UX

flochtililoch:
This PR brings UX and code improvement to the `date-field` and `picker-field` components.

Integrates the following changes:
- #630 
- #634
- #635
- #636
- #638
- #647
- #648
- #652
- #654 

https://app.asana.com/0/1204008699308084/1205235979122187/f

|  | Before | After |
|--|---|---|
| iOS | ios-before |  after-ios |
| Android | android-before | android-after |

TODO:
- [x] fix android selection bug
- [x] animate overlay fade in / out
- [x] allow separate configuration for open vs close animation timings


examples/advanced_behaviors/set_value/index.xml.njk
```diff
@@ -75,6 +75,10 @@ tags: advanced_behaviors
         shadowOffsetY="-5"
         shadowOpacity="0.2"
         shadowRadius="5"/>
+      <style
+        id="PickerModal__overlay"
+        backgroundColor="#1f1f1fa0"
```
hgray-instawork, line 80: Hard to tell from the syntax. Are these values deleted so they can be provided further down stream? Are they required fields?

src/components/hv-picker-field/index.js
hgray-instawork, line 1: is what's remaining exclusively android? Would it be better to create an index.android file?

hgray-instawork (approved):

examples/advanced_behaviors/set_value/index.xml.njk
flochtililoch: there's not deleted, I think the GH syntax coloring makes it looks like it's removed but it's not.

src/components/hv-picker-field/index.js
flochtililoch: Linking [this previous comment](https://github.com/Instawork/hyperview/pull/636#discussion_r1289109514) for visibility.

src/core/components/modal/index.js
```diff
@@ -0,0 +1,152 @@
...
+      styleAttr: 'modal-overlay-style',
+    }),
+  );
+
+  const onLayout = (event: LayoutEvent) => {
+    setHeight(event.nativeEvent.layout.height);
+  };
+
+  const animationDuration: number =
+    parseInt(props.element.getAttribute('modal-animation-duration'), 10) || 250;
```
adamstep, line 66: this might be surprising behavior if someone sets the duration to `0` (they want no animation) and it defaults to 250.

flochtililoch: good catch, I'll fix this

adamstep (commented):
Just checking, did you test these changes with the `set-value` action to make sure that functionality still works with the different picker UIs?

flochtililoch:
> did you test these changes with the set-value action

thanks for flagging, I'll test this

src/core/components/modal/index.js
flochtililoch: addressed in #655

adamstep (approved):

PR #636, opened 2023-08-08
refactor(`hv-picker-field`): separate out platform specific code

flochtililoch:
Note: This PR is one in a series of refactors, see #633 for other related changes.

Since the picker implementation has a lot of variations based on the platform, use RN platform includes to better isolate the code.


src/components/hv-picker-field/index.js
hgray-instawork, line 1: Did you want to rename this index.android.js to explicitly define it as android?

hgray-instawork (approved):

src/components/hv-picker-field/index.js
flochtililoch: no, I explicitly kept this to be the "default" handler, which is also used by the web platform (while support isn't great, in particular around events, it still correctly renders as a browser select drop down).
It also appear that not having a default handler would cause the unit tests to fail, and despite trying to fiddle with jest configuration, I couldn't get the tests to continue passing.

PR #640, opened 2023-08-09
chore(refactor) - small improvements to navigator code organization

hgray-instawork:
A few code refactors which will help the next step:
1. Created a helper to determine if a route id is one of the generated 'dynamic/modal' routes
2. Centralized the `screenOptions` for stack navigators to reduce redundant code
3. Created a better `getId` function for getting the id of dynamic and defined routes
4. Moved the creation of dynamic routes into its own function
5. Re-use `buildScreen` when building dynamic routes to avoid redundant code

### DIFF
--- a/src/core/components/hv-navigator/index.tsx
+++ b/src/core/components/hv-navigator/index.tsx
@@ -204,15 +204,6 @@ export default class HvNavigator extends PureComponent<Types.Props> {
       throw new NavigatorService.HvNavigatorError('No id found for navigator');
     }
 
-    const navigationContext: NavigationContext.NavigationContextProps | null = useContext(
-      NavigationContext.Context,
-    );
-    const navigatorMapContext: NavigatorMapContext.NavigatorMapContextProps | null = useContext(
-      NavigatorMapContext.NavigatorMapContext,
-    );
-    if (!navigationContext || !navigatorMapContext) {
-      throw new NavigatorService.HvRouteError('No context found');
-    }
 
     const type:
       | TypesLegacy.DOMString
--- a/src/core/components/hv-navigator/index.tsx
+++ b/src/core/components/hv-navigator/index.tsx
@@ -27,7 +27,6 @@ export default class HvNavigator extends PureComponent<Types.Props> {
    * Encapsulated options for the stack screenOptions
    */
   stackScreenOptions = (
-    id: string,
     route: Types.ScreenParams,
   ): {
     headerMode: 'float' | 'screen' | undefined;
@@ -36,7 +35,7 @@ export default class HvNavigator extends PureComponent<Types.Props> {
   } => ({
     headerMode: 'screen',
     headerShown: SHOW_NAVIGATION_UI,
-    title: route.params?.url || id,
+    title: this.getId(route.params),
   });
 
   /**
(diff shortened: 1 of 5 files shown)


src/core/components/hv-navigator/index.tsx
```diff
@@ -23,6 +23,38 @@ const Stack = NavigatorService.createStackNavigator<Types.ParamTypes>();
...
+  /**
+   * Encapsulated options for the stack screenOptions
+   */
+  stackScreenOptions = (
+    route: Types.ScreenParams,
+  ): {
+    headerMode: 'float' | 'screen' | undefined;
+    headerShown: boolean;
+    title: string | undefined;
+  } => ({
```
flochtililoch, line 35: could we define this as a type alias in `Types`, to make the method signature more readable?

src/core/components/hv-navigator/index.tsx
```diff
@@ -64,6 +99,33 @@ export default class HvNavigator extends PureComponent<Types.Props> {
     );
   };
 
+  /**
+   * Build the dynamic and modal screens for a stack navigator
```
flochtililoch, line 103: (maybe not in context with this PR): I understand that `dynamic` represents the regular stack, while `modal` represents a stack specifically configured for modals, but I find the naming a bit confusing.
Our previous naming convention on our app was `home` vs `modals`, which honestly isn't that much better, but I wonder if we should streamline this. I think `dynamic` is a misnomer since the modal stack is also dynamic, so maybe we can use `regular` and `modal`? Up for other suggestions

src/core/components/hv-navigator/index.tsx
```diff
@@ -199,10 +224,7 @@ export default class HvNavigator extends PureComponent<Types.Props> {
           <Stack.Navigator
             id={id}
             screenOptions={({ route }: Types.NavigatorParams) => ({
-              header: undefined,
-              headerMode: 'screen',
-              headerShown: SHOW_NAVIGATION_UI,
-              title: route.params?.url || id,
+              ...this.stackScreenOptions(route),
             })}
```
flochtililoch, line 228: minor: remove an iteration
```suggestion
            screenOptions={({ route }: Types.NavigatorParams) =>
              this.stackScreenOptions(route),
            }
```

src/services/navigator/helpers.ts
```diff
@@ -14,6 +14,13 @@ import * as TypesLegacy from 'hyperview/src/types-legacy';
 import * as UrlService from 'hyperview/src/services/url';
 import { ANCHOR_ID_SEPARATOR } from './types';
 
+/**
+ * Dynamic and modal routes are not defined in the document
+ */
+export const isDynamicId = (id: string): boolean => {
```
flochtililoch, line 20: Does `dynamic` in the method name represents the same as the current navigator type "Dynamic"? I assume not given the code below, but naming also is confusing here.

src/core/components/hv-navigator/index.tsx
hgray-instawork: Agreed naming isn't ideal. I'll put some thought into it

src/services/navigator/helpers.ts
hgray-instawork: Agreed. In this case, I mean routes which are not pre-defined. Collectively, they are dynamically generated by the component. A better name for this and for the actual two screens would be ideal.

flochtililoch (approved):

PR #642, opened 2023-08-09
chore(dynamic) - refactor 'dynamic' to 'card'

hgray-instawork:
Replaced all uses of 'dynamic' as a type of route to 'card' to better reflect the implementation details.


flochtililoch (approved):
LGTM

PR #643, opened 2023-08-09
chore(destructuring) - remove prop and component destructuring

hgray-instawork:
Remove instances of
`const { someFunction } = this`
and
`const { someProp } = props`

### DIFF
--- a/src/core/components/hv-route/index.tsx
+++ b/src/core/components/hv-route/index.tsx
@@ -277,7 +277,7 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, Types.State> {
    */
   Route = (): React.ReactElement => {
     const isModal = this.props.route?.params.isModal
-      ? this.props.route.params
+      ? this.props.route.params.isModal
       : false;
 
     const renderElement: TypesLegacy.Element | undefined = isModal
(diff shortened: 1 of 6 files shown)


hgray-instawork:
Having an issue where nested components don't allow accessing each other via `this.SomeComponent`. For example, this fails:
`return <this.Route />`
but this passes:
`const { Route } = this;
return <Route />`

The error is 'Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.'

src/core/components/hv-route/index.tsx
```diff
@@ -276,7 +276,9 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, Types.State> {
    * Evaluate the <doc> element and render the appropriate component
    */
   Route = (): React.ReactElement => {
-    const { isModal } = this.props.route?.params || { isModal: false };
+    const isModal = this.props.route?.params.isModal
+      ? this.props.route.params
```
flochtililoch, line 280: I _think_ this is incorrect
```suggestion
    const isModal = this.props.route?.params.isModal ?? false
```

hgray-instawork: Yes, good catch. Fixed.

flochtililoch (approved):

PR #645, opened 2023-08-10
chore(OKR): Lint configuration synced with master

mkhizeryounas:
### Changelog
- Removed following ignored rules by updating the code
  - `max-len`
  - `implicit-arrow-linebreak`
  - `@typescript-eslint/no-explicit-any`
- Removed unused rules patterns from `import/no-extraneous-dependencies` as those are overridden by `TS` configuration
- Added `import/no-extraneous-dependencies` rule for `stories` in overrides


.eslintrc
```diff
@@ -72,11 +64,15 @@
         "no-use-before-define": "off",
         "object-curly-newline": "off",
         "instawork/error-object": "off",
-        "@typescript-eslint/no-explicit-any": "off",
-        "implicit-arrow-linebreak": "off",
```
tecnoloco, line 76: > Removed following ignored rules by updating the code
> max-len
> implicit-arrow-linebreak
> @typescript-eslint/no-explicit-any

Does this mean all arrow line breaks errors are resolved?

mkhizeryounas: Yes

.eslintrc
```diff
@@ -72,11 +64,15 @@
...
-        "max-len": "off",
-        "import/no-extraneous-dependencies": "off" // TODO: fix?
+        "import/no-extraneous-dependencies": [
+          "warn",
+          {
+            "devDependencies": [
+              "**/stories/*.tsx"
```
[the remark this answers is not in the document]
mkhizeryounas: These rules are overridden by TS config so added this rule just to make sure the conventions followed in the project is satisfied.

tecnoloco (approved):
No further comments from my side. LGTM. I would recommend another pair of eyes before merging though.

src/services/navigator/helpers.test.ts
```diff
@@ -2,7 +2,7 @@ import * as DomErrors from '../dom/errors';
 import * as Errors from './errors';
 import * as Namespaces from '../namespaces';
 import * as Types from './types';
-import * as TypesLegacy from '../../types-legacy';
+import * as TypesLegacy from '../../types';
```
flochtililoch, line 5: is this change correct? It doesn't look like this PR renames the file.

test/helpers.tsx
```diff
@@ -27,7 +27,9 @@ export const getElements = (
 };
 
 export const getDummyHvProps = () => ({
-  onUpdate: () => {},
+  onUpdate: () => {
+    /** PASS */
```
flochtililoch, line 31: curious why this is needed

.eslintrc
```diff
@@ -54,11 +50,7 @@
     },
     {
       "files": ["**/*.ts", "**/*.tsx"],
-      "extends": [
-        "plugin:@typescript-eslint/eslint-recommended",
-        "plugin:@typescript-eslint/recommended",
-        "plugin:instawork/recommended-typescript"
-      ],
+      "extends": ["plugin:instawork/recommended-typescript"],
```
flochtililoch, line 53: any reasons to remove the two other extended plugins?

.flowconfig
```diff
@@ -14,6 +14,7 @@
 .*/node_modules/pug-filters/.*
 .*/node_modules/eslint-import-resolver-node/.*
 .*/node_modules/eslint-plugin-import/.*
+.*/node_modules/@react-native-picker/.*
```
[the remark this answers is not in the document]
mkhizeryounas: yes, otherwise flow is failing.

src/services/dom/types.ts
```diff
@@ -33,7 +33,8 @@ export const X_RESPONSE_STALE_REASON = {
   STALE_IF_ERROR: 'stale-if-error',
 } as const;
 
-export type XResponseStaleReason = typeof X_RESPONSE_STALE_REASON[keyof typeof X_RESPONSE_STALE_REASON];
+type Reason = typeof X_RESPONSE_STALE_REASON[keyof typeof X_RESPONSE_STALE_REASON];
+export type XResponseStaleReason = Reason;
```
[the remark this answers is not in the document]
mkhizeryounas: this is because we have a limit of 100 chars but this becomes bigger than 100 chars so to fix that changed this here.

src/services/navigator/helpers.test.ts
mkhizeryounas: yes, due to this tests were failing as this file doesn't exist

test/helpers.tsx
mkhizeryounas: lint fails, complaining that unexpected arrow function.

.eslintrc
mkhizeryounas: Not really, these didn't exist in master so just to sync both.

src/services/inline-context/index.ts
```diff
@@ -50,11 +50,12 @@ import { preorder } from 'hyperview/src/services/dom/helpers';
                |                            |
                |                            |
                |                            |
-         [⏎◦◦◦◦◦◦Hello⏎]              [⏎◦◦◦◦◦◦World⏎◦◦◦◦]
```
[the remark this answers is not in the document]
mkhizeryounas: Max len rule fails if we don't do this.

flochtililoch (approved):

PR #646, opened 2023-08-11
fix: stop usage of deprecated api

### LINKED ISSUE #461: Demo app: Clipboard behavior using deprecated API
Still functional, but we should fix the warning by installing [the community plugin](https://docs.expo.dev/versions/latest/sdk/clipboard/).

Simulator Screen Shot - iPhone 13 mini - 2022-07-17 at 15 53 30
Simulator Screen Shot - iPhone 13 mini - 2022-07-17 at 15 53 33

bronislav:
The `Clipboard` class has been extracted from the react-native core, and the proper way to use it is through a package `@react-native-clipboard/clipboard`.

- Add the required package to the peer and dev dependencies.
- Import the Clipboard object from the correct package.

Functional and no warning:

https://github.com/Instawork/hyperview/assets/125880/c8963eb4-a219-4c2e-ae39-a2692b21bfe0

Closes #461

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -11,6 +11,7 @@
   "dependencies": {
     "@babel/preset-env": "^7.22.9",
     "@react-native-community/datetimepicker": "7.2.0",
+    "@react-native-clipboard/clipboard": "^1.11.2",
     "@react-native-picker/picker": "2.4.10",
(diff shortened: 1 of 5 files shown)


flochtililoch (approved):
Thanks for the fix! Can I also put you in charge of installing that new dependency in our mobile repo?

PR #649, opened 2023-08-11
refactor: de-dup behavior trigger logic

flochtililoch:
Remove duplicated/identical implementations of logic triggering certain behaviors on elements.

Note: hv-alert, hv-list and `hyper-ref` still have their own custom implementation of selecting/executing behaviors.


hgray-instawork (approved):
Nice cleanup!

PR #658, opened 2023-08-18
refactor(`HyperRef`): Remove unnecessary render cycles

flochtililoch:
With previous code, we would first render the Hyperview element and its children (example: [HvText](https://github.com/Instawork/hyperview/blob/25d1633ca92fd2d87a8c504a1b61245a50ab6a2f/src/components/hv-text/index.js#L36-L49)) before [delegating to addHref](https://github.com/Instawork/hyperview/blob/25d1633ca92fd2d87a8c504a1b61245a50ab6a2f/src/components/hv-text/index.js#L53), which would then check for presence of behavior, and when some found, would [render element's children again](https://github.com/Instawork/hyperview/blob/25d1633ca92fd2d87a8c504a1b61245a50ab6a2f/src/core/hyper-ref/index.js#L451), before rendering `HyperRef`, which internally would [render the element again](https://github.com/Instawork/hyperview/blob/25d1633ca92fd2d87a8c504a1b61245a50ab6a2f/src/core/hyper-ref/index.js#L411-L416).

This change skips the initial rendering of element and children and delegate immediately to HyperRef.

### DIFF
--- a/src/components/hv-image/index.js
+++ b/src/components/hv-image/index.js
@@ -11,9 +11,9 @@
 import * as Namespaces from 'hyperview/src/services/namespaces';
 import React, { PureComponent } from 'react';
 import type { HvComponentProps } from 'hyperview/src/types';
+import HyperRef from 'hyperview/src/core/hyper-ref';
 import { Image } from 'react-native';
 import { LOCAL_NAME } from 'hyperview/src/types';
-import { addHref } from 'hyperview/src/core/hyper-ref';
 import { createProps } from 'hyperview/src/services';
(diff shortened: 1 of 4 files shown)


hgray-instawork (approved):

adamstep (approved):
Do you expect any change from code that expected the extra render cycles?

flochtililoch:
> Do you expect any change from code that expected the extra render cycles?

From my understanding, the extra render cycles were never inserted in the DOM, so therefore never used. I don't expect any difference in behaviors with previous code, but we'll need to do a rigorous regression testing.

PR #659, opened 2023-08-21
fix: `visible` behavior missing detections

flochtililoch:
When toggling views with `show`/`hide` behaviors, if the layout is identical enough (i.e. same elements structure), then React does not unmount the instances of `VisibilityDetectingView`, which in turn cause the previous "visible" state to not be reset. The code change here works around the issue, by adding an `id` prop to `VisibilityDetectingView`, which is then used to decide internally to reset the "visible" state. The `id` prop is either set from an existing `id` attribute on the element targeted by the `visible` behavior, or generated, using every behavior attributes available on the element (which means if all attributes are identical, the observed issue will remain - however unlikely to happen).

The fix also addresses another bug that was uncovered after the first resolution: we were relying on an older version of the DOM when triggering the "visible" callbacks. Instead, query the DOM in the callback.

Note: the use case "Tabs" in example has also been reworked to help diagnose the issue.

| Before | After |
|-------|------|
| before | after |

### DIFF
--- a/src/core/hyper-ref/index.js
+++ b/src/core/hyper-ref/index.js
@@ -389,16 +389,33 @@ export default class HyperRef extends PureComponent<Props, State> {
     if (!behaviors.length) {
       return children;
     }
-    const visibleHandlers = behaviors.map(behaviorElement =>
-      this.createActionHandler(behaviorElement, this.props.onUpdate),
-    );
     const onVisible = () => {
-      visibleHandlers.forEach(h => h(this.props.element));
+      // We don't want to use the cached `behaviors` list here because
+      // the DOM might have been mutated since.
+      this.getBehaviorElements(TRIGGERS.VISIBLE)
+        .map(behaviorElement =>
+          this.createActionHandler(behaviorElement, this.props.onUpdate),
+        )
+        .forEach(h => h(this.props.element));
     };
+
+    // If element does not have an `id` attribute, generate a pseudo id from list of attributes
+    // This is necessary to indicate to the VisibilityDetectingView that the element has changed
+    // and the internal state needs to be reset.
+    const id =
+      this.props.element.getAttribute('id') ||
+      Object.values(ATTRIBUTES)
+        // $FlowFixMe: every value in ATTRIBUTES are strings
+        .reduce((acc: string[], name: string) => {
+          const value = this.props.element.getAttribute(name);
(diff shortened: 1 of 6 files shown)


src/core/hyper-ref/types.js
```diff
@@ -32,10 +32,12 @@ export type State = {|
 export const ATTRIBUTES = {
   ACTION: 'action',
   DELAY: 'delay',
+  EVENT_NAME: 'event-name',
```
tecnoloco, line 35: Is this intended? I couldn't figure it out why this attribute is added now.

src/core/hyper-ref/types.js
```diff
@@ -32,10 +32,12 @@ export type State = {|
...
   HIDE_DURING_LOAD: 'hide-during-load',
   HREF: 'href',
   HREF_STYLE: 'href-style',
   IMMEDIATE: 'immediate',
+  NEW_VALUE: 'new-value',
```
tecnoloco, line 40: Same here. Is this intended? I couldn't figure it out why this attribute is added now.

tecnoloco (approved):
No major comments from my side. Overall LGTM. I would recommend another pair of eyes here.

src/core/hyper-ref/types.js
flochtililoch: yeah, I added these attributes as they are valid behavior attributes, and this list is used when generating an `id` for `VisibilityDetectingView` - I figured I'd add these to increase chances of generating a unique id

hgray-instawork (approved):

src/core/hyper-ref/index.js
```diff
@@ -389,16 +389,33 @@ export default class HyperRef extends PureComponent<Props, State> {
...
+        .map(behaviorElement =>
+          this.createActionHandler(behaviorElement, this.props.onUpdate),
+        )
+        .forEach(h => h(this.props.element));
     };
 
+    // If element does not have an `id` attribute, generate a pseudo id from list of attributes
+    // This is necessary to indicate to the VisibilityDetectingView that the element has changed
+    // and the internal state needs to be reset.
+    const id =
```
adamstep, line 405: I guess an alternative would be to take a hash of the stringified content of the element. I think this should work for now, but we should look out for any potential issues down the line.

PR #663, opened 2023-08-22
Revert "refactor(`HyperRef`): Remove unnecessary render cycles (#658)"

flochtililoch:
This reverts commit dae263f4b7926ffe477024a0892ed374a8b93860, introduced in #658.

The change appears to cause regression in our mobile app. The change was a nice to have, but not a requirement for #659.


adamstep (approved):
Interesting, would be good to understand why there was a regression.

PR #664, opened 2023-08-24
feat(`HvTextField`): debounced `change` trigger

flochtililoch:
See commits breakdown:

- Updates 11ty setup to allow custom JS handler
- Create new example under case study, which renders a large list of paginated contacts that can be searched
- refactor `HvTextField` into a functional component
- implement debounce mechanism

Note: last commit is only satisfying CI, but prevents the case study to load correctly, issue seem related to 11ty and needs to be investigated.

| Before | After |
|--------|------|
| before | after |


src/components/hv-text-field/index.js
```diff
@@ -34,6 +34,8 @@ const HvTextField = (props: HvComponentProps) => {
 
   // Extract known attributes into their own variables
   const autoFocus = props.element.getAttribute('auto-focus') === 'true';
+  const debounceTime =
```
adamstep, line 37: Maybe call this `debounceMs` to clarify the units?

src/components/hv-text-field/index.js
```diff
@@ -57,12 +59,21 @@ const HvTextField = (props: HvComponentProps) => {
     }
   };
 
+  // Create a memoized, debounced function to trigger the "change" behavior
+  const triggerChangeBehaviors = useCallback(
+    debounce((newElement: Element) => {
```
adamstep, line 64: Where is `debounce()` defined?

src/components/hv-text-field/index.js
flochtililoch: done in [74ea644](https://github.com/Instawork/hyperview/pull/664/commits/74ea644f1fc3846d198ce23419971f3e6ceec60e)

src/components/hv-text-field/index.js
flochtililoch: `debounce` is imported from lodash [here](https://github.com/Instawork/hyperview/pull/664/files#diff-72175845930d650efd9d7e8b95f74c4eed08cc08a2a78c67f57b4ed81c2853e2R22)

adamstep (approved):
nice work!

PR #669, opened 2023-08-29
feat(lists): `scroll` behavior action

flochtililoch:
Add new behavior action `scroll`, that scrolls a `<list>` or `<section-list>` to desired `target`. Implementing the same support as part of scrollable views is a larger effort, so we'll postpone this until needed.

See included list and section list examples for usage.

TODO:
- [X] Add docs

| iOS | Android |
|-----|--------|
| iOS | android |

### DIFF
--- a/schema/scroll.xsd
+++ b/schema/scroll.xsd
@@ -1,0 +1,35 @@
+<xs:schema
+  xmlns:xs="http://www.w3.org/2001/XMLSchema"
+  attributeFormDefault="qualified"
+  elementFormDefault="qualified"
+  targetNamespace="https://hyperview.org/hyperview-scroll"
+  xmlns:hv="https://hyperview.org/hyperview"
+>
+  <xs:import
+    namespace="https://hyperview.org/hyperview"
+    schemaLocation="hyperview.xsd"
+  />
+
+  <xs:element name="option">
+    <xs:complexType>
+      <xs:sequence minOccurs="0" maxOccurs="unbounded">
+          <xs:element ref="hv:behavior" />
+      </xs:sequence>
+      <xs:attribute name="label" type="xs:string" use="required" />
+      <xs:attributeGroup ref="hv:behaviorAttributes" />
+    </xs:complexType>
+  </xs:element>
+
+  <xs:attributeGroup name="scrollAttributes">
+    <xs:attribute name="animated" type="xs:boolean" />
+    <xs:attribute name="offset" type="xs:integer" />
+    <xs:attribute name="position">
+      <xs:simpleType>
+        <xs:restriction base="xs:float">
+          <xs:minInclusive value="0" />
+          <xs:maxInclusive value="1" />
+        </xs:restriction>
+      </xs:simpleType>
+    </xs:attribute>
+  </xs:attributeGroup>
+</xs:schema>
--- a/schema/scroll.xsd
+++ b/schema/scroll.xsd
@@ -9,17 +9,6 @@
     namespace="https://hyperview.org/hyperview"
     schemaLocation="hyperview.xsd"
   />
-
-  <xs:element name="option">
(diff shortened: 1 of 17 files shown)


adamstep:
The two behaviors that use custom namespaces are `alert` and `share`. The thinking was that these are "system behaviors", meaning they interact with OS features vs being self-contained to the Hyperview world. That's also why they are treated a bit differently in the reference docs: https://hyperview.org/docs/reference_index

I think `scroll` fits in with the core behaviors (contained to the Hyperview world, uses `target`) that it doesn't need a separate namespace.

schema/scroll.xsd
```diff
@@ -0,0 +1,35 @@
...
+
+  <xs:element name="option">
+    <xs:complexType>
+      <xs:sequence minOccurs="0" maxOccurs="unbounded">
+          <xs:element ref="hv:behavior" />
+      </xs:sequence>
+      <xs:attribute name="label" type="xs:string" use="required" />
+      <xs:attributeGroup ref="hv:behaviorAttributes" />
+    </xs:complexType>
+  </xs:element>
```
adamstep, line 21: I think this is leftover from copying?

flochtililoch: yes good catch

flochtililoch: [cf7106d](https://github.com/Instawork/hyperview/pull/669/commits/cf7106d742c4b1f4f2682a3f9bc53fb23e0a1639)

src/components/hv-list/index.js
```diff
@@ -37,10 +46,100 @@ export default class HvList extends PureComponent<HvComponentProps, State> {
...
+    const targetElement: ?Element = doc?.getElementById(targetId);
+    if (!targetElement) {
+      return;
+    }
+
+    const targetElementParentList = getAncestorByTagName(
+      targetElement,
+      LOCAL_NAME.LIST,
+    );
+    if (targetElementParentList === this.props.element) {
```
adamstep, line 88: maybe return early in this case to remove a layer of nesting?

adamstep (commented):
If I understand correctly, the "scroll" action can only be triggered from within a `<list>`, `<section-list>`, etc. Is there any use case for triggering a scroll from outside those elements? For example, if we had quick links to jump to certain sections of the profile screen (this pattern is common in apps with menus, like DoorDash).

flochtililoch:
> Is there any use case for triggering a scroll from outside those elements?

I consider list and section list elements that take over the entire screen, aside from the navigation bars, so in theory it should be rare to need a behavior to live outside of it. Having the behavior nested under the list allow for the behavior to know which list to deal with, and it also is necessary for the implementation to have access to the component's reference that accesses the scroll API.
If we wanted to trigger the behavior from outside of the list, I think we could use our Hyperview event mechanism, so in theory we shouldn't be limited. I can build an example to validate this use case.

adamstep:
> > Is there any use case for triggering a scroll from outside those elements?
> 
> I consider list and section list, elements that take over the entire screen, aside from the navigation bars, so in theory it should be rare to need a behavior to live outside of it. Having the behavior nested under the list allow for the behavior to know which list to deal with, and it also is necessary for the implementation to have access to the component's reference that accesses the scroll API. If we wanted to trigger the behavior from outside of the list, I think we could use our Hyperview event mechanism, so in theory we shouldn't be limited. I can build an example to validate this use case.

Sounds reasonable. I guess we now also have support for sticky elements, which should make it possible to build what I was referring to.

src/components/hv-list/index.js
flochtililoch: [c5f912b](https://github.com/Instawork/hyperview/pull/669/commits/c5f912be897d590557aa44d2aa590d98ba7915a1)

adamstep (approved):
nice work!

PR #673, opened 2023-09-01
refactor: `RefreshControl` context

flochtililoch:
`RefreshControl` context allows for customizing the refresh control in `<list>` and `<section-list>`. Consumers were setup in a React class component fashion, using the static `contextType` property. This approach is limiting, as only one context can be set on a given class component, but is necessary when the context needs to be accessed outside the render method. Since this is not the case for either of these classes, refactor to use the JSX consumer style in the render methods, which now leaves room for introducing another context (future PR coming soon) that does need to be accessed outside the render method. Additionally, this re-adds the missing `refreshControl` property being passed down from `hv-root` to `hv-screen`.

| `<list>` | `<section-list>` |
|-------|-----------|
| list | sectionlist |

Captures above tested in the demo app with the following local changes:

<details>

```diff
diff --git a/demo/src/HyperviewScreen.js b/demo/src/HyperviewScreen.js
index e91c8f9..7669382 100644
--- a/demo/src/HyperviewScreen.js
+++ b/demo/src/HyperviewScreen.js
@@ -9,6 +9,7 @@
 import React, { PureComponent } from 'react';
 import HandleBack from './HandleBack';
 import Hyperview from 'hyperview';
+import { RefreshControl } from 'react-native';
 import moment from 'moment';
 import { MAIN_STACK_NAME, MODAL_STACK_NAME } from './constants';

@@ -63,6 +64,10 @@ export default class HyperviewScreen extends PureComponent {
     });
   }

+  Refresh = () => (
+    <RefreshControl tintColor="red" />
+  )
+
   render() {
     const entrypointUrl = this.props.route.params?.url;

@@ -79,6 +84,7 @@ export default class HyperviewScreen extends PureComponent {
           openModal={this.openModal}
           push={this.push}
           route={this.props.route}
+          refreshControl={this.Refresh}
         />
       </HandleBack>
     );
```
</details>

### DIFF
--- a/src/core/components/hv-root/index.tsx
+++ b/src/core/components/hv-root/index.tsx
@@ -53,7 +53,6 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
             onParseBefore={this.props.onParseBefore}
             openModal={this.props.openModal}
             push={this.props.push}
-            refreshControl={this.props.refreshControl}
             route={this.props.route}
           />
         </Contexts.RefreshControlComponentContext.Provider>
--- a/src/core/components/hv-root/index.tsx
+++ b/src/core/components/hv-root/index.tsx
@@ -53,6 +53,7 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
             onParseBefore={this.props.onParseBefore}
             openModal={this.props.openModal}
             push={this.props.push}
+            refreshControl={this.props.refreshControl}
             route={this.props.route}
           />
         </Contexts.RefreshControlComponentContext.Provider>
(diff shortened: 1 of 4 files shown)


src/core/components/hv-root/index.tsx
```diff
@@ -53,6 +53,7 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
             onParseBefore={this.props.onParseBefore}
             openModal={this.props.openModal}
             push={this.props.push}
+            refreshControl={this.props.refreshControl}
```
adamstep, line 56: I don't understand this line given the context provider wrapping the component.

flochtililoch: it was needed because the context was re-injected in the `hv-screen` component, using that prop. I removed the duplication, and will address the duplication for other contexts separately.
[cd658f4](https://github.com/Instawork/hyperview/pull/673/commits/cd658f4dc3c968c72925a6afdbbe0793cdde78bb)

adamstep (approved):

PR #674, opened 2023-09-02
Updated Navidator.js to use SafeAreaProvider and SafeAreaView

gorak:
Resolves issue with Android header (Issue 550)


adamstep:
Thanks @gorak . @hgray-instawork can you take a look?

hgray-instawork (approved):
I think we should do a few quick tests with the default navigation UI to make sure this doesn't cause problems.

PR #675, opened 2023-09-06
fix(`hv-text-field`): `focused` state

flochtililoch:
When the component backing the `<text-field>` element re-rendered with a new `element` prop, the state previously set locally is lost, causing the styles to no longer apply the `focused` modifier. This fix removes the local DOM state for `focused`, and instead reads the state from the actual RN component.

| Before | After |
|-------|-------|
| before | after |

### DIFF
--- a/src/components/hv-text-field/index.js
+++ b/src/components/hv-text-field/index.js
@@ -11,11 +11,12 @@
 import * as Behaviors from 'hyperview/src/services/behaviors';
 import * as Namespaces from 'hyperview/src/services/namespaces';
 import type { Element, HvComponentProps } from 'hyperview/src/types';
-import React, { useCallback } from 'react';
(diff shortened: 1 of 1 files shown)


src/components/hv-text-field/index.js
```diff
@@ -76,17 +76,24 @@ const HvTextField = (props: HvComponentProps) => {
...
   };
 
   return (
     <TextInput
       {...p}
-      ref={props.options.registerInputHandler}
+      ref={(ref: ?ElementRef<typeof TextInput>) => {
+        textInputRef.current = ref;
+        if (props.options?.registerInputHandler) {
+          props.options?.registerInputHandler(ref);
```
palaparthi, line 94:
```suggestion
          props.options.registerInputHandler(ref);
```

palaparthi (approved):

PR #678, opened 2023-09-06
feat(scrollable-views): `dismiss-keyboard-on-drag`

flochtililoch:
Add attribute to `view`, `list` and `sectionlist` elements, which condition whether the keyboard should be dismissed upon user interacting with the scroll view.

https://github.com/Instawork/hyperview/assets/309515/8c4dbcc0-064a-473c-a1d9-e37d79b19b3f

### DIFF
--- a/schema/core.xsd
+++ b/schema/core.xsd
@@ -419,7 +419,15 @@
     </xs:attribute>
     <xs:attribute name="shows-scroll-indicator" type="xs:boolean" />
     <xs:attribute name="content-container-style" type="hv:styleList" />
-    <xs:attribute name="dismiss-keyboard-on-drag" type="xs:boolean" />
+    <xs:attribute name="keyboard-dismiss-mode">
+      <xs:simpleType>
+        <xs:restriction base="xs:string">
+          <xs:enumeration value="none" />
+          <xs:enumeration value="on-drag" />
+          <xs:enumeration value="interactive" />
+        </xs:restriction>
+      </xs:simpleType>
+    </xs:attribute>
   </xs:attributeGroup>
 
   <xs:element name="doc">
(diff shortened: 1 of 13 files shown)


schema/core.xsd
```diff
@@ -419,6 +419,7 @@
     </xs:attribute>
     <xs:attribute name="shows-scroll-indicator" type="xs:boolean" />
     <xs:attribute name="content-container-style" type="hv:styleList" />
+    <xs:attribute name="dismiss-keyboard-on-drag" type="xs:boolean" />
```
palaparthi, line 422: Looks like rn supports another mode `interactive` for ios. What do you think about keeping the the prop same same(`keyboard-dismiss-mode`) and making it an enum instead?

flochtililoch: I originally had an enum instead of boolean, but couldn't get the interactive mode to work in Expo, which led me to refactor with a boolean instead

src/services/keyboard/index.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+import * as TypesLegacy from 'hyperview/src/types-legacy';
+
+export const getKeyboardDismissMode = (
+  element: TypesLegacy.Element,
+): 'on-drag' | undefined => {
+  return element.getAttribute('dismiss-keyboard-on-drag') === 'true'
+    ? 'on-drag'
+    : undefined;
```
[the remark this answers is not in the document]
flochtililoch: The prop `keyboardDismissMode` accepts `undefined` as a value, which feels more correct when dev has not explicitly defined the value. While unlikely to change, RN could in the future tweak the behavior of this prop by platform, and explicitly setting it here would prevent the platform default to apply.

schema/core.xsd
palaparthi: ok, imo I think we can still have it as an enum with only 2 options for now. That will give us flexibility in the future, but I will leave it up to you.

palaparthi (approved):

schema/core.xsd
hgray-instawork: If there's any chance of additional options, I'd definitely go with enum, even if only two values

flochtililoch: I'll defer to the general opinion here and will revert back to using an enum

flochtililoch: @hgray-instawork @palaparthi I've updated this to use an enum in [b9c89ef](https://github.com/Instawork/hyperview/pull/678/commits/b9c89ef4f6b04e94232a711eba43c50a91b240f4)

palaparthi: @flochtililoch is interactive working on expo now? If not maybe we shouldn't add it?

flochtililoch: No, but I also have not added any example in the Hyperview repo to use it since it didn't work with Expo. I don't know yet if it work with bare RN app.

Issue #550, opened 2023-05-24
App bar issues, header overlaping with phone top bar / header


Tobi-De:
![WhatsApp Image 2023-05-24 at 18 44 15](https://github.com/Instawork/hyperview/assets/40334729/d8c503b2-4d17-4b2a-8c7c-04cd4d842331)

I'm not sure how this happened, I don't even know what is an app bar really (I'm not into mobile dev), but it seems hyperview rendered app have no app bar or something like that ? What I'm I missing, can styling solve the issue ?
I'm running the demo app btw

adamstep:
Hi @Tobi-De , I see you are testing on Android. Are you using a physical device, or emulator? Thanks.

adamstep:
Thanks for reporting. I believe this is an issue with not properly accounting for the safe area view (which includes the status bar rendered by Android). One more question: is this happening on every screen in the demo app, or just "Case Studies" ?

Tobi-De:
> Thanks for reporting. I believe this is an issue with not properly accounting for the safe area view (which includes the status bar rendered by Android). One more question: is this happening on every screen in the demo app, or just "Case Studies" ?

every screen

gorak:
The problem will occur on any Android device with a notched camera and fixed top navigation bar. The following changes to _Navigator.js_ adds in **SafeAreaProvider** and **SafeAreaView** to resolve the issue.

```
import * as Constants from './constants';
import HyperviewScreen from './HyperviewScreen';
import { NavigationContainer } from '@react-navigation/native';
import React from 'react';
import { createStackNavigator } from '@react-navigation/stack';
import { SafeAreaProvider, SafeAreaView } from 'react-native-safe-area-context';

const Stack = createStackNavigator();

export default () => {
  return (
    <SafeAreaProvider>
      <NavigationContainer>
        <SafeAreaView style={{ flex: 1 }}>
          <Stack.Navigator screenOptions={{ headerShown: false }}>
            <Stack.Group>
              <Stack.Screen
                component={HyperviewScreen}
                initialParams={{ url: Constants.ENTRY_POINT_URL }}
                name={Constants.MAIN_STACK_NAME}
              />
            </Stack.Group>
            <Stack.Group screenOptions={{ presentation: 'modal' }}>
              <Stack.Screen
                component={HyperviewScreen}
                name={Constants.MODAL_STACK_NAME}
              />
            </Stack.Group>
          </Stack.Navigator>
        </SafeAreaView>
[...]

PR #682, opened 2023-09-11
fix(safe area) - adding safe area to demo

hgray-instawork:
Resolves safe area handling for Android. iOS should work the same as before.

Uses the approach [recommended by react-navigation](https://reactnavigation.org/docs/7.x/handling-safe-area/) (including not using `SafeAreaView`)

Tested on iOS simulator and Android device

| | IOS | Android |
| --- | --- | --- |
| before | ios-original | android-original |
| after | ios-safe-area | android-safe-area |

Asana: https://app.asana.com/0/1204008699308084/1205467413007360/f

resolves https://github.com/Instawork/hyperview/pull/674
resolves https://github.com/Instawork/hyperview/issues/550

### DIFF
--- a/demo/src/Navigator.js
+++ b/demo/src/Navigator.js
@@ -7,31 +7,57 @@
  */
 
 import * as Constants from './constants';
+import {
+  SafeAreaProvider,
+  useSafeAreaInsets,
+} from 'react-native-safe-area-context';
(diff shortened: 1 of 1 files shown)


flochtililoch (approved):
Looks good!

PR #685, opened 2023-09-15
Fix <option> description.

pepe:
It seems to me as some forgotten after the editing "artifact."

OT: I found this system through the Hepermedia systems book and like what I see.


adamstep (approved):
thanks for the fix!

PR #686, opened 2023-09-20
chore(navigation) - cleanup attribute retrieval

hgray-instawork:
Using constants for consistency

### DIFF
--- a/src/services/navigator/types.ts
+++ b/src/services/navigator/types.ts
@@ -14,6 +14,8 @@ export const ID_MODAL = 'modal';
 export const KEY_MERGE = 'merge';
 export const KEY_SELECTED = 'selected';
 export const KEY_MODAL = 'modal';
+export const KEY_ID = 'id';
+export const KEY_TYPE = 'type';
 
 /**
  * Definition of the available navigator types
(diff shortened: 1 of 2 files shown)


src/services/navigator/types.ts
```diff
@@ -14,6 +14,8 @@ export const ID_MODAL = 'modal';
 export const KEY_MERGE = 'merge';
 export const KEY_SELECTED = 'selected';
 export const KEY_MODAL = 'modal';
+export const KEY_ID = 'id';
+export const KEY_TYPE = 'type';
```
flochtililoch, line 18: in the future, we'll need to standardize how various components define their constant for attributes / values. We had started putting these in a ATTRIBUTES dict for some components (i.e. [hv-view](https://github.com/Instawork/hyperview/blob/master/src/components/hv-view/types.js#L14), [hyper-ref](https://github.com/Instawork/hyperview/blob/master/src/core/hyper-ref/types.js#L32-L46)).

flochtililoch (approved):

src/services/navigator/types.ts
hgray-instawork: Would you rather I put them into a dict to start setting precedent?

flochtililoch: No, I think the precedent is already set on other files, but was just flagging this so we can course correct in the future. I think TypeScript can help us with enums, so we don't have 100s of floating constants. Not needed for this change though.

PR #687, opened 2023-09-20
chore(navigation) - ensure correct starting selection state

hgray-instawork:
Ensure that only one <nav-route> has a true selection state for incoming documents


hgray-instawork:
We decided this is not necessary

PR #690, opened 2023-09-25
chore (context) - merge document contexts

hgray-instawork:
- Merged the required new functionality from `route-doc` into the newly added `DocContext`
- Updated the implementation of HvRoute to provide the getter and setter
- Cleaned up usage and naming of the value


src/contexts/index.ts
```diff
@@ -25,5 +25,6 @@ export const RefreshControlComponentContext = React.createContext<
 >(undefined);
 
 export const DocContext = React.createContext<{
-  getDoc: () => TypesLegacy.Document;
+  getDoc: () => TypesLegacy.Document | undefined;
```
flochtililoch, line 28: FYI - this signature was somehow incorrect - you can see the context provider was passing a function instead of an object with `getDoc` key. We need to update the other places in the code where the context consumer calls that function directly (`this.context()`)

src/contexts/index.ts
```diff
@@ -25,5 +25,6 @@ export const RefreshControlComponentContext = React.createContext<
...
+  setDoc: (doc: TypesLegacy.Document) => void;
```
flochtililoch, line 29: since `hv-screen` consumer does not explicitly set this key, we probably should make it optional
```suggestion
  setDoc?: (doc: TypesLegacy.Document) => void;
```

flochtililoch (approved):

PR #692, opened 2023-09-26
Fix: Ignore stories rule for core components

ashwinibm:
Warn instead of erroring if stories are missing for core components. Related to https://github.com/Instawork/eslint-plugin-instawork/pull/43 where we're expanding the matcher to any `components/`

Screenshot 2023-09-26 at 8 05 36 PM

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -69,6 +69,14 @@
         "object-curly-newline": "off",
         "instawork/error-object": "off"
       }
+    },
+    {
+      "files": [
+        "src/core/components/**/*",
+      ],
+      "rules": {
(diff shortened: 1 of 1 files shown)


flochtililoch (commented):
I think we still may want to add basic stories for these components in the future, so maybe we should configure the rule to warn

ashwinibm:
> I think we still may want to add basic stories for these components in the future, so maybe we should configure the rule to warn

Fixed in https://github.com/Instawork/hyperview/pull/692/commits/ee488bb6c810d82b94c511201c4f0f4c3c0ea083

flochtililoch (approved):

PR #699, opened 2023-10-03
chore: remove DOM lib type overrides everywhere

flochtililoch:
This requires some additional explicit casting to `Element`.


src/types.ts
```diff
@@ -74,31 +74,6 @@ export const NODE_TYPE = {
 
 export type NodeType = typeof NODE_TYPE[keyof typeof NODE_TYPE];
 
-export type Node = {
```
hgray-instawork, line 77: YES!

hgray-instawork (approved):
Love it!

PR #702, opened 2023-10-03
chore(ts): package updates

flochtililoch:
- 316e181c44662d51836eae975943b005af9ddca4 Update package versions
  - Update `react`/`react-native` min version dependency to match our current app
  - Update `@instawork/xmldom` to include fixes needed for TS checks
  - Update `@tsconfig/react-native` to latest
  - Match `@types/*` minor version to their associated libraries
  - Force resolution for `react` types to avoid TS check issues due to multiple versions being brought in by various libraries
- 095007f54031981f523a032f13daf8f9e8d19e64 Hook TS checks to CI / remove flow from tests
- ef0f802307e9ec0ad62906c18211449990243f94 Fix improper type
- 371260e4d541d8c05d6f790916e618f6b1614972 Fix Jest config


src/core/components/hv-navigator/types.ts
```diff
@@ -32,5 +33,5 @@ export type NavigatorParams = {
  */
 export type Props = {
   element: TypesLegacy.Element;
-  routeComponent: FC;
+  routeComponent: FC<HvRouteProps>;
```
hgray-instawork, line 36: Should we create a local version of the props rather than reaching across modules?

```
/**
 * All of the props used by hv-route
 */
export type RouteProps = {
  navigation?: NavigatorService.NavigationProp;
  route?: NavigatorService.Route<string, RouteParams>;
};
```

hgray-instawork (approved):

src/core/components/hv-navigator/types.ts
flochtililoch: is this desirable? I would imaging Props type of a component to be part of the public interface of a module, hence available for other modules to consume. Are there down sides in your opinion in this approach?

hgray-instawork: Just keeping modules insular.

PR #706, opened 2023-10-04
chore(ts): remove `files` restriction in TS config

flochtililoch:
TS should now capture every file of the project.

This PR also fixes new errors resulting from that config update.


hgray-instawork (approved):
Thanks for fixing those tests

PR #708, opened 2023-10-04
chore(ts): migrate mocks and VisibilityDetectingView

flochtililoch:


src/VisibilityDetectingView.tsx
```diff
@@ -39,49 +37,56 @@ export default class VisibilityDetectingView extends PureComponent<Props> {
 
     // UIManager.measure may not exist during render-testing, which might break the
     // `view.measure` call
-    if (this.view && UIManager.measure) {
+    if (this.view && this.view.measure) {
```
flochtililoch, line 40: suspicious - is this change needed?

hgray-instawork: This was throwing an error saying that it always would resolve to true

hgray-instawork (approved):

src/VisibilityDetectingView.tsx
flochtililoch: I see - as the comment mentions above, the test is there for render-testing. I'll see if we can remove and mock instead the function in test env

flochtililoch: in theory this won't change the behavior since we're testing on the method we're calling, so I'll leave it alone

PR #721, opened 2023-10-10
fix: revert package.json `main` field

flochtililoch:
This was inadvertently changed in #604 - revert it back to point to the transpiled version by default


hgray-instawork (approved):
Thanks for catching that

PR #726, opened 2023-10-17
chore(behaviors): centralize behavior trigger code

hgray-instawork:
Moving the behavior trigger code out of `hyper-ref` into a shared location to allow access by other components

Asana: https://app.asana.com/0/1204008699308084/1205741965789631/f


src/types.ts
```diff
@@ -255,6 +255,23 @@ export type BehaviorRegistry = {
   [key: string]: HvBehavior;
 };
 
+// https://hyperview.org/docs/reference_behavior_attributes
+export const ATTRIBUTES = {
```
flochtililoch, line 259: might make sense to rename to `BEHAVIOR_ATTRIBUTES` to remove ambiguity since this file is central for types

src/services/behaviors/index.ts
```diff
@@ -138,3 +138,5 @@ export const trigger = (
     });
   });
 };
+
+export { createActionHandler, triggerBehaviors } from './helpers';
```
flochtililoch, line 142: was there a particular intention in putting these files under helpers instead of into this file directly?

src/services/behaviors/helpers.ts
```diff
@@ -0,0 +1,81 @@
...
+  ACTIONS,
+  ATTRIBUTES,
+  HvComponentOnUpdate,
+  NAV_ACTIONS,
+  NavAction,
+  UPDATE_ACTIONS,
+  UpdateAction,
+} from 'hyperview/src/types';
+
+export const triggerBehaviors = (
```
flochtililoch, line 19: As we're creating these new services, I'm curious what's the overlap between the new `triggerBehaviors` method and the `trigger` method available in the service's index. Maybe more just an naming issue that creates ambiguity

src/types.ts
hgray-instawork: Funny, I had that and was about to commit it but because "TRIGGERS" and others were not named such, I backed out. I'll add that back in.

src/services/behaviors/index.ts
hgray-instawork: It's not directly related to the purpose of the service. Can move if you think that makes more sense.

src/services/behaviors/helpers.ts
hgray-instawork: There is decent overlap, but there are differences in their usage. We may want to create a separate task to consolidate these.

flochtililoch (approved):

PR #727, opened 2023-10-18
chore(on update): lift onUpdate and its methods from hv-screen to hv-root

hgray-instawork:
Follow the commits for the individual steps taken
- moved the 'once' handling methods into /services/behaviors
- updated `BehaviorOptions` to properly reflect the state of the data
- created new `RootOnUpdate` and `OnUpdateCallbacks` types
- added `onUpdate` to all props which occur from `hv-root` to `hv-screen`
- ported the `onUpdate` method and all of its related methods from `hv-screen` to `hv-root`
  - use the callbacks prop to inject any of the instance-specific callbacks needed for the `onUpdate`
- inject the new `onUpdate` into the props

Asana: https://app.asana.com/0/1204008699308084/1205741965789634/f

### DIFF
--- a/src/core/components/hv-root/index.tsx
+++ b/src/core/components/hv-root/index.tsx
@@ -59,9 +59,6 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
   constructor(props: HvScreenProps.Props) {
     super(props);
 
-    this.onUpdate = this.onUpdate.bind(this);
-    this.reload = this.reload.bind(this);
-
     this.parser = new Dom.Parser(
       this.props.fetch,
       this.props.onParseBefore,
--- a/src/core/components/hv-root/index.tsx
+++ b/src/core/components/hv-root/index.tsx
@@ -80,14 +80,14 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
   reload = (
     optHref: DOMString | null | undefined,
     opts: HvComponentOptions,
-    callbacks: OnUpdateCallbacks,
+    onUpdateCallbacks: OnUpdateCallbacks,
   ) => {
     const isBlankHref =
       optHref === null ||
       optHref === undefined ||
       optHref === '#' ||
       optHref === '';
-    const stateUrl = callbacks.getState().url;
+    const stateUrl = onUpdateCallbacks.getState().url;
     const url = isBlankHref
       ? stateUrl
       : UrlService.getUrlFromHref(optHref, stateUrl || '');
@@ -124,7 +124,7 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
       Behaviors.setRanOnce(behaviorElement);
     }
 
-    let newRoot = callbacks.getDoc();
+    let newRoot = onUpdateCallbacks.getDoc();
     if (showIndicatorIdList || hideIndicatorIdList) {
       newRoot = Behaviors.setIndicatorsBeforeLoad(
         showIndicatorIdList,
@@ -134,8 +134,8 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
     }
 
     // Re-render the modifications
-    callbacks.setNeedsLoad();
-    callbacks.setState({
+    onUpdateCallbacks.setNeedsLoad();
+    onUpdateCallbacks.setState({
       doc: newRoot,
       elementError: null,
       error: null,
@@ -157,7 +157,7 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
     method: DOMString | null | undefined = Dom.HTTP_METHODS.GET,
     root: Document,
     formData: FormData | null | undefined,
-    callbacks: OnUpdateCallbacks,
+    onUpdateCallbacks: OnUpdateCallbacks,
   ): Promise<Element | null> => {
     if (!href) {
       throw new Error('No href passed to fetchElement');
@@ -174,7 +174,7 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
     try {
       const url = UrlService.getUrlFromHref(
         href,
-        callbacks.getState().url || '',
+        onUpdateCallbacks.getState().url || '',
       );
       const httpMethod: Dom.HttpMethod = method as Dom.HttpMethod;
       const { doc, staleHeaderType } = await this.parser.loadElement(
@@ -184,15 +184,15 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
       );
       if (staleHeaderType) {
         // We are doing this to ensure that we keep the screen stale until a `reload` happens
-        callbacks.setState({ staleHeaderType });
+        onUpdateCallbacks.setState({ staleHeaderType });
       }
-      callbacks.clearElementError();
+      onUpdateCallbacks.clearElementError();
       return doc.documentElement;
     } catch (err: Error | unknown) {
       if (this.props.onError) {
         this.props.onError(err as Error);
       }
-      callbacks.setState({ elementError: err as Error });
+      onUpdateCallbacks.setState({ elementError: err as Error });
     }
     return null;
   };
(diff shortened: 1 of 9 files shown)


src/types.ts
```diff
@@ -400,3 +403,23 @@ export type Fetch = (
   input: RequestInfo | URL,
   init?: RequestInit | undefined,
 ) => Promise<Response>;
+
+export type RootOnUpdate = (
```
flochtililoch, line 407: Would we be better off passing the callbacks as a separate prop, and type the `onUpdate` prop with the existing `HvComponentOnUpdate` type? I think this could help preventing duplication.

src/types.ts
```diff
@@ -369,8 +372,8 @@ export const UPDATE_ACTIONS = {
 export type UpdateAction = typeof UPDATE_ACTIONS[keyof typeof UPDATE_ACTIONS];
 
 export type BehaviorOptions = {
-  newElement: Element;
-  behaviorElement: Element;
+  newElement?: Element;
+  behaviorElement?: Element;
```
flochtililoch, line 376: it's not apparent to me why we need this change?

src/services/behaviors/index.ts
```diff
@@ -220,3 +220,22 @@ export const createActionHandler = (
...
+/**
+ * Checks if `once` is previously applied.
+ */
+export const isOncePreviouslyApplied = (behaviorElement: Element) => {
+  const once = behaviorElement.getAttribute('once');
+  const ranOnce = behaviorElement.getAttribute('ran-once');
+  if (once === 'true' && ranOnce === 'true') {
+    return true;
+  }
+  return false;
```
flochtililoch, line 240: nit/minor:
```suggestion
  return once === 'true' && ranOnce === 'true';
```

src/services/behaviors/index.ts
```diff
@@ -220,3 +220,22 @@ export const createActionHandler = (
...
+ * Set the `ran-once` attribute to true.
+ */
+export const setRanOnce = (behaviorElement: Element) => {
+  behaviorElement.setAttribute('ran-once', 'true');
+};
+
+/**
+ * Checks if `once` is previously applied.
+ */
+export const isOncePreviouslyApplied = (behaviorElement: Element) => {
```
flochtililoch, line 234: let's type the returned value
```suggestion
export const isOncePreviouslyApplied = (behaviorElement: Element): boolean => {
```

src/core/components/hv-screen/index.js
```diff
@@ -408,205 +295,27 @@ export default class HvScreen extends React.Component {
...
+      getDoc: () => this.doc,
+      getNavigation: () => this.navigation,
+      getOnUpdate: () => this.onUpdate,
+      getStateDoc: () => this.state.doc,
+      getStateUrl: () => this.state.url,
+      registerPreload: (id, element)=>this.registerPreload(id, element),
+      setNeedsLoad: () => {
+        this.needsLoad = true
+      },
+      // eslint-disable-next-line flowtype/no-types-missing-file-annotation
```
flochtililoch, line 313: I'm curious why this would be needed

src/core/components/hv-screen/index.js
```diff
@@ -15,16 +15,12 @@ import * as Events from 'hyperview/src/services/events';
...
-import { ACTIONS, NAV_ACTIONS, UPDATE_ACTIONS } from 'hyperview/src/types';
-// eslint-disable-next-line instawork/import-services
-import Navigation, { ANCHOR_ID_SEPARATOR } from 'hyperview/src/services/navigation';
-import { createProps, createStyleProp, getElementByTimeoutId, getFormData, later, removeTimeoutId, setTimeoutId, shallowCloneToRoot } from 'hyperview/src/services';
-import { Linking } from 'react-native';
+import { createProps, createStyleProp, later } from 'hyperview/src/services';
 import LoadElementError from '../load-element-error';
 import LoadError from 'hyperview/src/core/components/load-error';
 import Loading from 'hyperview/src/core/components/loading';
+// eslint-disable-next-line instawork/import-services
```
flochtililoch, line 22: I'm curious if that's still needed

src/core/components/hv-root/index.tsx
```diff
@@ -28,6 +48,460 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
...
+   *   used to render the screen.
+   * Returns a promise that resolves to a DOM element.
+   */
+  fetchElement = async (
+    href: DOMString | null | undefined,
+    method: DOMString | null | undefined = Dom.HTTP_METHODS.GET,
+    root: Document,
+    formData: FormData | null | undefined,
+    callbacks: OnUpdateCallbacks,
+  ) => {
```
flochtililoch, line 163: let's type the returned value:
```suggestion
  ): Promise<Element | null> => {
```

src/core/components/hv-root/index.tsx
```diff
@@ -28,6 +48,460 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
...
+  formComponentRegistry: ComponentRegistry;
+
+  parser: Dom.Parser;
+
+  constructor(props: HvScreenProps.Props) {
+    super(props);
+
+    this.onUpdate = this.onUpdate.bind(this);
+    this.reload = this.reload.bind(this);
+
```
flochtililoch, line 64: I know we have this in HvScreen, but this is no longer needed in modern JS (the binding already happens when we declare an instance method using the "fat arrow" syntax
```suggestion
```

src/types.ts
```diff
@@ -400,3 +403,23 @@ export type Fetch = (
...
+export type OnUpdateCallbacks = {
+  clearElementError: () => void;
+  getNavigation: () => Navigation;
+  getOnUpdate: () => HvComponentOnUpdate;
+  getStateDoc: () => Document;
+  getDoc: () => Document;
+  getStateUrl: () => string;
+  registerPreload: (id: number, element: Element) => void;
+  setNeedsLoad: () => void;
+  setState: (state: object) => void;
```
flochtililoch, line 424: any ways we could type the state with a more explicit type?

src/types.ts
hgray-instawork: I tried using a composite but I wasn't able to get it to work. Maybe I'll take another stab.
```onUpdate = HvComponentOnUpdate & { callbacks: OnUpdateCallbacks }```
But that didn't satisfy the signature for some reason.

File: src/types.ts (hunk 2 above)
Comment (hgray-instawork) [reply src/types.ts:376]: `HvComponentOptions ` is the type which populates the onUpdate. It defines these two values as optional.
https://github.com/Instawork/hyperview/blob/4a570ab01b3652fdf7840974bb50b0a958c724e8/src/types.ts#L180
https://github.com/Instawork/hyperview/blob/4a570ab01b3652fdf7840974bb50b0a958c724e8/src/types.ts#L206

I can't pass an optional value from `HvComponentOptions` into the required value in here (`BehaviorOptions`). Further, I'm not sure those two values are even needed, but this was the quickest way to get past the error.

You can see here that those values don't appear to be read from the opts at all
https://github.com/Instawork/hyperview/blob/4a570ab01b3652fdf7840974bb50b0a958c724e8/src/services/navigation/index.ts#L71-L74

File: src/services/behaviors/index.ts (hunk 3 above)
Comment (hgray-instawork) [reply src/services/behaviors/index.ts:240]: I just copied what we had.

File: src/core/components/hv-screen/index.js (hunk 5 above)
Comment (hgray-instawork) [reply src/core/components/hv-screen/index.js:313]: This is part of the reload functionality, now moved to `hv-root`
https://github.com/Instawork/hyperview/blob/dda5288214a223851ca56536df82baaa63d557f9/src/core/components/hv-root/index.tsx#L138-L139


In the original implementation, `hv-screen` set it for itself.
https://github.com/Instawork/hyperview/blob/c6435a7e6f3240b51a131ef6dd7282ae0b93175a/src/core/components/hv-screen/index.js#L283-L284

I didn't want to reach into `hv-screen` from `hv-root` so I provided a callback to allow it to set its value.

File: src/core/components/hv-screen/index.js (hunk 6 above)
Comment (hgray-instawork) [reply src/core/components/hv-screen/index.js:22]: It is. Without it, it wants me to import "Navigation" as a wildcard and that doesn't allow it to be instantiated with "new". Another area where we probably want to do some cleanup eventually and not export it as the default value.

File: src/core/components/hv-root/index.tsx (hunk 8 above)
Comment (hgray-instawork) [reply src/core/components/hv-root/index.tsx:64]: Thanks, I was very curious about what benefit it was providing. Answers online didn't provide much insight.

File: src/types.ts (hunk 9 above)
Comment (hgray-instawork) [reply src/types.ts:424]: I started down that road, but realized that `hv-screen` and `hv-route` have different states (at this point). I can take another look at doing this as I think all of the values are optional in the state object.

File: src/core/components/hv-screen/index.js (hunk 5 above)
Comment (flochtililoch) [reply src/core/components/hv-screen/index.js:313]: I meant the eslint comment that disable flow annotation - given the file is not/shouldn't be covered by flow

Comment (hgray-instawork) [reply src/core/components/hv-screen/index.js]: Ha, yeah, I removed that in the next commits. I didn't need to define a type there.

File: src/types.ts (hunk 1 above)
Comment (flochtililoch) [reply src/types.ts:408]: I meant, the component would have two props:
```js
...
onUpdate: HvComponnetUpdate;
onUpdateCallbacks: OnUpdateCallbacks;
```

hgray-instawork: Resolved by adding to the `HvComponentOptions`

PR #728, opened 2023-10-19
chore(document): account for possible null doc in state

hgray-instawork:
Both `hv-screen` and `hv-route` have the potential to have a null `doc` in state. This can happen from a bad load, or from an override document being provided by a parent component.
The code used in `onUpdate` and its related methods were not accounting for the possibility of a null `doc'.

- Updated `ScreenState` to reflect the values of the state as assigned in `hv-screen`
- Updated callback signature to allow `getDoc` to return null
- Updated `hv-root` implementation of `onUpdate` and related to account for null doc
- Updated `HvGetRoot` to account for null doc
- Updated behavior implementations which used `HvGetRoot` to account for null doc

Asana: https://app.asana.com/0/0/1205741965789643/f

### DIFF
--- a/src/behaviors/hv-hide/index.ts
+++ b/src/behaviors/hv-hide/index.ts
@@ -36,9 +36,9 @@ export default {
 
     const hideElement = () => {
       const doc: Document | null = getRoot();
-      const targetElement: Element | null | undefined = doc
-        ? doc.getElementById(targetId)
-        : null;
+      const targetElement: Element | null | undefined = doc?.getElementById(
+        targetId,
+      );
       if (!targetElement) {
         return;
       }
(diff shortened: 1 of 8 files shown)


flochtililoch (approved):
Looks good overall. One minor improvement would be to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) instead of ternary expression when assigning variables, i.e.

instead of 
```js
const targetElement: Element | null | undefined = doc
  ? doc.getElementById(targetId)
  : null;
```

do

```js
const targetElement: Element | null | undefined = doc?.getElementById(targetId);
```

hgray-instawork:
Yes, of course. Don't know why I didn't do that first time around. Fixed.

PR #729, opened 2023-10-20
chore(hvroute): hvroute as an onUpdate provider for navigator

hgray-instawork:
Implement `hv-route` as a provider of `onUpdate` to allow navigators to process behaviors

Only routes which own a document are providers, any children which do not own a document will inherit.

Asana: https://app.asana.com/0/1204008699308084/1205741965789637/f

### DIFF
--- a/src/contexts/index.ts
+++ b/src/contexts/index.ts
@@ -7,7 +7,7 @@
  */
 
 import type { ComponentType } from 'react';
-import type { HvComponentOnUpdate } from '../types';
+import type { HvComponentOnUpdate } from 'hyperview/src/types';
 import React from 'react';
 import type { RefreshControlProps } from 'react-native';
 
--- a/src/contexts/index.ts
+++ b/src/contexts/index.ts
@@ -7,6 +7,7 @@
  */
 
 import type { ComponentType } from 'react';
+import type { HvComponentOnUpdate } from '../types';
 import React from 'react';
 import type { RefreshControlProps } from 'react-native';
 
@@ -27,3 +28,8 @@ export const DocContext = React.createContext<{
   getDoc: () => Document | undefined;
   setDoc?: (doc: Document) => void;
 } | null>(null);
+
+export const OnUpdateContext = React.createContext<{
+  onUpdate: HvComponentOnUpdate;
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+}>({ onUpdate: () => {} });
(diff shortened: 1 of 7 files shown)


src/core/components/hv-root/index.tsx
```diff
@@ -220,7 +220,10 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
       if (navigation && doc) {
         const { behaviorElement, delay, newElement, targetId } = options;
         const delayVal: number = +(delay || '');
-        navigation.setUrl(options.onUpdateCallbacks.getState().url || '');
+        const state = options.onUpdateCallbacks.getState();
+        if (state.url) {
+          navigation.setUrl(state.url);
```
flochtililoch, line 225: we no longer set the URL to '' - there's no side effect with this change?

flochtililoch (approved):

src/core/components/hv-root/index.tsx
hgray-instawork: The url in `navigation` is used to get the domain for other requests, it seems to have no other purpose. Presumably once it's set once, the domain should remain the same.

setUrl() only accepts `string` not `string | null`. The previous code I had in there was to work around state's url value supporting null values. I also changed the state setter in `hv-route` so that doc and url will always either both have a value or neither will have a value, so this part of the code should not actually ever be reached if the doc+url are null. (There is a null check for `doc` a little higher up)

hgray-instawork: Code rewritten to improve intention

src/contexts/index.ts
```diff
@@ -7,6 +7,7 @@
  */
 
 import type { ComponentType } from 'react';
+import type { HvComponentOnUpdate } from '../types';
```
flochtililoch, line 10:
```suggestion
import type { HvComponentOnUpdate } from 'hyperview/src/types';
```

src/core/components/hv-route/index.tsx
```diff
@@ -178,6 +212,46 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, Types.State> {
     this.props.setPreload(id, element);
   };
 
+  /**
+   * Implement the callbacks from this class
+   */
+  updateCallbacks = (): OnUpdateCallbacks => {
+    if (!this.state.doc || !this.props.navigation) {
+      throw new NavigatorService.HvRouteError(
```
flochtililoch, line 220: how will this error bubble up? Is this mostly a guard for TS, or is there a risk this will be raised in which case shouldn't we try make the return value optional?

hgray-instawork: You are correct. The doc is optional and the props.navigation is no longer necessary

PR #734, opened 2023-11-01
chore: add on-event behavior to allowed list for navigation behaviors

hgray-instawork:
Deep links would benefit from allowing `on-event` triggered behaviors


flochtililoch (commented):
Where are the event listeners added when loaded in the context of a navigator? From the change, I'm under the impression that we're only changing the "validation" aspect.
Also side note, if we go with this change, there's a few changes that will need to be done (presumably `triggerLoadBehaviors` will need to be either renamed or limited to filtering out load behaviors, and the comment above `updateBehaviorElements` is also specific to "load")

hgray-instawork:
Closing. Replaced by https://github.com/Instawork/hyperview/pull/946

PR #736, opened 2023-11-01
proposal: preserve fetching order

flochtililoch:
This PR introduce a new behavior boolean attribute, `preserve-order`. This attribute is used by behaviors that "fetch" content, i.e. anytime there's a `href` attribute set, with an update action such as `replace`, `replace-inner`, etc. When present, we compute a unique id before the request is emitted, and store it on an attribute of the target element. When the response arrives, we read the attribute of the target element and compare it to the one we generated before the request was sent. If the attribute is set with a value that's different from the one set before the request, we assume this means another request/response cycle took place, and the response we're presently dealing with is "outdated", so we discard it.

To be discussed:
- [ ] naming: is "preserve-order" descriptive enough? should we use a boolean or an enum?
- [ ] visibility: should we make this feature public / documented or keep it "dark" for now?
- [ ] scope: should we think of this capability to be applied to other, non-fetching behaviors?

### DIFF
--- a/examples/case_studies/contacts/index.xml.js
+++ b/examples/case_studies/contacts/index.xml.js
@@ -11,7 +11,9 @@ module.exports = function handler(req, res, next) {
   const { query } = urlParse(req.originalUrl, true);
 
   // no search or next page? pass through to 11ty to render the entire document
-  if (query.search === undefined && !query.page) {
+  // note: we also test for "template" to work mitigate this bug
+  // https://github.com/Instawork/hyperview/issues/735
+  if (query.search === undefined && !query.page && !query.template) {
     next();
     return;
   }
--- a/examples/case_studies/contacts/list.xml.njk
+++ b/examples/case_studies/contacts/list.xml.njk
@@ -37,6 +37,8 @@
         action="replace"
         target="contacts"
         href="?template=list"
+        debounce="200"
+        preserve-order="true"
         show-during-load="loading-indicator"
       />
     </text-field>
--- a/src/core/components/hv-screen/index.js
+++ b/src/core/components/hv-screen/index.js
@@ -505,27 +505,52 @@ export default class HvScreen extends React.Component {
       doc: newRoot,
     });
 
+    // If a target is specified and exists, use it. Otherwise, the action target defaults
+    // to the element triggering the action.
+    const getTargetElement = () => {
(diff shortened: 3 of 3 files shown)


palaparthi:
> naming: is "preserve-order" descriptive enough?

"cancel-stale-requests" or "discard-stale-requests" might be more descriptive?

adamstep:
I was curious how this is implemented in htmx. Looks like that library stores a queue of requests, and using an attribute there are different strategies for managing the queue if the element issues multiple requests. See the docs here: https://htmx.org/attributes/hx-sync/

Do you think a similar approach could work in Hyperview?

adamstep:
Source code from htmx to deal with this: https://github.com/bigskysoftware/htmx/blob/master/src/htmx.js#L2980

hgray-instawork (approved):

adamstep:
Looking at the htmx implementation a bit more, I think perhaps we could support this as follows:

Introduce two new optional behavior attributes:
- `sync-id`: An ID that will be used to synchronize requests. Could be the same as the ID of the element triggering the behavior, but doesn't need to be
- `sync-method`:  `drop` (default), `replace`
    - `drop`: if there's a request in flight associated with the sync ID, drop this request
    - `replace`: if there's a request in flight associated with the sync ID, abort the in-flight request and make the new request

Over time, we can add queuing support as well if we think it's necessary.

**Implementation:**
- Capture the two new attributes and pass them to `Parser.loadElement`.
- In `Parser`, create an object that maps the sync IDs to the current request (promise).
- In `loadElement`:
   - if NO SYNC ID: proceed as before, make request
   - if HAS SYNC ID: look up the value in the map for the ID
     - if REQ IN FLIGHT FOR SYNC ID: check the sync method. If `drop`, do a no-op. If `replace`, make the request and update the promise in the map
     - if NO REQ IN FLIGHT FOR SYNC ID: make the request and update the promise in the map
  - When we get a response:
    - if NO SYNC ID: proceed as before, return new doc
    - if HAS SYNC ID: look up value in the map for the ID. If method is `replace` and the promise in the map is different, do a no-op (we've been replaced with a newer request).

We will need to update the return type of `loadElement` so that it can either contain a doc/headerType or a `no-op` constant, indicating to the caller that no change should happen.

adamstep:
Superceded with syncing implementation

PR #740, opened 2023-11-03
chore: test schema changes

flochtililoch:
Add new directory `test/schema/` that contains raw xml files that are validated with the `test:validate-xml` command.


hgray-instawork (approved):
Great!

PR #747, opened 2023-11-08
feat: report route focus and blur events

hgray-instawork:
Allow navigation actions to report the blur/focus state of each route. The entire `route` object is returned. This is a combination of
- Elements injected by react-navigation (key and maybe other values)
- Elements injected by Hyperview (params)
  - The `id` is pulled from the `<nav-route>` attribute in the navigation structure.
  - The `url` is from either the `<nav-route>` `url` attribute or from the `href` passed in a `<behavior>`

The `id` will be undefined for any navigation events which occur from `<behavior>` triggers.
The 'url' will be undefined for any `<nav-route>` which does not define an `href` value.

I have verified that the listeners are properly cleaned and resubscribed without causing duplicate listeners.

Asana: https://app.asana.com/0/1204008699308084/1205094358200852/f

https://github.com/Instawork/hyperview/assets/127122858/40f3a3ea-c046-4b94-b514-39f9ff4cb708

### DIFF
--- a/src/contexts/navigation.ts
+++ b/src/contexts/navigation.ts
@@ -11,7 +11,7 @@ import type {
   HvComponent,
   HvComponentOnUpdate,
   Reload,
-  RouteParams,
+  Route,
 } from 'hyperview/src/types';
 
 import React, { ComponentType, ReactNode } from 'react';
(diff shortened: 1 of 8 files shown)


flochtililoch (commented):
discussed offline - the callbacks might instead just pass the `route` object, and leave it to the consumer to pick some content out of it

hgray-instawork:
Changed to pass the `routeParams` object

flochtililoch (approved):
LGTM

flochtililoch (approved):

PR #748, opened 2023-11-09
chore: cleanup legacy types

hgray-instawork:
The 'types-legacy' naming came about as a result of navigation work performed in TypeScript while the bulk of Hyperview was in Javascript. After the TypeScript migration, some remnants of this naming and extra files remained.

- Removed `types-legacy.ts` - this file was redundant with `types.ts` as all of the updates were already ported
- Removed references to `types-legacy.ts`
- Removed usages of `TypesLegacy` imports - these were largely used when using wildcard imports from the root types file
- Replaced all wildcard imports of src/types with importing the specific types and constants required
- Removed namespaces for any uses of the root types

Viewing the sequence of commits will help seeing the changes as the prettier formatting makes it look like a lot more was changed.

Asana: https://app.asana.com/0/1204008699308084/1205741965789657/f


flochtililoch (approved):
🎉

PR #752, opened 2023-11-14
fix: deep link merging

hgray-instawork:
When an existing structure exists and a deep link comes in, merging must take into account that the document structure has changed:
1. The primary element may change from a <navigator> to a <screen> or vice versa
2. Any navigator could change types (stack, tab)

These changes validate the incoming type matches the existing. In the case of mismatches, the incoming data replaces existing.

Commits were performed sequentially for easier visibility.

Example below uses an error page to illustrate the issue.

| Before | After |
| -- | -- |
| before | after |

Asana: https://app.asana.com/0/1204008699308084/1205888766113001/f

### DIFF
--- a/src/services/dom/helpers.ts
+++ b/src/services/dom/helpers.ts
@@ -90,3 +90,18 @@ export const preorder = (
   }
   return acc;
 };
+
+/**
+ * Get an array of the direct decendant elements of a document or element
+ */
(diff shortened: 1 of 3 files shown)


hgray-instawork:
I found a similar utility elsewhere, I'll replace it.

flochtililoch (approved):

PR #753, opened 2023-11-17
fix: deep link state handling

hgray-instawork:
An issue was identified where the deep link state of a stack navigator could be handled incorrectly when the incoming navigation didn't include changes to the stack navigator. At the same time, we decided to improve the logic of incoming state affects existing state.

We determined that in cases where a user has already opened views, and they match the order of the incoming deep link, they can remain with their existing state. For example:
- user clicks "company", then views deep link "company"
  - in this case, the user's action is identical to the deep link so their view remains including any existing state

This shows the original bug finding and its resolution

| Before | After |
| -- | -- |
| before | after |

In the following examples, I have modified my "developer menu" to launch storybook directly.

| | |
| -- | -- |
| Launching storybook as a view or by deep link independently work as expected |  ex1 |
| Viewing storybook then triggering deep link retains state | ex2 |
| Triggering deep link with company after storybook retains state (storybook in same sequential position) | ex3 |
| Triggering deep link with storybook after company resets state (not same sequence) | ex4 |
| Triggering deep link with new base resets state (while storybook is still in same index, the changed base forces redraw) | ex5 |

Asana: https://app.asana.com/0/1204008699308084/1205888765821740/f

### DIFF
--- a/src/core/components/navigator-stack/router.tsx
+++ b/src/core/components/navigator-stack/router.tsx
@@ -9,6 +9,16 @@
 import * as NavigatorService from 'hyperview/src/services/navigator';
 import * as Types from './types';
 import { StackNavigationState, StackRouter } from '@react-navigation/native';
+import { LOCAL_NAME } from 'hyperview/src/types';
+
+type Route = {
+  key: string;
+  name: string;
+  params?: {
+    id?: string;
+    url?: string;
+  };
+};
 
 /**
  * Provides a custom stack router that allows us to set the initial route
@@ -21,8 +31,10 @@ export const Router = (stackOptions: Types.StackOptions) => {
 
     getInitialState(options: Types.RouterConfigOptions) {
       const initState = router.getInitialState(options);
-
-      return mutateState(initState, { ...options, routeKeyChanges: [] });
+      return mutateState(initState, stackOptions, {
+        ...options,
+        routeKeyChanges: [],
+      });
     },
 
     getStateForRouteNamesChange(
@@ -30,7 +42,7 @@ export const Router = (stackOptions: Types.StackOptions) => {
       options: Types.RouterRenameOptions,
     ) {
       const changeState = router.getStateForRouteNamesChange(state, options);
-      return mutateState(changeState, options);
+      return mutateState(changeState, stackOptions, options);
     },
   };
 };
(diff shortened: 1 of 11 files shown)


src/core/components/navigator-stack/router.tsx
```diff
@@ -9,6 +9,16 @@
 import * as NavigatorService from 'hyperview/src/services/navigator';
 import * as Types from './types';
 import { StackNavigationState, StackRouter } from '@react-navigation/native';
+import { LOCAL_NAME } from 'hyperview/src/types';
+
+type Route = {
```
flochtililoch, line 14: do we not have a similar type defined already somewhere else? should this belong in `types.ts`?

src/core/components/navigator-stack/router.tsx
```diff
@@ -40,29 +52,63 @@ export const Router = (stackOptions: Types.StackOptions) => {
...
+      ? NavigatorService.getUrlFromHref(r.params.url, entrypointUrl)
+      : undefined;
   });
 
-  filteredNames.forEach((name: string) => {
-    if (options.routeParamList) {
-      const params = options.routeParamList[name] || {};
-      if (!routes.find(route => route.name === name)) {
+  let sequenceBroken = false;
+  for (let i = 0; i < elementRoutes.length; i += 1) {
```
flochtililoch, line 78: the logic seems a bit intricate, would it be worth pulling it out in a helper method and write unit tests against it?

src/services/navigator/helpers.test.ts
```diff
@@ -863,3 +873,55 @@ describe('mergeDocuments', () => {
     });
   });
 });
+
+describe('add route', () => {
```
flochtililoch, line 877: minor/nit: the first arg of the top level `describe` is generally the function name that's being tested
 ```suggestion
describe('addStackRoute', () => {
```

File: src/services/navigator/helpers.test.ts [hunk 4]
```diff
@@ -863,3 +873,55 @@ describe('mergeDocuments', () => {
...
+  it('should find 2 route elements', () => {
+    const navigators = doc.getElementsByTagNameNS(
+      Namespaces.HYPERVIEW,
+      'navigator',
+    );
+    const routes = getChildElements(navigators[0]);
+    expect(routes.length).toEqual(2);
+  });
+
+  describe('add stack route', () => {
```
flochtililoch, line 889: maybe that describe text can be more specific to what the nested tests are about

src/services/navigator/helpers.test.ts
```diff
@@ -863,3 +873,55 @@ describe('mergeDocuments', () => {
...
+
+    it('should contain added route', () => {
+      expect(routes[2].getAttribute('id')).toEqual('test-route');
+    });
+
+    removeStackRoute(doc, '/a/b/c', 'http://foo.com');
+  });
+});
+
+describe('find navigator by id', () => {
```
flochtililoch, line 916:
```suggestion
describe('getNavigatorById', () => {
```

src/core/components/navigator-stack/router.tsx
hgray-instawork: Marked as TODO for now

src/core/components/navigator-stack/router.tsx
hgray-instawork: Moved to 'helpers.ts', added tests.

src/services/navigator/helpers.test.ts
hgray-instawork: Fixed

src/services/navigator/helpers.test.ts
hgray-instawork: Fixed

src/services/navigator/helpers.test.ts
hgray-instawork: Fixed

flochtililoch (approved):

PR #758, opened 2023-11-27
fix: format single digit months in date-field

### LINKED ISSUE #757: Date field value is not always YYYY-MM-DD
`createStringFromDate` in `<date-field>` states that it [should return](https://github.com/Instawork/hyperview/blob/master/src/components/hv-date-field/index.tsx#L47) `YYYY-MM-DD`, but for example June 10 2023 becomes `2023-6-10` (`YYYY-M-DD`). This breaks parsing of the date in some languages.

I can create a PR with a solution if you want to. Something like this should solve it:

```js
// Before
const month = date.getMonth() + 1;

// After
const month = `${date.getMonth() + 1}`.padStart(2, '0');
```

believer:
Fixes #757 

Previously, dates where the month is a single digit (January – September / 1 - 9) created formatted dates like `2023-1-27`. This would break parsing the date as an ISO string in certain languages.

This PR pads the month part with a leading zero to make it correctly formatted as `YYYY-MM-DD`.

### DIFF
--- a/src/components/hv-date-field/index.test.ts
+++ b/src/components/hv-date-field/index.test.ts
@@ -61,4 +61,22 @@ describe('HvDateField', () => {
       );
     });
   });
+
+  describe('createDateFromString', () => {
+    it('returns null if no date', () => {
+      expect(HvDateField.createDateFromString(null)).toBeNull();
+    });
+
+    it('returns correct date for double digit month', () => {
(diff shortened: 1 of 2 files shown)


hgray-instawork (approved):
Looks good to me. Could you add tests for `createDateFromString` as well?

hgray-instawork:
Thanks very much for the fix!

PR #761, opened 2023-11-29
fix: avoid multiple DOM mutations for dynamic stack navigator

hgray-instawork:
When the component is re-rendered, there was a possibility of adding a duplicate node into the DOM.

This is a follow up to the work from https://github.com/Instawork/hyperview/pull/760

Asana: https://app.asana.com/0/1204008699308084/1206058760311122/f


flochtililoch (approved):
I don't fully understand the fix, but I assume you have tested it well with various scenarios.

hgray-instawork:
Just checking that the <navigator> hasn't already been created before creating a new one. They all get destroyed when the parent route is removed as part of the closure.

PR #766, opened 2023-12-07
fix: refresh control should only be added when a refresh trigger exists

hgray-instawork:
Check for the "refresh" trigger before adding the <RefreshControl> to `list` and `section-list`.

Confirmed in Pro app and HV Demo.

Fixes https://instawork.sentry.io/issues/4682025463/
Fixes https://instawork.sentry.io/issues/4585045549/

Sentry: this should resolve any of the `Error: No href passed to fetchElement` errors which contain `hyperview/worker_app/helpdesk-parent` at the end of the breadcrumbs.

| Before | After |
| -- | -- |
| before | after |

Asana: https://app.asana.com/0/1204008699308084/1206114436157353/f


flochtililoch (approved):
thanks for the fix!

PR #771, opened 2023-12-14
feature: add ability to block back navigation from hardware or software event

hgray-instawork:
Allow developers to block the user going back or closing a screen. Example use cases:
- User has unsaved changes
- User may have not scrolled to see all content
- User may miss important information on the screen

The implementation adds a new "back" trigger into Hyperview.

See the commits for a breakdown of the changes.

| Software back | iOS swipe | Android hardware |
| -- | -- | -- |
| software | iOS | android |

**Example implementation**
In the following example, when the user attempts to go back, the top behavior with `trigger="back"` will be called. Since there is a behavior to handle `back`, the normal navigation is canceled. The resulting `alert` performs two behaviors on click of the OK button: it hides the containing view (thus removing behaviors) and then it calls an `action="back"` which will continue the back navigation.

```XML
<view id="alert-container">
  <behavior
    trigger="back"
    action="alert"
    alert:title="You have unsaved changes. Go back?"
  >
  <alert:option alert:label="Cancel" />
  <alert:option
    alert:label="OK"
  >
    <behavior
      action="hide"
      target="alert-container"
    />
    <behavior
      action="back"
    />
  </alert:option>
  </behavior>
</view>
```

Asana: https://app.asana.com/0/1204008699308084/1205954494594099/f


flochtililoch (commented):
Couple of questions/comments regarding implementation:
- why not reusing `HvDocState` as a provider for the behavior elements instead of creating a new component?
- the naming `ElementRegistry` might be ambiguous/inconsistent with regards to other existing registries (`ComponentRegistry`, `BehaviorRegisty`), which are simple object and not React components. It also is a bit too generic ("Element") since we're only intending to store references of Behavior elements

hgray-instawork:
> Couple of questions/comments regarding implementation:
> 
> * why not reusing `HvDocState` as a provider for the behavior elements instead of creating a new component?

Since the two serve different purposes, I didn't want to combine and potentially confuse. This new one is portable and may find other uses.

> * the naming `ElementRegistry` might be ambiguous/inconsistent with regards to other existing registries (`ComponentRegistry`, `BehaviorRegisty`), which are simple object and not React components. It also is a bit too generic ("Element") since we're only intending to store references of Behavior elements

Agreed, I was having a hard time with naming. I was thinking about "cache" instead of "registry", and I was trying to decide if "behavior" and "trigger" should be part of the name. But, the things that are captured are just `Element` and are gathered by the trigger value. Would `BehaviorTriggerCacheContext` be a better name?

adamstep:
As discussed in the TAD, are you considering the name `on-back` for the trigger?

hgray-instawork:
We thought it may be better to stick with the simple "back" as only one other trigger is currently "on-".

src/contexts/index.ts
```diff
@@ -39,12 +39,24 @@ export type SetState = (
   callback?: () => void,
 ) => void;
 
-export type DocStateContextProps = {
+export type BackBehaviorElements = {
+  add: (elements: Element[]) => void;
+  get: () => Element[];
+  remove: (elements: Element[]) => void;
+};
```
flochtililoch, line 46: Instead of defining this as a `type`, I think it might be worth defining it as a class that will encapsulate the actual logic for these methods:

```suggestion
export class BackBehaviorElements {
  add: (elements: Element[]) => void = {
    ...
  },
  ...
```

then it can simply be instantiated when needed.

src/core/components/hv-doc/index.tsx
```diff
@@ -39,12 +40,36 @@ export default class HvDocState extends PureComponent<
...
     };
+    this.backBehaviorElementCache = [];
   }
 
   render(): React.ReactNode {
     return (
-      <Contexts.DocStateContext.Provider
+      <Contexts.DocContext.Provider
         value={{
+          backBehaviorElements: {
```
flochtililoch, line 50: Following my other comment, the instance of `BackBehaviorElements` could be created in the constructor of this component, stored on an instance property and referenced here - that way we don't recreate the object with inner functions on every render cycle.

src/core/components/hv-doc/index.tsx
```diff
@@ -39,12 +40,36 @@ export default class HvDocState extends PureComponent<
...
+            add: (elements: Element[]): void => {
+              if (elements?.length > 0) {
+                elements.forEach(e => {
+                  this.backBehaviorElementCache.push(e);
+                });
```
flochtililoch, line 55: I think the check for length > 0 is not needed, and the expression can be simplified to:
```suggestion
              this.backBehaviorElementCache.push(...elements);
```

src/core/components/hv-doc/index.tsx
```diff
@@ -39,12 +40,36 @@ export default class HvDocState extends PureComponent<
...
+                elements?.length > 0 &&
+                this.backBehaviorElementCache.length > 0
+              ) {
+                elements.forEach(e => {
+                  const ind = this.backBehaviorElementCache.indexOf(e);
+                  if (ind > -1) {
+                    this.backBehaviorElementCache.splice(ind, 1);
+                  }
+                });
+              }
```
flochtililoch, line 70: Given the types of `elements` and `this.backBehaviorElementCache`, I think the checks for `elements?.length` and `this.backBehaviorElementCache.length` are not needed, and it might be more "elegant" to use `Array.reduce` to perform the mutation of `this.backBehaviorElementCache` only once:

```suggestion
              this.backBehaviorElementCache = elements.reduce((acc, e) => {
                const i = acc.indexOf(e);
                return i > -1
                  ? [...acc].splice(i, 1)
                  : acc;
                }, this.backBehaviorElementCache);
```

src/core/components/hv-doc/index.tsx
```diff
@@ -29,6 +27,9 @@ export default class HvDocState extends PureComponent<
   // </HACK>
   localDoc: Document | null = null;
 
+  // Cache elements
+  backBehaviorElementCache: Element[];
```
flochtililoch, line 31: minor/nitpick - if this instead stores an instance of the `BackBehaviorElements` class, we might want to rename / drop the `Cache` suffix.

flochtililoch (commented):
Looks good overall from a functionality stand point, nice work!
I made suggestions to improve the code for optimization / readability.

src/contexts/index.ts
hgray-instawork: Done.

src/core/components/hv-doc/index.tsx
hgray-instawork: Done

src/core/components/hv-doc/index.tsx
hgray-instawork: Thanks, I meant to go back and look for a cleaner way to implement this push

src/core/components/hv-doc/index.tsx
hgray-instawork: Almost worked. The return from splice() is the array which was sliced out. I'm returning the inverse of that. I tried using toSpliced() but it wasn't available in this case.

src/core/components/hv-doc/index.tsx
hgray-instawork: Done

src/contexts/index.ts
```diff
@@ -39,12 +39,41 @@ export type SetState = (
   callback?: () => void,
 ) => void;
 
-export type DocStateContextProps = {
+export class BackBehaviorElements {
+  backBehaviorElements: Element[] = [];
```
flochtililoch, line 43: minor/nit: naming - since the class now confers the name spacing, that internal prop can be simplified to `elements`

src/contexts/index.ts
```diff
@@ -39,12 +39,41 @@ export type SetState = (
...
+  add = (elements: Element[]): void => {
+    this.backBehaviorElements.push(...elements);
+  };
+
+  get = (): Element[] => this.backBehaviorElements;
+
+  remove = (elements: Element[]): void => {
+    if (elements.length > 0) {
+      console.log(
+        `Removing ${elements.length} elements from backBehaviorElements cur length: ${this.backBehaviorElements.length}`,
```
flochtililoch: I don't think we want to emit logs via `console.log` as these create performance bottleneck in the production bundle (https://reactnative.dev/docs/performance.html#using-consolelog-statements) and we have not configured our apps to remove these (which might be a performance improvement axis we explore soon).
In the future, we might want to create a logger class for Hyperview so the debugging is possible without affecting production.

flochtililoch (approved):
LGTM!

PR #772, opened 2023-12-15
fix: getElementById errors

hgray-instawork:
Attempting to reduce getElementById errors by evaluating whether the passed object is an element. Barring success, improve error logs to include the id being retrieved.

The bug is not easily reproducible locally, putting this into the next release should give us more information into what's happening.

Example of the existing error: https://instawork.sentry.io/issues/4653618920/?query=getElementById&referrer=issue-stream&sort=freq&statsPeriod=30d&stream_index=1

Asana: https://app.asana.com/0/1204008699308084/1205467413007358/f

### DIFF
--- a/src/core/components/hv-navigator/index.tsx
+++ b/src/core/components/hv-navigator/index.tsx
@@ -334,9 +334,8 @@ export default class HvNavigator extends PureComponent<Props> {
 
     // Mutate the dom to match
     if (props.doc) {
-      const route: Element | null = props.doc.getElementById(
-        this.props.params.id,
-      );
+      const route: Element | null =
+        Dom.getElementById(props.doc, this.props.params.id) || null;
       if (route && !NavigatorService.getNavigatorById(props.doc, navigatorId)) {
         const navigator = props.doc.createElementNS(
           Namespaces.HYPERVIEW,
--- a/src/services/dom/errors.ts
+++ b/src/services/dom/errors.ts
@@ -92,7 +92,7 @@ const docToString = (doc: Document): string => {
   try {
     const serializer = new XMLSerializer();
     return serializer.serializeToString(doc);
-  } catch {
-    return 'serializing error';
+  } catch (e) {
+    return `serializing error: ${e.message}`;
   }
 };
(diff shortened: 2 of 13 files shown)


src/services/dom/helpers.ts
```diff
@@ -90,3 +90,36 @@ export const preorder = (
...
+/**
+ * Attempt to find an element by id in the given node
+ * Handle cases where an element is passed in instead of a document
+ */
+export const getElementById = (
+  doc: Document | null | undefined,
+  id: string,
+): Element | null | undefined => {
+  if (!doc) {
+    return doc;
```
flochtililoch, line 103: I assume here Document would be `null` or `undefined`, but I wonder if that makes sense to return "doc". Should we explicitly return `null` instead?

src/services/dom/helpers.ts
```diff
@@ -90,3 +90,36 @@ export const preorder = (
...
+  }
+  try {
+    return doc.getElementById(id);
+  } catch {
```
flochtililoch, line 107: lets discuss this implementation on zoom

src/core/components/hv-navigator/index.tsx
```diff
@@ -334,9 +334,8 @@ export default class HvNavigator extends PureComponent<Props> {
 
     // Mutate the dom to match
     if (props.doc) {
-      const route: Element | null = props.doc.getElementById(
-        this.props.params.id,
-      );
+      const route: Element | null =
+        Dom.getElementById(props.doc, this.props.params.id) || null;
```
flochtililoch, line 338: do we need the `or` statement here?
```suggestion
        Dom.getElementById(props.doc, this.props.params.id);
```

src/services/dom/errors.ts
```diff
@@ -70,3 +73,26 @@ export class ServerError extends Error {
...
+    this.setExtraContext('error', error);
+  }
+}
+
+const docToString = (doc: Document): string => {
+  try {
+    const serializer = new XMLSerializer();
+    return serializer.serializeToString(doc);
+  } catch {
+    return 'serializing error';
```
flochtililoch, line 96: it's probably worth getting the error message
```suggestion
  } catch (e) {
    return `serializing error: ${e.message}`;
```

flochtililoch (approved):

src/core/components/hv-navigator/index.tsx
hgray-instawork: We do because getElementById returns "Element | null | undefined" which doesn't match "Element | null"

flochtililoch (approved):

PR #783, opened 2024-01-09
fix: replacement behavior issue

hgray-instawork:
Dimpal reported an issue with live shift booking. Passing the hv-route's onUpdate was causing the wrong document to be targeted in behaviors.

Asana: https://app.asana.com/0/0/1206225556817618/f

### DIFF
--- a/src/core/components/hv-route/index.tsx
+++ b/src/core/components/hv-route/index.tsx
@@ -556,7 +556,7 @@ function RouteFC(props: Types.FCProps) {
   return (
     <HvRouteInner
       // eslint-disable-next-line react/jsx-props-no-spreading
-      {...{ ...props, navigationService, navigator, needsLoad, onUpdate }}
+      {...{ ...props, navigationService, navigator, needsLoad }}
     />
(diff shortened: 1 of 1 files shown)


src/core/components/hv-route/index.tsx
```diff
@@ -556,7 +556,7 @@ function RouteFC(props: Types.FCProps) {
   return (
     <HvRouteInner
       // eslint-disable-next-line react/jsx-props-no-spreading
-      {...{ ...props, navigationService, navigator, needsLoad, onUpdate }}
+      {...{ ...props, navigationService, navigator, needsLoad }}
```
flochtililoch, line 559: does `HvRouteInner` end up having access to the `onUpdate` via another mean?

hgray-instawork: The navigator will, which is handled in a separate PR. Any screens generated in the `HvRouteInner` will create their own `onUpdate`

flochtililoch (approved):

PR #788, opened 2024-01-18
fix: reload not triggering a load

hgray-instawork:
Fixing an issue where reload behaviors didn't reload content.

The issue found was that the state set in hv-root did not cause the hv-screen to update. This meant the `componentDidUpdate` was never fired so the "needsLoad" value wasn't evaluated to cause a reload.

The fix is to remove the setting of "needsLoad" from within hv-root and replace it with an explicit call to load.

A separate task has been created to evaluate ways to remove the "needsLoad" from hv-screen and hv-route. https://app.asana.com/0/1204323441804469/1206385492929255/f

| Before | After |
| -- | -- |
| before | after |
| before-perms | after-perms |

Asana bugs fixed:
- https://app.asana.com/0/1204008699308084/1206367117739199/f
- https://app.asana.com/0/1204008699308084/1206367117739209/f

### DIFF
--- a/src/core/components/hv-root/index.tsx
+++ b/src/core/components/hv-root/index.tsx
@@ -134,14 +134,18 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
     }
 
     // Re-render the modifications
-    opts.onUpdateCallbacks?.setState({
-      doc: newRoot,
-      elementError: null,
-      error: null,
-      url,
-    });
-    // Force a reload regardless of state
-    opts.onUpdateCallbacks?.reload();
+    opts.onUpdateCallbacks?.setState(
+      () => {
+        return {
+          doc: newRoot,
+          elementError: null,
+          error: null,
+          url,
+        };
+      },
+      // Force a reload once state is set
+      opts.onUpdateCallbacks?.reload,
+    );
   };
 
   /**
--- a/src/core/components/hv-root/index.tsx
+++ b/src/core/components/hv-root/index.tsx
@@ -134,13 +134,14 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
     }
 
     // Re-render the modifications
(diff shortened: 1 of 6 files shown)


src/core/components/hv-route/index.tsx
```diff
@@ -472,6 +472,7 @@ function RouteFC(props: Types.FCProps) {
           getState: () => docContext.getState(),
           registerPreload: (id: number, el: Element) =>
             props.setPreload(id, el),
+          reload: () => {},
```
flochtililoch, line 475: can't we also hook the `load` function here?

src/core/components/hv-root/index.tsx
```diff
@@ -134,13 +134,14 @@ export default class Hyperview extends PureComponent<HvScreenProps.Props> {
...
     // Re-render the modifications
-    opts.onUpdateCallbacks?.setNeedsLoad();
     opts.onUpdateCallbacks?.setState({
       doc: newRoot,
       elementError: null,
       error: null,
       url,
     });
+    // Force a reload regardless of state
+    opts.onUpdateCallbacks?.reload();
```
flochtililoch, line 144: Now I wonder if this call should be made as a callback of the `setState`, to match more closely previous implementation?
```suggestion
    }, () => {
      // Force a reload regardless of state
      opts.onUpdateCallbacks?.reload();
    });
```

hgray-instawork: I'll have a look

hgray-instawork: Updated the setState to support the functional call with callback.

src/core/components/hv-route/index.tsx
hgray-instawork: Reworked the needsLoad to use a state which will trigger the inner class component reload.

flochtililoch (approved):

PR #789, opened 2024-01-25
feature: Adding support for blocking close/back navigation

hgray-instawork:
Re-working the support for back navigation control with a simplified approach.

- added new "back" trigger
- added new "BackBehaviorContext" to manage a behavior registry and store the current 'onUpdate' which will handle the behaviors
- modified hyper-ref to register/deregister its 'trigger=back' behaviors along with its 'onUpdate'
- wrapped all instances of 'hv-route' with a BackBehaviorContext and also wrapped the standalone hv-screen

Tested in iOS simulator for Pro and Partner app
- Pro app won't currently support the feature but its addition doesn't break anything
- Partner app was tested with 'back' behaviors in place and omitted and worked as expected

Verified the following test setups:
```XML
<view id="back-container">
  <behavior
    trigger="back"
    action="hide"
    target="back-container"
  />
  <behavior
    trigger="back"
    action="back"
  />
</view>
```

```XML
<view id="back-container">
  <behavior
    trigger="back"
    action="hide"
    target="back-container"
  />
  <behavior
    trigger="back"
    action="alert"
    alert:title="Test"
    alert:message="Click 'go back' to go back"
  >
    <alert:option alert:label="go back">
      <behavior
        action="back"
      />
    </alert:option>
  </behavior>
</view>
```

https://github.com/Instawork/hyperview/assets/127122858/caccf30b-d4f2-4a64-9874-bb8522f9c7dd

### DIFF
--- a/src/contexts/back-behaviors.tsx
+++ b/src/contexts/back-behaviors.tsx
@@ -9,23 +9,6 @@
 import React, { ReactNode, createContext, useRef, useState } from 'react';
 import { HvComponentOnUpdate } from 'hyperview/src/types';
 
-class BackBehaviorRegistry {
-  backBehaviorElements: Element[] = [];
-
-  add = (elements: Element[]): void => {
-    this.backBehaviorElements.push(...elements);
-  };
-
-  get = (): Element[] => this.backBehaviorElements;
-
-  remove = (elements: Element[]): void => {
-    this.backBehaviorElements = elements.reduce((acc, e) => {
-      const i = acc.indexOf(e);
-      return i > -1 ? [...acc.slice(0, i), ...acc.slice(i + 1)] : acc;
-    }, this.backBehaviorElements);
-  };
-}
-
 export type BackBehaviorContextProps = {
   add: (elements: Element[], onUpdate: HvComponentOnUpdate) => void;
   get: () => Element[];
@@ -41,22 +24,32 @@ export const BackBehaviorContext = createContext<BackBehaviorContextProps | null
   null,
 );
 
+export function removeElements(
+  registry: Element[],
+  remove: Element[],
+): Element[] {
+  return remove.reduce((acc, e) => {
+    const i = acc.indexOf(e);
+    return i > -1 ? [...acc.slice(0, i), ...acc.slice(i + 1)] : acc;
+  }, registry);
+}
+
 export function BackBehaviorProvider(props: { children: ReactNode }) {
-  const registry = useRef<BackBehaviorRegistry>(new BackBehaviorRegistry());
+  const registry = useRef<Element[]>([]);
   const [onUpdate, setOnUpdate] = useState<HvComponentOnUpdate>(() => null);
 
   const add = (elements: Element[], update: HvComponentOnUpdate): void => {
     if (elements.length === 0) {
       return;
     }
-    registry.current.add(elements);
+    registry.current.push(...elements);
     setOnUpdate(() => update);
   };
 
-  const get = (): Element[] => registry.current.get();
+  const get = (): Element[] => registry.current;
 
   const remove = (elements: Element[]): void => {
-    registry.current.remove(elements);
+    registry.current = removeElements(registry.current, elements);
   };
 
   return (
--- a/src/contexts/back-behaviors.tsx
+++ b/src/contexts/back-behaviors.tsx
@@ -1,0 +1,74 @@
+/**
+ * Copyright (c) Garuda Labs, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ */
+
+import React, { ReactNode, createContext, useRef, useState } from 'react';
+import { HvComponentOnUpdate } from 'hyperview/src/types';
+
+class BackBehaviorRegistry {
+  backBehaviorElements: Element[] = [];
+
+  add = (elements: Element[]): void => {
+    this.backBehaviorElements.push(...elements);
+  };
+
+  get = (): Element[] => this.backBehaviorElements;
+
+  remove = (elements: Element[]): void => {
+    this.backBehaviorElements = elements.reduce((acc, e) => {
+      const i = acc.indexOf(e);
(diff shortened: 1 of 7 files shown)


src/core/components/hv-route/index.tsx
```diff
@@ -577,12 +581,40 @@ export default function HvRoute(props: Types.Props) {
...
+            event.preventDefault();
+            elements.forEach(behaviorElement => {
+              const href = behaviorElement.getAttribute('href');
+              const action = behaviorElement.getAttribute('action');
+              console.log(
+                'onupdate: ',
+                href,
+                action,
+                backContext.onUpdate !== null,
+              );
```
flochtililoch, line 599: I think you forgot to remove this

src/core/components/hv-route/index.tsx
```diff
@@ -605,4 +643,14 @@ export default function HvRoute(props: Types.Props) {
   );
 }
 
+function HvRouteBehaviorWrapper(props: Types.Props) {
```
flochtililoch, line 646: do we need this additional component, or could the `<BackBehaviorProvider/>` be added to `HvRoute`?

src/core/hyper-ref/index.tsx
```diff
@@ -81,28 +84,49 @@ export default class HyperRef extends PureComponent<Props, State> {
...
+    // Register behavior elements for back triggers
+    this.addBackBehaviors();
   }
 
   componentDidUpdate(prevProps: Props) {
     if (prevProps.element === this.props.element) {
       return;
     }
 
+    // Deregister event listener for on-event triggers
```
flochtililoch, line 97:
```suggestion
    // Deregister event listener for back triggers
```

src/core/hyper-ref/index.tsx
```diff
@@ -81,28 +84,49 @@ export default class HyperRef extends PureComponent<Props, State> {
...
     this.updateStyle();
     this.triggerLoadBehaviors();
+    // Register behavior elements for back triggers
+    this.addBackBehaviors();
   }
 
   componentWillUnmount() {
     // Remove event listener for on-event triggers to avoid memory leaks
     Events.unsubscribe(this.onEventDispatch);
+    // Deregister event listener for on-event triggers
```
flochtililoch, line 110:
```suggestion
    // Deregister event listener for back triggers
```

src/contexts/back-behaviors.tsx
```diff
@@ -0,0 +1,74 @@
...
+/*
+ * Provides a registry of back behaviors to allow sharing between hv-screen and hv-route
+ * Additionally contains the onUpdate to use for the behaviors
+ */
+export const BackBehaviorContext = createContext<BackBehaviorContextProps | null>(
+  null,
+);
+
+export function BackBehaviorProvider(props: { children: ReactNode }) {
+  const registry = useRef<BackBehaviorRegistry>(new BackBehaviorRegistry());
```
flochtililoch, line 45: I feel the benefits of adding the BackBehaviorRegistry class over using a simple object here are diminished by the amount of boilerplate/repeated code. IMO the only piece of logic that could deserve isolation and unit tests is the `remove` method.

src/core/components/hv-route/index.tsx
hgray-instawork: It needs to be separate to allow the FC to access the context. I have updated this to rename it as "HvRoute" so that all instances are wrapped correctly.

src/core/components/hv-route/index.tsx
```diff
@@ -577,12 +581,34 @@ export default function HvRoute(props: Types.Props) {
...
+              const action = behaviorElement.getAttribute('action');
+              if (backContext.onUpdate) {
+                backContext.onUpdate(href, action, behaviorElement, {
+                  behaviorElement,
+                  showIndicatorId: behaviorElement.getAttribute(
+                    'show-during-load',
+                  ),
+                  targetId: behaviorElement.getAttribute('target'),
+                });
+              }
```
flochtililoch, line 602: minor: to avoid having to test for `backContext?.onUpdate` twice, you can deconstruct / assign new constants which will satisfy TS:
```suggestion
          const { get, onUpdate } = backContext || {};
          const elements: Element[] | undefined = get && get();
          if (elements?.length > 0 && onUpdate) {
            // Process the elements
            event.preventDefault();
            elements.forEach(behaviorElement => {
              const href = behaviorElement.getAttribute('href');
              const action = behaviorElement.getAttribute('action');
              onUpdate(href, action, behaviorElement, {
                behaviorElement,
                showIndicatorId: behaviorElement.getAttribute(
                  'show-during-load',
                ),
                targetId: behaviorElement.getAttribute('target'),
              });
```

src/core/components/hv-route/index.tsx
```diff
@@ -605,4 +637,14 @@ export default function HvRoute(props: Types.Props) {
   );
 }
 
+export default function HvRoute(props: Types.Props) {
+  return (
+    <BackBehaviorProvider>
+      <HvRouteFC
+        // eslint-disable-next-line react/jsx-props-no-spreading
+        {...props}
```
flochtililoch, line 645: As a general practice, I can't recommend props spreading, and the linting rule is actually here for a good reason - it's easy with refactors to have unwanted props sneak in and cause bugs.

flochtililoch (approved):
Looks good overall, added a couple code suggestions - should this go through QA before we merge?

src/contexts/back-behaviors.tsx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export function BackBehaviorProvider(props: { children: ReactNode }) {
+  const registry = useRef<Element[]>([]);
+  const [onUpdate, setOnUpdate] = useState<HvComponentOnUpdate>(() => null);
+
+  const add = (elements: Element[], update: HvComponentOnUpdate): void => {
+    if (elements.length === 0) {
+      return;
+    }
```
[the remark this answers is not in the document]
hgray-instawork: Discussed and leaving in to avoid re-render on state change when unnecessary

PR #790, opened 2024-01-26
chore: remove HvRoute prop spreading

hgray-instawork:
Removing the prop spreading between the functional component and the class component

Asana: https://app.asana.com/0/1204008699308084/1206433111684027/f

### DIFF
--- a/src/core/components/hv-route/index.tsx
+++ b/src/core/components/hv-route/index.tsx
@@ -628,9 +628,23 @@ function HvRouteFC(props: Types.Props) {
 
   return (
     <HvRouteInner
-      // eslint-disable-next-line react/jsx-props-no-spreading
-      {...{ ...props, ...navigationContext, ...navigatorMapContext }}
+      behaviors={navigationContext.behaviors}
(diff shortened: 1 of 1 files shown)


flochtililoch (approved):

src/core/components/hv-route/index.tsx
```diff
@@ -628,9 +628,23 @@ function HvRouteFC(props: Types.Props) {
 
   return (
     <HvRouteInner
-      // eslint-disable-next-line react/jsx-props-no-spreading
-      {...{ ...props, ...navigationContext, ...navigatorMapContext }}
```
flochtililoch, line 632: The order in which the final object might be changing - we used to have
`props`, on top of which we merged `navigationContext`, then `navigatorMapContext` (in this order).
Is there any possible overlap of props between these 3 objects, and if so, does this refactor accounts for it?

hgray-instawork: Confirmed 'url' is the only overlapping property and is handled properly

PR #799, opened 2024-02-02
chore: update demo

hgray-instawork:
Updating the demo to include example using `<navigator>` HXML source.

- Created a toggle in the header to switch between external navigator (legacy) and internally generated navigator using new `<navigator>` syntax
- Fixed an issue with the 'changed' file containing invalid attributes (can't have scrolling and safe enabled on the same view)
- Added 'back' example for behavior triggers (this will only work using the internal navigator)

Examples showing different functionality with 'legacy' and 'updated' navigation:
| Close action | Back trigger |
| -- | -- |
| close | back |

Asana: https://app.asana.com/0/1204008699308084/1206290261007827/f


flochtililoch (commented):
I think we should revise the approach to:
- separate out the code for internal and external navigator - most users will use one or the other, and having them both on the same file will make it harder to reason about.
- handle the toggling via a custom behavior - this will allow for the toggle UI to be rendered within Hyperview itself and also serve as a showcase for custom HV behaviors.

demo/App.tsx
```diff
@@ -6,10 +6,82 @@
...
-import React from 'react';
+import { NavigationContainer } from '@react-navigation/native';
 
-export default function App() {
-  return <Navigator />;
+const NAMESPACE_URI = 'https://instawork.com/hyperview-set-navigator-source';
+
+/**
+ * Define expected values
+ */
```
flochtililoch, line 27: Looks like this is coming from a copy/paste, probably not useful here
```suggestion
```

demo/App.tsx
```diff
@@ -6,10 +6,82 @@
...
+enum NavigationSource {
+  HXML = 'hxml',
+  EXTERNAL = 'external',
```
flochtililoch, line 30: Maybe instead of `html`, `hyperview` would be more representative. Also sorting alphabetically:
```suggestion
  EXTERNAL = 'external',
  HYPERVIEW = 'hyperview',
```

demo/App.tsx
```diff
@@ -6,10 +6,82 @@
...
 }
+
+export default () => {
+  const [useHXML, setUseXML] = useState<boolean>(false);
```
flochtililoch, line 34: I think these could also be named after what they represent, and we can reuse the enum
```suggestion
  const [navigationSource, setNavigationSource] = useState<NavigationSource>(NavigationSource.EXTERNAL);
```

demo/App.tsx
```diff
@@ -6,10 +6,82 @@
...
+
+  /**
+   * Create a custom behavior to toggle using HXML to define the navigator hierarchy
+   */
+  const setNavigatorSrcBehavior = {
```
flochtililoch, line 39: nit: not abbreviating / changing terminology to keep naming consistent, and updating comment
```suggestion
   * Create a custom behavior to toggle between using Hyperview or externally provided navigator hierarchy
   */
  const setNavigationSourceBehavior = {
```

demo/App.tsx
```diff
@@ -6,10 +6,82 @@
...
+    action: 'set-navigator-source',
```
flochtililoch, line 40:
```suggestion
    action: 'set-navigation-source',
```

demo/App.tsx
```diff
@@ -6,10 +6,82 @@
...
+    callback: (element: Element) => {
+      const source = element.getAttributeNS(NAMESPACE_URI, 'source');
+      if (source === NavigationSource.HXML) {
+        setUseXML(true);
+      } else if (source === NavigationSource.EXTERNAL) {
+        setUseXML(false);
+      }
```
flochtililoch, line 47:
```suggestion
      setNavigationSource(source as NavigationSource);
```

demo/App.tsx
hgray-instawork: Changed name of enum instead of action

demo/App.tsx
```diff
@@ -6,10 +6,77 @@
...
-  return <Navigator />;
+const NAMESPACE_URI = 'https://instawork.com/hyperview-set-navigator-source';
+
+enum NavigatorSource {
+  EXTERNAL = 'external',
+  HYPERVIEW = 'hyperview',
 }
+
+export default () => {
+  const [navigationSource, setNavigationSource] = useState<NavigatorSource>(
```
flochtililoch, line 31:
```suggestion
  const [navigatorSource, setNavigatorSource] = useState<NavigatorSource>(
```

demo/App.tsx
```diff
@@ -6,10 +6,77 @@
...
+  );
+
+  /**
+   * Create a custom behavior to toggle using HXML to define the navigator hierarchy
+   */
+  const setNavigatorSourceBehavior = {
+    action: 'set-navigator-source',
+    callback: (element: Element) => {
+      const source = element.getAttributeNS(NAMESPACE_URI, 'source');
+      setNavigationSource(source as NavigatorSource);
```
flochtililoch, line 42:
```suggestion
      setNavigatorSource(source as NavigatorSource);
```

demo/App.tsx
```diff
@@ -6,10 +6,77 @@
...
+              flex: 1,
+
+              // Padding to handle safe area
+              paddingBottom: insets?.bottom,
+              paddingLeft: insets?.left,
+              paddingRight: insets?.right,
+            }}
+          >
+            <NavigationContainer>
+              {navigationSource == NavigatorSource.HYPERVIEW ? (
```
flochtililoch, line 64:
```suggestion
              {navigatorSource == NavigatorSource.HYPERVIEW ? (
```

demo/App.tsx
```diff
@@ -6,10 +6,77 @@
...
+} from 'react-native-safe-area-context';
+import HyperviewScreen from './src/HyperviewScreen';
+import { View } from 'react-native';
 import Navigator from './src/Navigator';
-import React from 'react';
+import { NavigationContainer } from '@react-navigation/native';
 
-export default function App() {
-  return <Navigator />;
+const NAMESPACE_URI = 'https://instawork.com/hyperview-set-navigator-source';
```
flochtililoch, line 23: let's use a namespace that's more generic and that isn't coupled to Instawork
```suggestion
const NAMESPACE_URI = 'https://hyperview.org/demo';
```

examples/_includes/base.xml.njk
```diff
@@ -3,14 +3,30 @@
...
     <body style="Body" safe-area="true">
       <header style="Header">
         {% if not hide_back_button %}
           <text action="back" href="#" style="Header__Back">Back</text>
         {% endif %}
         <text style="Header__Title">{{ title }}</text>
+        {% if show_hxml_toggle %}
+          <view style="Header__Switch">
+            <text style="Switch__Label">HXML Navigator</text>
+            <switch id="navigator-switch" xmlns:set-navigator-source="https://instawork.com/hyperview-set-navigator-source" trigger="change" action="set-navigator-source" set-navigator-source:source={% if hxml_toggle_enabled %} "external" {% else %} "hyperview" {% endif %} value={% if hxml_toggle_enabled %} "on" {% else %} "off" {% endif %}/>
```
flochtililoch, line 27: Adjusting NS / breaking up line / sorting attributes:
```suggestion
            <switch
              id="navigator-switch"
              trigger="change"
              action="set-navigator-source"
              xmlns:demo="https://hyperview.org/demo"
              demo:source="{% if hxml_toggle_enabled %}external{% else %}hyperview{% endif %}"
              value="{% if hxml_toggle_enabled %}on{% else %}off{% endif %}"
            />
```

additional question: could we rename `hxml_toggle_enabled` to `hv_navigator_enabled`, and rename the label to "Hyperview Navigator"?

examples/index_hxml.xml.njk
```diff
@@ -0,0 +1,14 @@
+---
+title: "Hyperview"
+permalink: "index_hxml.xml"
+list_tag: "root"
+hide_back_button: true
+list_href: "list.xml"
+show_hxml_toggle: true
+hxml_toggle_enabled: true
```
flochtililoch, line 8: minor/nit: I wonder if it's possible to use a single flag, and test for `!undefined` for "visibility" vs `== true/false` for "on/off" state. Also worth renaming to match the "navigator" terminology here.

examples/index_navigator.xml.njk
```diff
@@ -0,0 +1,11 @@
+---
+title: "Hyperview navigation"
```
flochtililoch, line 2:
```suggestion
title: "Hyperview navigator"
```

examples/_includes/base.xml.njk
hgray-instawork: Fixed except the spacing as discussed

flochtililoch (approved):
LGTM

PR #805, opened 2024-02-08
fix: 'close' behavior should close a modal

hgray-instawork:
- Identify routes by including the presentation in the params so we don't have to rely on naming conventions
- If a targetId is assigned, the existing logic is used without change
- Look through previous routes in the stack, if one is found which is modal
  - update the action to "navigate" and pass the name of the route which precedes the modal
- Continue up the hierarchy to find a navigator which is showing a modal route

| HV Demo | Single-stack | Multi-stack |
| -- | -- | -- |
| demo | single | multi |

### DIFF
--- a/src/services/navigation/index.ts
+++ b/src/services/navigation/index.ts
@@ -14,7 +14,6 @@ import type {
   NavAction,
   NavigationProps,
 } from 'hyperview/src/types';
-import { ID_CARD, ID_MODAL } from 'hyperview/src/services/navigator/types';
 import { NAV_ACTIONS } from 'hyperview/src/types';
 import { getFormData } from 'hyperview/src/services';
 
@@ -110,11 +109,9 @@ export default class Navigation {
         }
       }
     }
-    const presentation = action === NAV_ACTIONS.NEW ? ID_MODAL : ID_CARD;
     const routeParams = {
       delay,
       preloadScreen,
-      presentation,
       targetId,
       url,
     } as const;
--- a/src/services/navigation/index.ts
+++ b/src/services/navigation/index.ts
@@ -14,6 +14,7 @@ import type {
   NavAction,
   NavigationProps,
 } from 'hyperview/src/types';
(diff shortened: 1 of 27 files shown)


src/services/navigator/index.ts
```diff
@@ -58,7 +58,7 @@ export class Navigator {
   sendRequest = (action: NavAction, routeParams?: NavigationRouteParams) => {
     const navAction: NavAction = Helpers.getNavAction(action, routeParams);
 
-    const [navigation, routeId, params] = Helpers.buildRequest(
+    const [requestAction, navigation, routeId, params] = Helpers.buildRequest(
```
flochtililoch, line 61: there's a bit of tangling we can remove here:
- `navAction` is now only used as an entry param of `buildRequest`, so `Helpers.buildRequest` interface can be adjusted to receive `action` instead of `navAction`, and delegate to `Helpers.getNavAction` internally.
- instead of `requestAction`, we can preserve the name `navAction` which will help reduce the surface of the change

src/services/navigation/index.ts
```diff
@@ -109,8 +109,12 @@ export default class Navigation {
...
       }
     }
-
-    const routeParams = { delay, preloadScreen, targetId, url } as const;
+    const routeParams = {
+      delay,
+      preloadScreen,
+      targetId,
+      url,
+    } as const;
```
flochtililoch, line 117: this change can be reverted

src/services/navigator/helpers.test.ts
```diff
@@ -663,15 +663,19 @@ describe('buildRequest', () => {
...
       describe(`action:${action}`, () => {
         const params = { url: 'url' };
-        it('should ignore object without params', () => {
-          expect(buildRequest(undefined, action, undefined)).toEqual([
-            undefined,
-            '',
-            {},
-          ]);
-        });
+        if (action !== NAV_ACTIONS.CLOSE) {
```
flochtililoch, line 666: is there another kind of assertion we should be doing for close action?

flochtililoch (approved):

PR #808, opened 2024-02-12
chore: update prettier configs and apply to docs folder

hgray-instawork:
- Updated .editorconfig and .vscode/settings.json with line length settings
- Ignore generated demo/.expo/README.md from prettier check
- Re-saved all /doc files to apply formatting
- No changes were made to the files

Asana: https://app.asana.com/0/1204008699308084/1206525468973413/f


flochtililoch (approved):
I think this is good to go, as my other PR #812 only fixes indentation so there shouldn't be any overlaps. However, after digging a bit, there appears to be a way to configure the editor, so that lines have a max-length - see https://github.com/eseom/nunjucks-template/issues/24.
Until we can apply this formatting on CI directly, it might be worth configuring [here](https://github.com/Instawork/hyperview/blob/master/.editorconfig) and set the same value as [Prettier's default (80)](https://prettier.io/docs/en/options.html#print-width)

examples/ui_elements/text/escaped.xml.njk
```diff
@@ -13,4 +13,4 @@ tags: text
     <![CDATA[Escaped with <![CDATA[]>:
     <hello>
       <world>]]></text>
-{% endblock %}
+    {% endblock %}
```
hgray-instawork, line 16: This seems to be a result of the CDATA block. I wasn't able to get it to stop indenting it with the current content.

flochtililoch (approved):

PR #811, opened 2024-02-13
fix: dom navstate sync

hgray-instawork:
Fixing two bugs:
1. The recursive tab issue https://app.asana.com/0/1204008699308084/1206481489666893/f
2. The issue where multiple modals won't open and some pushes require two clicks https://app.asana.com/0/1204008699308084/1206525468973401/f

Both of these issues were caused by the DOM and the react-navigation state being out of sync:
- Tab selection occurs, or route is added/removed
- DOM was updated but not cloned
- Something changes state
- Deep link logic (for tab or stack) tries to reconcile current state against DOM
- DOM being out of sync shows the old state
  - Tab reverts to previous selection
  - Stack closed the just-opened route

| Before | After |
| -- | -- |
| before-tab | after-tab |
| before-stack | after-stack |


flochtililoch (approved):
🎉

PR #812, opened 2024-02-13
chore: indent Nunjucks files

flochtililoch:
Update CI to use [djhtml](https://github.com/rtts/djhtml) to indent our Nunjucks files.


scripts/format-xml.sh
hgray-instawork, line 1: Don't we still want this for the files which are not nunjucks (see examples/behaviors). The pattern below would have to be modified to ignore the *.xml.njk files.

hgray-instawork (approved):

Issue #816, opened 2024-02-14
Scroll indicator is shown for `<view trigger="refresh" scroll="true" shows-scroll-indicator="false"/>`


Charles-Johnson:
`trigger="refresh"` wraps a `ScrollView` without taking into account the `shows-scroll-indicator` attribute

I think this code in `src/core/hyper-ref/index.tsx` needs changing to fix this bug

```tsx
ScrollableView = ({ children }: { children: JSX.Element }): JSX.Elem
  const behaviors = this.getBehaviorElements(TRIGGERS.REFRESH);
  if (!behaviors.length) {
    return children;
  }
  const refreshHandlers = behaviors.map(behaviorElement =>
    Behaviors.createActionHandler(behaviorElement, this.props.onUpda
  );
  const onRefresh = () => refreshHandlers.forEach(h => h(this.props.
  const refreshControl = React.createElement(RefreshControl, {
    onRefresh,
    refreshing: this.state.refreshing,
  });
  return React.createElement(
    ScrollView,
    { refreshControl, style: this.getStyle() },
    children,
  );
};
```

Charles-Johnson:
I think I can use `this.props.element.getAttribute('shows-scroll-indicator')` to determine the `showsVerticalScrollIndicator` prop for the `ScrollView`

Charles-Johnson:
I tested these changes:

Charles-Johnson:
I tested these changes:
```tsx
```

Charles-Johnson:
I tested changing the above code to this which fixed the issue:
```tsx
ScrollableView = ({ children }: { children: JSX.Element }): JSX.Element => {
  const behaviors = this.getBehaviorElements(TRIGGERS.REFRESH);
  if (!behaviors.length) {
    return children;
  }
  const refreshHandlers = behaviors.map(behaviorElement =>
    Behaviors.createActionHandler(behaviorElement, this.props.onUpdate),
  );
  const onRefresh = () => refreshHandlers.forEach(h => h(this.props.element));
  const refreshControl = React.createElement(RefreshControl, {
    onRefresh,
    refreshing: this.state.refreshing,
  });
  return React.createElement(
    ScrollView,
    {
      refreshControl,
      showsVerticalScrollIndicator:
        this.props.element.getAttribute('shows-scroll-indicator') !== 'false',
      style: this.getStyle(),
    },
    children,
  );
};
```

flochtililoch:
[...]

PR #819, opened 2024-02-16
Prevent `trigger="refresh"` adding scroll indicator if `shows-scroll-indicator="false"`

Charles-Johnson:
This resolves issue #816


flochtililoch (approved):
LGTM, thanks for the fix!

PR #831, opened 2024-02-26
chore: prototype custom tab bar

hgray-instawork:
Implemented as a new view type `<footer>` which lives outside the `<body>` element of a `<doc>`
As each screen is first loaded, it passes its `<footer>` element to a context which can then provide it to the `<navigator>` for display. Currently there is no state handling.

Pros:
- Styles and behaviors are handled as part of the standard `<hvscreen>` component

Cons:
- each tab's screen will have to contain the same `<footer>` element
- have to wait for at least one screen to load before tab bar is shown

https://github.com/Instawork/hyperview/assets/127122858/b4797651-acce-4534-9ab3-b5005a6e14dd

Asana: https://app.asana.com/0/1204008699308084/1206459525167059/f


examples/navigator/custom_tabbar/tab_1.xml.njk
```diff
@@ -0,0 +1,42 @@
...
+{% extends 'base.xml.njk' %}
+{% block container %}
+  <view scroll="true">
+    <text style="Description">This is the content for tab 1.</text>
+  </view>
+{% endblock %}
+
+{% block footer %}
+  <footer>
+    <view id="footer" style="Footer">
```
flochtililoch, line 15: do we need a separate nested container, or could `footer` hold the props `id` and `style`?

src/core/components/hv-navigator/index.tsx
```diff
@@ -26,7 +30,7 @@ import { getFirstChildTag } from 'hyperview/src/services/dom/helpers';
 /**
  * Flag to show the navigator UIs
  */
-const SHOW_NAVIGATION_UI = false;
+const SHOW_NAVIGATION_UI = true;
```
flochtililoch, line 33: I think this override and commented use of `SHOW_NAVIGATION_UI` need to be reverted

examples/navigator/custom_tabbar/index.xml.njk
```diff
@@ -0,0 +1,17 @@
...
+title: "Custom Tabbar"
+permalink: "/navigator/custom_tabbar/index.xml"
+tags: navigator
+---
+
+{# Example tab navigator inside a stack navigator #}
+<doc xmlns="https://hyperview.org/hyperview">
+  <navigator id="top-stack" type="stack">
+    <nav-route id="tabs-route" href="">
+      <navigator id="simple-tab-navigator" type="tab" tabbar-href="/navigator/custom_tabbar/tabbar.xml">
```
flochtililoch, line 11: it looks like `tabbar-href`, and the file it references, are not used?

hgray-instawork:
Closing in favor of @flochtililoch's work in PRs like https://github.com/Instawork/hyperview/pull/962

PR #838, opened 2024-03-19
fix(navigation): routes can only close themselves

hgray-instawork:
**Issue:**
When cards close themselves behind a modal, the modal was being closed too. This is most easily seen in the Partner app in MDS Viewer/Navigation/Push then modal where the final "done" button opens and immediately closes the modal.

**Resolution:**
When a "close" or "back" action is triggered, only the route which originated the behavior will be removed from state.
If no "source" is provided for the action, default back implementation is used

| Before | After |
| -- | -- |
| before | after |

Asana: https://app.asana.com/0/1204008699308084/1206868462427486/f


hgray-instawork:
Closing in favor of other solutions

PR #850, opened 2024-04-29
chore: create script to update readme from package dependencies

hgray-instawork:
- Added a shell script to automatically update parts of the readme with the dependencies and peer dependencies.
  - Requires [jq](https://jqlang.github.io/jq/) to run but will exit early and not block the remaining script if not included.
- Updated pre-version to run this script
- Updated post-version to include readme in git commit

| package.json | readme.md |
| --- | --- |
| package | readme |

Asana: https://app.asana.com/0/1204008699308084/1207182290354073/f

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,9 +26,9 @@
   "scripts": {
     "sync": "babel-node ./scripts/sync",
     "generate": "babel-node ./scripts/generate",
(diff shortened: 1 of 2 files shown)


hgray-instawork:
We decided this is not worth implementing

PR #859, opened 2024-05-13
chore: add logging support

hgray-instawork:
- Added new types for a standard log payload
- Added ability to pass an external logger into Hyperview for use in legacy or navigation variations
- Updated demo app to illustrate logging options

Asana: https://app.asana.com/0/1204008699308084/1207303445405256/f

log

### DIFF
--- a/src/services/logging/index.ts
+++ b/src/services/logging/index.ts
@@ -1,0 +1,32 @@
+import type { Log } from 'hyperview/src/types';
+
+export default class Logging {
+  logger?: Log;
+
+  prefix?: string;
+
+  constructor(logger?: Log, prefix?: string) {
+    this.logger = logger;
+    this.prefix = prefix;
+  }
+
+  log = (
+    message: string,
+    severity: number,
+    data?: unknown,
+    source?: string,
+    timestamp?: Date,
+  ) => {
+    if (!this.logger) {
+      return;
+    }
+    this.logger({
+      data,
+      message,
+      prefix: this.prefix,
+      severity,
+      source,
+      timestamp: timestamp || new Date(),
+    });
+  };
+}
--- a/src/types.ts
+++ b/src/types.ts
@@ -431,3 +431,22 @@ export type Reload = (
   optHref: DOMString | null | undefined,
   opts: HvComponentOptions,
 ) => void;
+
+export type LogPayload = {
+  data: unknown;
+  message: string;
+  prefix?: string;
+  severity: number;
+  source?: string;
+  timestamp: Date;
+};
+
+export const LOG_SEVERITY = {
+  ERROR: 0,
+  INFO: 30,
+  LOG: 20,
+  WARN: 10,
+};
+Object.freeze(LOG_SEVERITY);
+
+export type Log = (payload: LogPayload) => void;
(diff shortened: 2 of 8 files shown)


src/services/logging/index.ts
```diff
@@ -0,0 +1,32 @@
+import type { Log } from 'hyperview/src/types';
+
+export default class Logging {
```
flochtililoch, line 3: this doesn't seem to be in use?

src/types.ts
```diff
@@ -431,3 +431,22 @@ export type Reload = (
   optHref: DOMString | null | undefined,
   opts: HvComponentOptions,
 ) => void;
+
+export type LogPayload = {
```
flochtililoch, line 435: I think that `data`, `prefix`, `source` and `timestamp` are elements of the log message that can remain discreet in the logger implementation, so don't need to be part of the signature exposed by Hyperview

src/types.ts
```diff
@@ -431,3 +431,22 @@ export type Reload = (
...
+  timestamp: Date;
+};
+
+export const LOG_SEVERITY = {
+  ERROR: 0,
+  INFO: 30,
+  LOG: 20,
+  WARN: 10,
+};
+Object.freeze(LOG_SEVERITY);
```
flochtililoch, line 450: Are the values set as severity used for any purpose? If not we may make this a string enum. Did you also try to use the ["enum" syntax](https://www.typescriptlang.org/docs/handbook/enums.html#numeric-enums) to declare this?

src/services/logging/index.ts
hgray-instawork: Not yet. I have some local changes which tested it out. I'll create separate stories to implement logging within Hyperview.

src/types.ts
hgray-instawork: Numeric vs string: you can't filter on "less than or equal to LOG" if the values are strings. Enum: I can try that out.

src/services/logging/index.ts
flochtililoch: I don't think we want Hyperview to implement a logger, more so use one if injected via props. This probably should remain outside of Hyperview core.

src/types.ts
hgray-instawork: You're suggesting we just export "any" and it's up to the logger to interpret the type? I thought it would be easier to be explicit.

src/types.ts
flochtililoch: I haven't seen this being used - but since there's a finite number of options, explicitly testing for the value might lead to same result?

src/services/logging/index.ts
hgray-instawork: This is just a helper within Hyperview to properly format and structure the data, auto-inject the current date (if none provided). The actual logger exists outside Hyperview.

src/types.ts
hgray-instawork: The spacing in-between each value was intended to provide sub-level severity. Example: log("checking something", 19). This would allow us to still see it when we do a filter:

if (severity <== LOG_SEVERITY.LOG){
  // log
}

But would also allow specific filtering:
if (severity === 19){
  // log
}

hgray-instawork:
Abandoning in favor of another approach

PR #864, opened 2024-05-15
chore: add logging support

hgray-instawork:
Adding the ability to provide external logging. Using the signature of `console`

Asana: https://app.asana.com/0/1204008699308084/1207303445405256/f


flochtililoch (commented):
To tie things up with this PR, I think it would make sense to:
- set a default value for the prop `logger` of `hv-root`, to `console`
- replace any calls to `console.*` in `src/` with that logger

I think this will surface some unanswered questions around how to pass the logger around in service function calls, where a component instance isn't part of the interface - we probably wouldn't want to pass the logger as an argument every time logging is needed, which might mean we should provide the logger in a static way to Hyperview.

src/core/logging/index.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export default {
+  initialize,
+  log: (m?: any, ...p: any[]) => logger.log(m, p),
+  // eslint-disable-next-line sort-keys
+  info: (m?: any, ...p: any[]) => logger.info(m, p),
+  warn: (m?: any, ...p: any[]) => logger.warn(m, p),
+  // eslint-disable-next-line sort-keys
+  error: (m?: any, ...p: any[]) => logger.error(m, p),
+};
```
flochtililoch, line 20: I think you might be able to instead export these two and get rid of the default export, which also removes the need for disabling the "any" TS rule:

```suggestion
let logger: Logger = console;

export default logger;

export function initialize(loggerInstance: Logger | undefined): void {
  if (loggerInstance) {
    logger = loggerInstance;
  }
}
```

src/core/logging/index.ts
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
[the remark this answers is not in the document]
hgray-instawork: Moved to /services

src/core/logging/index.ts
hgray-instawork: Had to do a slightly different approach

src/types.ts
```diff
@@ -431,3 +431,13 @@ export type Reload = (
   optHref: DOMString | null | undefined,
   opts: HvComponentOptions,
 ) => void;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type LogFunction = (message?: any, ...optionalParams: any[]) => void;
+
+export type Logger = {
```
flochtililoch, line 438: this might make more sense to be defined within `services/logging`? If we really need it outside of Hyperview, it can be re-exported from this file

flochtililoch (approved):
one last minor comment, looks good otherwise

PR #867, opened 2024-05-16
feat(web-view): add `allows-inline-media-playback` prop

sourabhdpunja:
Add allows-inline-media-playback prop to hyperview webview

### DIFF
--- a/src/components/hv-web-view/index.tsx
+++ b/src/components/hv-web-view/index.tsx
@@ -38,7 +38,7 @@ export default class HvWebView extends PureComponent<HvComponentProps> {
       this.props.options,
     );
     const color = props['activity-indicator-color'] || '#8d9494';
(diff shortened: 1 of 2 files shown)


src/components/hv-web-view/index.tsx
```diff
@@ -38,10 +38,12 @@ export default class HvWebView extends PureComponent<HvComponentProps> {
       this.props.options,
     );
     const color = props['activity-indicator-color'] || '#8d9494';
+    const allowsInlineMediaPlayback = props['allows-inline-media-playback'];
```
flochtililoch, line 41: I think we need to cast this to a boolean and only set when attribute exist:
```suggestion
    const allowsInlineMediaPlayback = props['allows-inline-media-playback'] ? props['allows-inline-media-playback'] === 'true' : undefined;
```

flochtililoch (approved):

PR #871, opened 2024-06-03
fix (error component): error component is covering tabor

hgray-instawork:
The internal `load-element-error` component was rendering at the bottom of the HVScreen. In applications which use ReactNavigation's tab bars, the bottom of the screen was above the tab bar. When tabs are rendered within the screen, the component floated part way up the tab bar (see 'original' below).

By moving the error to the top of the screen and pushing content down, we retain safe area boundaries and avoid floating over existing content.

| Original | Updated |
| -- | -- |
| shifts-before | updated |

The following show a few other examples of how the change affects screen appearance.

| Connected | Disconnected |
| -- | -- |
| refer-before | refer-error
| messages-before | messages-error |

Also tested in the demo using the default component and a custom component

| Default | Custom |
| -- | -- |
| default | custom |

Asana: https://app.asana.com/0/1204008699308084/1207426043394013/f

### DIFF
--- a/src/core/components/hv-screen/index.js
+++ b/src/core/components/hv-screen/index.js
@@ -281,13 +281,13 @@ export default class HvScreen extends React.Component {
         }}
       >
         <Contexts.DateFormatContext.Provider value={this.props.formatDate}>
-          {screenElement}
           {elementErrorComponent
             ? React.createElement(elementErrorComponent, {
                 error: this.state.elementError,
                 onPressReload: () => this.reload(),
               })
             : null}
+          {screenElement}
         </Contexts.DateFormatContext.Provider>
       </Contexts.DocContext.Provider>
     );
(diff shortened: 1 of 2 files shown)


adamstep (approved):

palaparthi (approved):

flochtililoch (commented):
it looks like we have the ability to specify the load error component from Hyperview's public API (via `elementErrorComponent` prop` - I wonder if a similar fix could be achieved directly in our apps, without the need to change Hyperview

hgray-instawork:
> it looks like we have the ability to specify the load error component from Hyperview's public API (via `elementErrorComponent` prop` - I wonder if a similar fix could be achieved directly in our apps, without the need to change Hyperview

That is how I did the last screen (THIS IS A CUSTOM ERROR COMPONENT...). I still thought it better to universally move the error component above the content area to avoid the tab bar area altogether.

PR #872, opened 2024-06-04
fix(webview): Hide webview loader when page renders before the loading of subresources

rohanbayya:
[Asana Ticket](https://app.asana.com/0/491474088351836/1207417081499074/f)

### Summary
1. WebView might render the document before other sub-resources have finished loading
2. This forces HvWebView to show a loading spinner rendering the WebView non-interactive.
3. Adding a custom javascript snipped to close the spinner upon initial render fixes this problem
4. Added Custom handler to capture document load event

### Screenshots
| Before | After |
| -- | -- |
| before | after |

### To-do

### DIFF
--- a/schema/core.xsd
+++ b/schema/core.xsd
@@ -1199,10 +1199,12 @@
       <xs:attribute name="allows-inline-media-playback" type="xs:boolean" />
       <xs:attribute name="injected-java-script" type="xs:string" />
       <xs:attribute name="show-loading-indicator">
-        <xs:restriction base="xs:string">
-          <xs:enumeration value="all"/>
-          <xs:enumeration value="document-only"/>
-        </xs:restriction>
+        <xs:simpleType>
+          <xs:restriction base="xs:string">
+            <xs:enumeration value="all"/>
+            <xs:enumeration value="document-only"/>
+          </xs:restriction>
+        </xs:simpleType>
       </xs:attribute>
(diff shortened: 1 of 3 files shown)


flochtililoch (commented):
Good progress on getting the code change ready. I did note that you had move the JS snippet into the component's code as I suggested, and the code change is concise and clean - nice!

I did think about some improvements we should consider before merging, see below.

I think we might need a clearer name for the new attribute, I'm thinking `show-load-indicator`, that could accept string values `all` (which would be default when not specified) and `document-only`. This gives us the opportunity to extend this feature further in the future, and perhaps support `never` or `main-frame` etc (looking at other hooks for inject-javascript).

A couple other notes about the implementation:
- Since we'll only have for now the handling for `document-only`, I think the mapping in `behaviorMapping` might be overkill for now, so you may just explicitly write in the render method a condition that test the attribute for this value (`document-only`), and appends the JS snippet to the `injectedJavascript` variable.
- The message to post might be made a bit more specific than `stopLoader` - perhaps we can use a namespace to better indicates this is an event "internal" to the Hyperview web view component, I'm thinking the string `hv-wev-view:render-loading:false`.

Let me know if these make sense or if you have any question! Thanks

flochtililoch (approved):
Looks good! As discussed, let's update the doc and examples In a follow up PR.

PR #879, opened 2024-06-11
fix(error): improve error component placement

hgray-instawork:
- Moved the component to the top
- Use top padding to provide safe area for Android
- Add a close button to dismiss the component

Android shows the close action, iOS shows the reload action.

| Android | iOS |
| -- | -- |
| android | iOS |

Asana: https://app.asana.com/0/1204008699308084/1207426043394013/f

### DIFF
--- a/src/core/components/load-element-error/index.tsx
+++ b/src/core/components/load-element-error/index.tsx
@@ -19,11 +19,6 @@ const LoadElementError = (props: Props) => {
 
   return (
     <SafeAreaView style={styles.safeArea}>
-      <TouchableOpacity onPress={props.onPressClose}>
-        <View style={styles.container}>
-          <Text style={styles.closeButton}>X</Text>
-        </View>
-      </TouchableOpacity>
       <TouchableOpacity onPress={props.onPressReload}>
         <View style={styles.container}>
           <View style={styles.textWrapper}>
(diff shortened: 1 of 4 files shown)


hgray-instawork:
@palaparthi feel free to tweak the styles as you see fit

flochtililoch (commented):
The rendered UI isn't great with the fake X button, I would suggest here to just tweak the string to read "TypeError: Network request failed. Reload or hide.", where both `Reload` and `hide` are tappable.

hgray-instawork:
Tweaked layout to bring action items closer together.

| DEV | PROD |
| -- | -- |
| dev | prod |

flochtililoch (approved):

src/core/components/load-element-error/index.tsx
```diff
@@ -19,16 +19,24 @@ const LoadElementError = (props: Props) => {
...
-      <TouchableOpacity onPress={props.onPressReload}>
-        <View style={styles.container}>
-          <View style={styles.textWrapper}>
-            <Text style={styles.title}>{getError()}</Text>
-          </View>
+      <View style={styles.container}>
+        <View style={styles.textWrapper}>
+          <Text style={styles.title}>{getError()}</Text>
+        </View>
+        <TouchableOpacity onPress={props.onPressReload}>
```
palaparthi, line 26: Maybe we could make both the error message and the reload button touchable since the earlier implementation was doing that and a better experience for the user as they don't need to target the button accurately while trying to reload.

palaparthi (approved):

src/core/components/load-element-error/index.tsx
hgray-instawork: I merged as is to get it shipped. I'm happy to go over updates with you if you'd like. One concern with having the touch area go beyond the verbs was when the text wraps to multiple I wanted to make sure it didn't overlap with the other button.

PR #882, opened 2024-06-13
Refactor ComponentRegistry into self-contained class

adamstep:
Previously, `ComponentRegistry` was a type that mapped namespace/local names to component classes. This object was passed around and used for rendering to map an XML element to a React component. It was becoming a bit cumbersome when we introduced a second form component registry used to collect form data. The functionality was scattered across helpers in different files, and we had to maintain these two registries. I was also running into issues when I wanted to expose the form registry in the public Hyperview API so custom components can access form data.

This refactor addresses those issues by creating a new `ComponentRegistry` class. This class handles lookups of components, with support for aliases. It also encapsulates the logic of pulling out form data. With this new class, we only need one registry object that can be passed around and handle component lookups and generating form data.

To-Do
- [x] Do we need to be backwards compatible? The component registry was exposed to custom components as `props.options.componentRegistry`, but it was never publicly documented. (Fixed in https://github.com/Instawork/mobile/pull/6186)


src/services/render/index.tsx
```diff
@@ -71,14 +72,14 @@ export const renderElement = (
...
-      options.componentRegistry[element.namespaceURI][element.localName]
-    ) {
-      const Component =
-        options.componentRegistry[element.namespaceURI][element.localName];
+    const Component: HvComponent | undefined = options.componentRegistry
+      ? options.componentRegistry.getComponent(
+          element.namespaceURI,
+          element.localName,
+        )
+      : undefined;
```
flochtililoch, line 80: nit: use optional chaining syntax
```suggestion
    const Component: HvComponent | undefined = options.componentRegistry?.getComponent(
        element.namespaceURI,
        element.localName,
      );
```

src/services/components/index.ts
```diff
@@ -32,27 +32,76 @@ const HYPERVIEW_COMPONENTS = [
...
-  );
-
-export const getFormRegistry = (
-  components: HvComponent[] = [],
-): ComponentRegistry =>
-  [...HYPERVIEW_COMPONENTS, ...components]
-    .filter(c => Object.prototype.hasOwnProperty.call(c, 'getFormInputValues'))
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    .reduce<Record<string, any>>(reducer, {});
+export class ComponentRegistry {
```
flochtililoch, line 35: nit: since the module name already convey the namespace, it might simplify usage throughout to rename this class to `Registry`.

flochtililoch (approved):
change looks good, noted a couple minor nit.

PR #884, opened 2024-06-17
feat:added tab navigator for home screen and template for examples

rohanbayya:
Asana ticket: https://app.asana.com/0/1204008699308084/1207532170446814/f

1. Added base template screen for examples to extend from
2. Tab navigation for home screen

<image src="https://github.com/Instawork/hyperview/assets/171086373/4c1fbf85-245b-418f-9931-e15ad5993624" width="400">
<image src="https://github.com/Instawork/hyperview/assets/171086373/ecc63ac8-fdb9-4b3b-b7ab-4bb3734b429e" width="400">
<image src="https://github.com/Instawork/hyperview/assets/171086373/0a710dbe-2830-4567-bfdc-6afbcf679c66" width="400">


examples/_includes/list.xml.njk
hgray-instawork, line 1: Is this needed?

examples/_includes/loading-screen.xml.njk
hgray-instawork, line 1: Should this extend `base` so we don't need to duplicate the header and style import?

rohanbayya: Extending base would create a circular dependency. instead extracting out header into another file.

examples/_includes/header.xml.njk
```diff
@@ -0,0 +1,11 @@
+<header style="header">
+  {% if hv_button_behavior == 'close' %}
+    <text action="close" href="#" style="header-btn">Close</text>
+  {% elif hv_button_behavior == 'none' %}
+    {# No button is shown #}
+  {% else %}
+    {# defaulting to back button #}
+    <text action="back" href="#" style="header-btn">Back</text>
+  {% endif %}
```
flochtililoch, line 9: I think it would be cleaner to explicitly handle the rendering of the button, and default to no button when prop is not set:
```suggestion
  {% if hv_button_behavior == 'close' %}
    <text action="close" href="#" style="header-btn">Close</text>
  {% elif hv_button_behavior == 'back' %}
    <text action="back" href="#" style="header-btn">Back</text>
  {% endif %}
```

examples/_includes/loading-screen.xml.njk
```diff
@@ -0,0 +1,11 @@
+{# Included as a loading screen in other templates #}
+<screen id="loadingScreen">
```
flochtililoch, line 2: we should follow the kebab-case convention for all attributes values
```suggestion
<screen id="loading-screen">
```

examples/_includes/styles.xml.njk
```diff
@@ -0,0 +1,119 @@
+<style
+  id="header"
+  alignItems="center"
+  backgroundColor="white"
+  borderBottomColor="#eee"
+  borderBottomWidth="1"
+  flexDirection="row"
+  paddingBottom="16"
+  paddingLeft="24"
+  paddingRight="24"/>
```
flochtililoch, line 10: to add to the conventions list:
> when rendering one XML attribute per line, the closing tag should also be on its own line and aligned with the opening tag

```suggestion
  paddingRight="24"
/>
```

flochtililoch (commented):
Could you add to our conventions list the following, then apply throughout the code change here? I'll do another pass after that, thanks!

We accept two formats to render attributes of an XML element:
- inline (provided that the total number of characters does not exceed 80)
- one attribute per line

When rendering the attributes inline, the element closing delimiter and the element closing tag should be on the same line. If the element is a self closing tag, there should be a single space between the last attribute's value and the self closing tag. If the element is not a self closing tag, there should be no leading space before the element delimiter.
When rendering one attribute per line, the element delimiter should be rendered on a new line, and the indentation level should match the element opening tag.

i.e:
**DO:**
```xml
<text id="foo"></text>
<behavior href="/" />
<behavior
  href="/"
  action="new"
/>
```

**DON'T:**
```xml
<text id="foo" ></text>
<behavior href="/"/>
<behavior
  href="/"
  action="new"/>
```

examples/_includes/tab.xml.njk
```diff
@@ -0,0 +1,40 @@
+{% extends 'base.xml.njk' %}
+
+{% block container %}
+  <text style="description">This is the content for tab 1.</text>
```
hgray-instawork, line 4: Do we need this?

examples/behaviors/index.xml.njk
```diff
@@ -0,0 +1,7 @@
+---
+hv_title: "Behaviors"
+permalink: "/behaviors/index.xml"
+hv_tab_id: "tab-behaviors"
+hv_button_behavior: "none"
+---
+{% extends 'tab.xml.njk' %}
```
hgray-instawork, line 7: Missing the content block?

examples/navigation/index.xml.njk
```diff
@@ -0,0 +1,7 @@
+---
+hv_title: "Navigation"
+permalink: "/navigation/index.xml"
+hv_tab_id: "tab-navigation"
+hv_button_behavior: "none"
+---
+{% extends 'tab.xml.njk' %}
```
hgray-instawork, line 7: Missing the content block?

examples/case-studies/index.xml.njk
```diff
@@ -0,0 +1,7 @@
+---
+hv_title: "Case Studies"
+permalink: "/case-studies/index.xml"
+hv_tab_id: "tab-case-studies"
+hv_button_behavior: "none"
+---
+{% extends 'tab.xml.njk' %}
```
hgray-instawork, line 7: Missing the content block?

examples/_includes/tab.xml.njk
rohanbayya: I initially added this to show-case support for content. Moving this from here to each individual tab.

PR #888, opened 2024-06-21
fix(logging): object debugging

hgray-instawork:
Forcing all params in DefaultLogger to string caused any objects to be converted to `[object Object]` making it impossible to inspect the element in Chrome.

Checking whether `toString` was available yielded false positive as all objects contain this function.

**Solution**
Relying on `instanceof` to ensure only `ToStringHelper` params are converted.

Before
before

After
after

### DIFF
--- a/src/services/logging/default-logger.ts
+++ b/src/services/logging/default-logger.ts
@@ -1,29 +1,39 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { ToStringHelper } from './tostring-helper';
+
 /**
  * The default logger performs toString() operations on each param passed to the logging actions
  */
 export class DefaultLogger {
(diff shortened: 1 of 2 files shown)


src/services/logging/default-logger.ts
```diff
@@ -1,29 +1,39 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { ToStringHelper } from './tostring-helper';
+
 /**
  * The default logger performs toString() operations on each param passed to the logging actions
  */
 export class DefaultLogger {
+  toString = (m?: any): any => {
```
flochtililoch, line 9: This does only perform toString on part of the code, so not sure that's the appropriate name.

flochtililoch (approved):

PR #892, opened 2024-06-24
feat(behavior): add support for prefetch action

palaparthi:
Add support for new action called `prefetch`, this will fetch an URL in the background.
Offline mode(built internally at Instawork) will take advantage of this by prefetching the necessary screens and saving them in the cache.

### DIFF
--- a/src/behaviors/hv-prefetch/index.ts
+++ b/src/behaviors/hv-prefetch/index.ts
@@ -31,10 +31,10 @@ export default {
     }
 
     Behaviors.setRanOnce(element);
-    const url = UrlService.getUrlFromHref(href, screenState.url || '');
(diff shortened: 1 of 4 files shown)


hgray-instawork (approved):
Not in scope for this but I wonder if the behaviors should receive the entire `onUpdateCallbacks` object?

palaparthi:
> Not in scope for this but I wonder if the behaviors should receive the entire `onUpdateCallbacks` object?

I thought about it too but decided otherwise since that needed a refactor but I agree with you.


@flochtililoch suggested that this is probably better suited as a custom behavior since this is not documented/available for public use. I will park the PR for now and come back to this later once I have offline write stuff working.

PR #896, opened 2024-07-10
fix(warning): explicitly set key instead prop spreading

fbinz:
Spreading the key-prop leads to annoying warnings in development. This PR explicitly sets the property.

Relevant react PR: https://github.com/facebook/react/pull/25697

### DIFF
--- a/src/services/render/index.tsx
+++ b/src/services/render/index.tsx
@@ -57,11 +57,6 @@ export const renderElement = (
     inlineFormattingContext = InlineContext.formatter(element);
   }
 
-  const componentOptions = {
-    ...options,
-    inlineFormattingContext,
-  } as const;
-
   if (element.nodeType === NODE_TYPE.ELEMENT_NODE) {
     if (!element.namespaceURI) {
       Logging.warn('`namespaceURI` missing for node:', element.toString());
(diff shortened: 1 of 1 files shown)


src/services/render/index.tsx
```diff
@@ -80,24 +75,27 @@ export const renderElement = (
...
+        element,
+        onUpdate,
+        options: {
+          ...options,
+          inlineFormattingContext,
+        },
+        stylesheets,
+      };
+
       /* eslint-disable max-len */
```
flochtililoch, line 89: minor: can this comment be removed without causing new eslint errors? if not, could you re-enable it after the comments block (as previously done line 87)

src/services/render/index.tsx
```diff
@@ -80,24 +75,27 @@ export const renderElement = (
...
-      // Object spreading will define the prop only when its value is truthy
-      /* eslint-enable max-len */
-      const extraProps: Partial<{ [key: string]: string | null }> = {};
-      if (element.getAttribute('key')) {
-        extraProps.key = element.getAttribute('key');
+      // Conditionally render the component with a key if it exists, to avoid
+      // warnings with current React versions, when the key attribute is set using the spread operator.
+      const key = element.getAttribute('key');
+
+      /* eslint-disable react/jsx-props-no-spreading */
```
flochtililoch, line 94: minor: either re-enable that rule after line 97, or explicitly disable both lines with `eslint-disable-line` comment.

flochtililoch (approved):
We appreciate your contribution @fbinz ! I added two minor comments related to eslint comments, would you have a couple minutes to address these before I merge? Thanks

fbinz:
@flochtililoch 
Thanks for the review. I changed the two lines as suggested.

src/services/render/index.tsx
```diff
@@ -80,24 +75,26 @@ export const renderElement = (
...
+        stylesheets,
+      };
+
+      // Conditionally render the component with a key if it exists, to avoid
+      // warnings with current React versions, when the key attribute is set
+      // using the spread operator.
+      const key = element.getAttribute('key');
+
+      if (key) {
+        <Component {...props} key={key} />; // eslint-disable-line react/jsx-props-no-spreading
```
flochtililoch, line 95: @fbinz I think you've accidentally removed the return statement here

fbinz: Oh, damn! Glad you caught that. Sorry for the troubles.

flochtililoch: thanks for the fix!

PR #897, opened 2024-07-10
chore(navigation): skip animation for first stack screen

hgray-instawork:
When navigating to a new root navigation structure, we are seeing a push animation as if the new structure is being pushed onto the stack.

Using the index of the defined screens, the first screen will inhibit its animations. Subsequent screens will retain their correct animations.

> If a nested stack exists, it is a route within its parent so it will still animate in correctly

| Before | After |
| -- | -- |
| before | after |

Example navigating beyond the initial reload
before

Asana: https://app.asana.com/0/1204008699308084/1205888766113004/f

### DIFF
--- a/src/core/components/hv-navigator/index.tsx
+++ b/src/core/components/hv-navigator/index.tsx
@@ -119,7 +119,7 @@ export default class HvNavigator extends PureComponent<Props> {
   };
 
   /**
-   * Build an individual tab screen
+   * Build an individual screen
    */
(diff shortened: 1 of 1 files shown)


hgray-instawork:
> should we add a text on the button next to toggle?

The button is just temporary UI to test the feature. It is not part of the PR.

rohanbayya (approved):

flochtililoch (approved):

PR #902, opened 2024-07-23
fix: VisibilityDectectingView requires a default style

hgray-instawork:
The partner app exhibits issues with layout when using a behavior with a trigger of `visible`. The root cause is the lack of styles assigned to `href-style` in the partner app templates.

- Apply a default style with `flex: 1` if no style is assigned.

| Before | After |
| -- | -- |
| before | after |

Asana: https://app.asana.com/0/1204008699308084/1207827683736609/f

### DIFF
--- a/src/VisibilityDetectingView.tsx
+++ b/src/VisibilityDetectingView.tsx
@@ -111,7 +111,7 @@ export default class VisibilityDetectingView extends PureComponent<Props> {
         ref={this.onRef}
         // collapsable has to be false for view.measure to work on Android
         collapsable={false}
-        style={this.props.style}
+        style={this.props.style || { flex: 1 }}
       >
         {this.props.children}
       </View>


src/VisibilityDetectingView.tsx
```diff
@@ -111,7 +111,7 @@ export default class VisibilityDetectingView extends PureComponent<Props> {
         ref={this.onRef}
         // collapsable has to be false for view.measure to work on Android
         collapsable={false}
-        style={this.props.style}
+        style={this.props.style || { flex: 1 }}
```
flochtililoch, line 114: this has the potential to break layouts in the app so I wouldn't apply a default here, rather use the style property where needed to apply the flex when needed.

hgray-instawork:
Implementing in templates instead of component

PR #906, opened 2024-07-24
adding style examples into new demo app

rohanbayya:


examples/ui/styling/button/disabled/index.xml.njk
```diff
@@ -0,0 +1,5 @@
+{% from 'macros/button/index.xml.njk' import button %}
+{% from 'macros/description/index.xml.njk' import description %}
+
+{{ description('Disabled') }}
+{{ button('Disabled', attributes={style:"button button-disabled"}) }}
```
rohanbayya, line 5: using macro instead of actual marku-up because in this case style is being show-cased. @flochtililoch

flochtililoch (approved):

Issue #914, opened 2024-07-31
`set-value` action does not fire `change` behavior of related fields


flochtililoch:
Simple repro case:
```
<text-field id="foo" value="Foo">
  <behavior trigger="change" action="show" target="bar" />
</text-field>
<text id="bar">Changed!</text>
<text action="set-value" target="foo" new-value="">Clear</text>
```

Upon pressing the Clear button, the change trigger should cause related change behavior to fire. While the use case described here isn't super useful, this change will comes handy when using the `change` behavior as a mean to perform custom logic via custom behaviors or custom elements that react to specific events.

Shaykoo:
@flochtililoch Is it ok for you if I can work on this one?

flochtililoch:
hi @Shaykoo, yes, your contributions are welcome!

PR #916, opened 2024-07-31
fix(text-field): `set-value` triggers `change` behaviors

flochtililoch:
React to text-field value being changed from props, and correctly call change handler.

Relates to #914.

Note - we currently have only two field types that support "change" event: `text-field` and `switch`. This PR only addresses the issue with `text-field` as this component was already migrated to a functional component, and the use of `useEffect`/`useRef` is trivial. Follow up needed:
- The switch component should be refactored to a functional component, then implement the same fix
- The "change" behavior trigger should be extended to other type of fields (related: https://github.com/Instawork/hyperview/issues/298)

https://github.com/user-attachments/assets/105fb0ac-9239-480a-b86c-e1234a07b28a

### DIFF
--- a/src/components/hv-text-field/index.tsx
+++ b/src/components/hv-text-field/index.tsx
@@ -2,7 +2,7 @@ import * as Behaviors from 'hyperview/src/services/behaviors';
 import * as Logging from 'hyperview/src/services/logging';
 import * as Namespaces from 'hyperview/src/services/namespaces';
 import type { HvComponentProps, TextContextType } from 'hyperview/src/types';
-import React, { MutableRefObject, useCallback, useRef } from 'react';
+import React, { MutableRefObject, useCallback, useEffect, useRef } from 'react';
 import {
   createProps,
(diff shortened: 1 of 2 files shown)


src/components/hv-text-field/index.tsx
```diff
@@ -77,6 +77,14 @@ const HvTextField = (props: HvComponentProps) => {
     null as TextInput | null,
   );
 
+  const prevDefaultValue = useRef<string | undefined>(defaultValue);
```
palaparthi, line 80: Can this be a state instead of a ref?

flochtililoch: It could, but I'm not sure why this would be better?

palaparthi: Thinking more about it I think ref makes sense here since we don't want to cause a re-render on updating the value, nvm.

palaparthi (approved):

hgray-instawork (approved):

PR #934, opened 2024-08-22
feat(network-retry): add network retry action

palaparthi:
This PR adds support for `network-retry-action`.
When the attribute is present, we will pass down the value in the request header (`X-Network-Retry-Action`).
This is meant for retrying non GET requests when the network call fails. It is up to the consumer of the Hyperview library to handle this header as part of the fetch implementation (passed in as a prop).

### DIFF
--- a/src/core/components/hv-root/index.tsx
+++ b/src/core/components/hv-root/index.tsx
@@ -19,6 +19,7 @@ import {
   HvComponentOptions,
   NAV_ACTIONS,
   NavAction,
+  NetworkRetryAction,
   OnUpdateCallbacks,
   UPDATE_ACTIONS,
   UpdateAction,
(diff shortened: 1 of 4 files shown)


flochtililoch (commented):
is there anything preventing this code to be added in our mobile app rather than Hyperview core?

palaparthi:
> is there anything preventing this code to be added in our mobile app rather than Hyperview core?

We would need to add a request [header](https://github.com/Instawork/hyperview/pull/934/files#diff-823ca03e16c6c1b0aff94177bfd83a0cdb65741ecc6e087572e3b463c74d66dfR87), I don't think that would be possible with a custom behavior but perhaps there is a way to do it?

adamstep:
@flochtililoch I think this is similar level support we needed like for the v1 of offline mode (for GET requests). Checking for headers to trigger certain low-level logic: https://github.com/search?q=repo%3AInstawork%2Fhyperview%20X_RESPONSE_STALE_REASON&type=code

adamstep (approved):

PR #939, opened 2024-09-04
fix(schema): view aliases common schema

flochtililoch:
`form` element is handled by same component as `view` and has the same capabilities, reflecting this on the schema


adamstep (approved):
Maybe we should create an attribute group `viewAttributes` to make this simpler to maintain

flochtililoch:
> Maybe we should create an attribute group `viewAttributes` to make this simpler to maintain

Good idea, I refactored to apply to every elements that are aliases of view

PR #940, opened 2024-09-04
feat(lists): Add support for `keyboardShouldPersistTaps`

reshavsingla:
Add support for `keyboardShouldPersistTaps` to List and SectionList. 

| List  | SectionList  |
|---|---|
| list  |  sectionlist |

### DIFF
--- a/examples/ui/ui-elements/list/keyboard-persist-tap/index.xml.njk
+++ b/examples/ui/ui-elements/list/keyboard-persist-tap/index.xml.njk
@@ -1,0 +1,48 @@
+---
+permalink: "/ui/ui-elements/list/keyboard-persist-tap/index.xml"
+tags: "List"
+hv_title: "Keyboard Persist Tap"
+hv_button_behavior: "back"
+---
+{% extends 'templates/base.xml.njk' %}
+
+{% block styles %}
+  <style
+    id="input"
+    borderBottomColor="#E1E1E1"
+    borderBottomWidth="1"
+    borderColor="#4E4D4D"
+    flex="1"
+    fontSize="16"
+    fontWeight="normal"
+    paddingBottom="8"
+    paddingTop="8">
+    <modifier pressed="true">
+    <style borderBottomColor="green"/>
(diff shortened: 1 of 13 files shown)


examples/ui/ui-elements/list/keyboard-persist-tap/index.xml.njk
```diff
@@ -0,0 +1,48 @@
...
+    <modifier focused="true">
+    <style borderBottomColor="#4778FF"/>
+    </modifier>
+  </style>
+{% endblock %}
+
+{% block container %}
+  <list keyboard-should-persist-taps="handled">
+    <item key="field" style="item">
+      <text-field name="field" style="input" auto-focus="true" placeholder="Tap!" />
```
flochtililoch, line 32: We now have the ability to create sudo component to abstract common XML markup, these live under `examples/_includes/` and can be included using the syntax `{% from ... import ... %}`.
Would you mind abstracting the text field and accompanying styles as a macro?

flochtililoch (commented):
Second thought on this, I wonder if we should make `handled` the default value to match how scroll views are setup. In theory this shouldn't affect existing usage of section lists, and we would still have a server side way to solve any issue if needed with the new argument.

reshavsingla:
> Second thought on this, I wonder if we should make `handled` the default value to match how scroll views are setup. In theory this shouldn't affect existing usage of section lists, and we would still have a server side way to solve any issue if needed with the new argument.

Making the default as `handled` might break the intended use for the community. So, I would argue supporting the current default behavior and add ability to override it when needed through pros.

examples/ui/ui-elements/list/keyboard-persist-tap/index.xml.njk
reshavsingla: Addressed in https://github.com/Instawork/hyperview/pull/940/commits/f366c4edd5b2da8cf7585c15b65cdb533070b3f6

flochtililoch (approved):

PR #941, opened 2024-09-11
chore(navigation): block gesture closing of modal

hgray-instawork:
To maintain existing Hyperview modal behavior, disable gesture (swipe) closing of modals. React-navigation uses a `true` default for iOS and `false` for Android.

Future tasks will consider overriding this default.

Asana: https://app.asana.com/0/1204008699308084/1208218286764693/f


src/core/components/hv-navigator/index.tsx
```diff
@@ -154,6 +154,7 @@ export default class HvNavigator extends PureComponent<Props> {
             cardStyleInterpolator: isModal
               ? NavigatorService.CardStyleInterpolators.forVerticalIOS
               : undefined,
+            gestureEnabled: false,
```
flochtililoch, line 157: Shouldn't we condition based on `isModal`? I assume this also disables the swipe right, which we wouldn't want. Also only iOS should be allowed to swipe right on stacks

flochtililoch (approved):

PR #945, opened 2024-09-24
fix(modal): visual bug during open animation

flochtililoch:
Fix a bug where the modal briefly renders full height before the open animation starts. This was caused by the `onLayout` callback from the inner content of the modal being triggered before the `onShow` callback of the `Modal` component.

| Before | After |
|-------|------|
| before | after |


adamstep:
Are picker/date-field the only components that use the modal?

adamstep (approved):

flochtililoch:
> Are picker/date-field the only components that use the modal?

yes

PR #946, opened 2024-09-27
feat(hv-navigator): Support for behaviors with `on-event` trigger

flochtililoch:
Add the ability to install event listener on the `<navigator>` tag, for example to allow reloading the navigator document upon certain event occurring.


adamstep (approved):

src/core/components/hv-navigator/index.tsx
```diff
@@ -43,8 +44,13 @@ export default class HvNavigator extends PureComponent<Props> {
 
   componentDidMount() {
     this.triggerLoadBehaviors();
+    Events.subscribe(this.onEventDispatch);
   }
 
+  componentWillUnmout = () => {
```
hgray-instawork, line 50: Looks like a typo on the function name here.

Issue #956, opened 2024-10-08
useEffect loop crashes new demo BottomTabBar component


ehenighan:
Hi,

By pure coincidence I was having similar problems relating to the lack of a sticky nav bar at the bottom and tried this recent new component out when my other idea didn't work (show-during-load = existing screen, hide-during-load = existing screen content box -> fails because the hide-during-load isn't applied during navigations, booo)

Immediately ran into a problem where the useEffect in the component gets into a loop, and I think it's because of the context provider:

https://github.com/Instawork/hyperview/blob/b5a6ada2bd5b7aa8d28a718644be27ef8a48cdf7/demo/src/Contexts/BottomTabBar.tsx#L27
```
<BottomTabBarContext.Provider
  value={{
    elementsProps,
    setElementProps: (id: string, p: HvComponentProps) => {
      //this de-structures the object and therefore always results in a new one.
      //I think this is what causes a render loop via the useEffect in the rendering component
      setElementsProps({ ...elementsProps, [id]: p });
    },
  }}
>
```

'Fixed' it with the following refactoring:
```
const setElementProps = useCallback((id: string, p: HvComponentProps) => {
    //only de-structure and replace the props on the element
    // if the component props for the specified ID is actually a new object
    if (!!p && (elementsProps[id] || {}) === p) {
        return;
    }
    setElementsProps({ ...elementsProps, [id]: p });
}, [elementsProps, setElementsProps]);
return (
    <BottomTabBarContext.Provider
        value={{
            elementsProps,
            setElementProps: setElementProps,
        }}
    >
        {props.children}
    </BottomTabBarContext.Provider>
);
```
I think the useCallback is probably overkill and it's the early-return condition in the function that's making the difference. Initially I thought this might be a niche case because I was using URL-loaded SVG icons and I considered if they might be causing lots of re-renders when the loading completed, but I took them out and the problem persisted.

I've not raised a PR for this directly because after my fix I'm also getting some other weird behaviour where after a successful initial load, the rest of the screen content just doesn't render when you select an item, leaving the nav bar floating in the centre of the page half-styled. The network request goes off in the background and the full screen content is returned to the app, but it doesn't load properly. If you click back and forth between tabs a bit, sometimes the destination page eventually loads once, but not again after that.

I've either got no rendering because of the useEffect crash, or weird UI behaviour _with_ my fix which I've not yet been able to debug, so I figured I'd raise an issue with what I've found so far and see if anyone has advice before I go further with it.

[...]

PR #962, opened 2024-10-14
fix(demo): custom bottom bar component

flochtililoch:
Fix double rendering of the custom bottom bar component by refactoring the following pieces:
- context: switch to using the reducer pattern, and expose 3 methods: `getElementProps`, `setElementProps`, `setElement`.
- core component: render any child element of the custom "non-render" element, not just the assumed single-child
- custom element: set the `registered` attribute on the element upon setting the props in the context, to avoid re-render loops when the screen holding the custom element re-renders. This does not prevent server side updates of the tab bar.
- update the demo app to factorize the markup for the tab bar as well as to show case the server updates of the tab bar.

Special thanks to @ehenighan for the thorough investigation and help solving #956.

https://github.com/user-attachments/assets/756d8585-d563-4a23-b845-3b42a843c4e7

## Summary by CodeRabbit

- **New Features**
  - Introduced a notification badge design for the bottom tab bar.
  - Added a new tab bar component to manage notifications within the second tab.

- **Enhancements**
  - Improved the BottomTabBar's context management with a reducer pattern for better state handling.
  - Streamlined the BottomTabBar component's logic and rendering process.
  - Enhanced tab navigation with a modular approach for dynamic rendering.

- **Bug Fixes**
  - Resolved rendering loops in the BottomTabBar component.

- **Documentation**
  - Updated context exports to prioritize the new custom hook for easier access.


demo/src/Components/BottomTabBar.tsx
```diff
@@ -28,9 +28,12 @@ const BottomTabBar = (props: HvComponentProps) => {
       );
       return;
     }
-    ctx.setElementProps?.(navigator, props);
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [navigator, props]); // Exclude ctx from dependencies to avoid infinite loop
+    // To avoid rendering loops, we only set the element props once
+    if (props.element.getAttribute('registered') !== 'true') {
```
adamstep, line 32: What happens when the screen triggers a `reload` action and potentially gets a new bottom tab bar from the server? Is that handled correctly?

flochtililoch: Yes, it is handled correctly - the video I attached to the PR description showcases this.

adamstep (approved):

demo/src/Core/BottomTabBar/index.tsx
```diff
@@ -13,43 +11,37 @@ import type { Props } from './types';
...
-  if (!props || !child) {
+  const onUpdateCustom = useCallback(
+    (href, action, currentElement, opts) => {
+      if (action === 'swap' && opts?.newElement) {
+        if (currentElement.parentNode) {
+          const newElement = currentElement.parentNode as Element;
+          newElement.replaceChild(opts.newElement, currentElement);
+          setElement?.(id, newElement);
+        } else {
+          console.warn('Parent node is null. Cannot replace child element.');
```
hgray-instawork, line 29: Better to use Logging.warn. See implementations in places like https://github.com/Instawork/hyperview/blob/231688b0b8416c74f2afa57b0edfffa2dfebfd33/src/behaviors/hv-unselect-all/index.ts

I found other console calls in the new tab bar components which aren't part of this PR.
```suggestion
          console.warn('Parent node is null. Cannot replace child element.');
```

hgray-instawork (approved):

demo/src/Core/BottomTabBar/index.tsx
flochtililoch: I don't believe the demo app has the logger set up already?

hgray-instawork: Disregard. This is not internal to hyperview.

PR #973, opened 2024-10-22
fix(demo): `BottomTabBar` tab content reload

flochtililoch:
Refactor `BottomTabBar` to use custom markup for items, that keeps track of selection state and performs navigation.
 This fixes an issue with the previous implementation that relied on a `<select-single>` / `<option>` combo to allow client side state + navigation with default Hyperview action, which would cause navigation to be performed when the tab bar was reloaded by a tab not currently focused.

See commits breakdown for details

| Before | After |
|--------|------|
| before | after |


hgray-instawork (approved):
Nice!

PR #954, opened 2024-10-06
getPickerInitialValue method for better redability and reduced redundency

Shaykoo:
I added the "some" method, which checks if any picker item has the given value, making the intent clearer. It avoids mapping the array to values when checking for existence, improving performance slightly.

## Summary by CodeRabbit

- **Refactor**
	- Streamlined the logic for determining the initial value of the picker in the `HvPickerField` component, enhancing performance and readability.

PR #975, opened 2024-10-23
Cache the value-label pairs when the picker items are fetched

Shaykoo:
The current getLabelForValue method loops through all picker items each time it is called. This can be inefficient if you have a large list of items. Caching the mapping of values to labels will improve performance.

### DIFF
--- a/src/components/hv-picker-field/index.ios.tsx
+++ b/src/components/hv-picker-field/index.ios.tsx
@@ -34,21 +34,21 @@ export default class HvPickerField extends PureComponent<HvComponentProps> {
     return getNameValueFormInputValues(element);
   };
 
+
   getPickerInitialValue = (): string => {
     const value = this.getValue();
     const pickerItems: Element[] = this.getPickerItems();
-    if (
-      pickerItems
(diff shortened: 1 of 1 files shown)


src/components/hv-picker-field/index.ios.tsx
```diff
@@ -37,50 +37,51 @@ export default class HvPickerField extends PureComponent<HvComponentProps> {
...
+    if (valueExists) {
       return value;
     }
-    if (pickerItems.length > 0) {
-      return pickerItems[0].getAttribute('value') || '';
-    }
-    return '';
+    return pickerItems.length > 0
+      ? pickerItems[0].getAttribute('value') || ''
+      : '';
```
flochtililoch, line 48: @Shaykoo your branch needs a rebase with the latest master, this is your previous change that got merged in #954

flochtililoch (commented):
@Shaykoo thanks for the submission. Looking over the change, I think we need some functionality to reset that cache, to ensure we're not operating on outdated data if the underlying XML has changed.

PR #978, opened 2024-10-23
feat(demo): Map/MapMarker components

flochtililoch:
Import Instawork's Map / MapMarker components into the demo app.

See commits breakdown for details.

| iOS | Android | Web |
|---|---|---|
| Simulator Screenshot - iPhone 16 Pro - 2024-10-23 at 08 15 07 | Screenshot 2024-10-23 at 8 17 32 AM | Screenshot 2024-10-23 at 10 14 33 AM |

### DIFF
--- a/demo/app.config.ts
+++ b/demo/app.config.ts
@@ -7,9 +7,8 @@ export default {
   },
   assetBundlePatterns: ['**/*'],
   extra: {
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
     baseUrl: process.env.BASE_URL || 'http://0.0.0.0:8085',
+    googleMapsApiKey: process.env.GOOGLE_MAPS_API_KEY || '', // Leaving the key empty will enable dev-only experience
   },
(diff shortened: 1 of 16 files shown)


demo/app.config.ts
```diff
@@ -7,9 +7,8 @@ export default {
   },
   assetBundlePatterns: ['**/*'],
   extra: {
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
     baseUrl: process.env.BASE_URL || 'http://0.0.0.0:8085',
+    googleMapsApiKey: process.env.GOOGLE_MAPS_API_KEY || '', // Leaving the key empty will enable dev-only experience
```
adamstep, line 11: What is the difference with dev-only mode? Is there a watermark, or some missing functionality?

flochtililoch: yes, there's a water mark and a popup when UI loads, and the markers are not rendering in this situation.

hgray-instawork (approved):
Very nice!

PR #995, opened 2024-11-15
feat(loading): add support for named loading screens

hgray-instawork:
Adding a new feature to Hyperview to allow passing in named loading screen components.

Example:
```es6
<Hyperview
  ...
  loadingScreens={{
    'green-loader': GreenLoadingScreen,
  }}
/>
```

```es6
<behavior
  action="new"
  href="/hyperview/public/navigation/behaviors/actions/modal/index.xml"
  show-during-load="green-loader"
/>
```

https://github.com/user-attachments/assets/8a3d3bd1-510b-42e5-95e4-b9a4f8505ddd

See commits for breakdown.

[Asana](https://app.asana.com/0/1204008699308084/1208750137800484/f)

PR #1001, opened 2024-11-22
feat(loading): update demo

hgray-instawork:
Update the demo with a "Loading Screen" section with examples of using loading screens.

>NOTE: This PR will fail tests until HV is published as the demo relies on the released version and `loadingScreens` will not be available until that is released.

Relates to: https://github.com/Instawork/hyperview/pull/995


hgray-instawork:
Closing in favor of alternate approach

PR #1004, opened 2024-12-03
refactor: Move default logger to demo app

flochtililoch:
When a logger is not specified as a prop of the entry point Hyperview component, we currently instantiate a default logger that will emit log lines to the console. This can causes performance issues for devs that don't implement a logger, and is noticed in the current demo app.

This PR removes the default logger from core, and implement it explicitly in the demo app. The default log level in the demo app is set to `log`, to avoid logging behavior logs, which contribute to a more sluggish UX.


hgray-instawork (approved):
Nice!

PR #1005, opened 2024-12-11
fix(navigation): action reload is not reading href

hgray-instawork:
Resolving issue with `on-event` behaviors embedded in `<navigator>` elements no properly performing `reload` action.

- [fix: only reset flag when loading](https://github.com/Instawork/hyperview/pull/1005/commits/b665f2d721a58cf5df0b97066470231defaa009f): copied over from [`hv-screen` implementation](https://github.com/Instawork/hyperview/blob/42e1534c371b32fff97efaaf122f79165c06d7aa/src/core/components/hv-screen/index.js#L176-L180). The issue with having it outside the if/then is that it could be cleared when a parent component causes an update before this component has updated its own state.
- [fix: include state in url logic](https://github.com/Instawork/hyperview/pull/1005/commits/dac168c5e69e0de7e414f623cab4328848c4e71d): as with `hv-screen`, the update mechanism needs to inject data that may have been passed into the `state`. See [hv-screen](https://github.com/Instawork/hyperview/blob/42e1534c371b32fff97efaaf122f79165c06d7aa/src/core/components/hv-screen/index.js#L205)
- Added a demo to illustrate the feature.
  - created an `index-reloaded` file to represent the content to load into the hv-route
  - the templates [have to inherit](https://github.com/Instawork/hyperview/pull/1005/commits/486bb1af5ce781e0878e22d258f6bd741a3cff8c) from an included template to ensure the headers didn't get copied into the generated xml templates

[Asana](https://app.asana.com/0/1204008699308084/1208942355610921/f)

| iOS | Android |
| -- | -- |
| iOS | Android |

QA Plan:
- ensure navigation works as normal including
  - deep links
  - modals
  - pushed screens

### DIFF
--- a/demo/backend/_includes/templates/root.xml.njk
+++ b/demo/backend/_includes/templates/root.xml.njk
@@ -1,3 +1,3 @@
-{# Base template that should be used for all screens. Shows the Header + back button #}
+{# Base template that should be used for all templates #}
 
 {% block root %}{% endblock %}
--- a/demo/backend/_includes/templates/root.xml.njk
+++ b/demo/backend/_includes/templates/root.xml.njk
@@ -1,0 +1,3 @@
+{# Base template that should be used for all screens. Shows the Header + back button #}
+
+{% block root %}{% endblock %}
--- a/demo/backend/index-reloaded.xml.njk
+++ b/demo/backend/index-reloaded.xml.njk
@@ -4,7 +4,6 @@ hv_button_behavior: "none"
 hv_title: "Hyperview"
 hv_list_tag: "root"
 ---
-
 {% extends './index.xml.njk' %}
 
 {% block advanced_attributes %}selected="true"{% endblock %}
--- a/demo/backend/index-reloaded.xml.njk
+++ b/demo/backend/index-reloaded.xml.njk
@@ -1,5 +1,5 @@
 ---
-permalink: "index.xml"
+permalink: "index-reloaded.xml"
 hv_button_behavior: "none"
 hv_title: "Hyperview"
 hv_list_tag: "root"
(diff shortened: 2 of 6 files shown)


src/core/components/hv-route/index.tsx
```diff
@@ -96,8 +96,8 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, ScreenState> {
   componentDidUpdate(prevProps: Types.InnerRouteProps) {
     if (prevProps.url !== this.props.url || this.needsLoad) {
       this.load();
+      this.needsLoad = false;
```
flochtililoch, line 99: is this change intentional? If so could you describe in the PR description why / how it affected the bug that's being fixed?

demo/backend/index-reloaded.xml.njk
```diff
@@ -0,0 +1,26 @@
+---
```
flochtililoch, line 1: to prevent complete duplication of the navigation markup from index.xml, could we:
- add template blocks in index.xml to allow setting the selected tab
- make this template extend index.xml and set the relevant block to select the desired tab

hgray-instawork: https://github.com/Instawork/hyperview/pull/1005/commits/54c0f7fd4a3b41884e760f1bab82bcd40f90e62d

demo/backend/_includes/templates/root.xml.njk
```diff
@@ -0,0 +1,3 @@
+{# Base template that should be used for all screens. Shows the Header + back button #}
```
flochtililoch, line 1: comment is outdated - perhaps its worth using this template as a root for the `doc` element and reuse in `base.xml` too?
```suggestion
{# Base template that should be used for all documents #}
```

hgray-instawork: Good catch, thanks. Yes, I was thinking I may as well make it root for everything.

https://github.com/Instawork/hyperview/pull/1005/commits/98355ce2a6d098de553ba4f75926ba5506c4d011

https://github.com/Instawork/hyperview/pull/1005/commits/6ccb5887a4a2255def3a35798a1adfa60136d393

flochtililoch: are your latest edit meant to address this or not? The root file is empty right now, my previous comment suggested to have:
```xml
<doc xmlns="https://hyperview.org/hyperview" {% block custom_ns %}{% endblock %}>
  {% block root %}{% endblock %}
</doc>
```

then de-dup the `doc` element from `demo/backend/_includes/templates/base.xml.njk` and `demo/backend/index.xml.njk`. Minor but since you're touching this code we may as well get that done.

flochtililoch (approved):

hgray-instawork:
QA completed through temp mobile integration branch: https://github.com/Instawork/mobile/pull/6837

PR #1007, opened 2024-12-16
chore(navigation): support alternate loading screens for behaviors and nav-routes

hgray-instawork:
As an update to Hyperview's loading screen functionality, this update allows us to pass a dynamic loading screen into Hyperview. The loading screen optionally receives the `<behavior>` or `<nav-route>` which caused it to appear, allowing the loading screen to retrieve attributes/children from the element and modify the component accordingly.

Example:
```
const LoadingScreen = (props: Props) => {
  const showDuringLoad = props.element?.getAttribute('show-during-load');
  const color = (showDuringLoad === 'green-loader') ? 'green' : '#8d9494';
  return (
    <View>
      <ActivityIndicator color={color} size="large" />
    </View>
  );
};
```

- [fix: use only first indicator](https://github.com/Instawork/hyperview/commit/7cab73c15f26db4e47db98856e4e60b7819d5d94) - [the documentation](https://hyperview.org/docs/reference_behavior_attributes#navigation-actions-1) states `show-during-load` should only contain a single screen id. This enforces this and also cleans up the naming/typing
- [fix: typo](https://github.com/Instawork/hyperview/commit/b667b2ab3bd63613f2e373233a02b8d8d3213a24)
- [chore: pass the behaviorElement](https://github.com/Instawork/hyperview/commit/56a27008192ac482f445e1f75b3b5aa20bb4ead9) - ensure the behavior element which triggered the navigation action is passed so it can be cached
- [chore: pass the behavior element id or route id in route params](https://github.com/Instawork/hyperview/pull/1007/commits/4dd745fd573c66a2beee1d12f1bca2de66983db3) - allow the routeId to be passed in params. Also, if no preloadScreen is found, attempt to inject a behavior element into the cache.
- [chore: simplify cache location](https://github.com/Instawork/hyperview/pull/1007/commits/976c83d75495fb898ff9cf8ea4d5621a94d14b21) - the element cache can be simplified by only having it at the top-level rather than in every navigator.
- [chore: support cache removal](https://github.com/Instawork/hyperview/pull/1007/commits/118083dae24d7fc77e0d0c5f218c625194aadd9c) - add support for clearing items from cache
- [chore: loadingScreen component receives optional element](https://github.com/Instawork/hyperview/pull/1007/commits/764ae7fe2edeabc0cb8523f9b4e8b264c64af13e) - the loading screen passed into Hyperview can now accept an optional element which represents the trigger of this loading event. This will either be the behavior which caused the navigation, or the `<nav-route>` which is being loaded.
- [chore: update Loading component and instantiation](https://github.com/Instawork/hyperview/pull/1007/commits/f78602acdfa580de500635de817661f7b0f5ba4f) - updating the internal `loading` component to support retrieval of cached elements, instantiation of any passed loading screen, and cache cleanup during unmount. Default spinner implementation remains as a backup.
- [chore: purge cache](https://github.com/Instawork/hyperview/pull/1007/commits/a2d0543f9f32c1a1f89605aef4a77adf01f2bc8a) - loading screens should purge the cache on unmount

Notes: some additional testing will be necessary in general with these changes but some smoke tests have shown positive results. Also, we should not have to worry about memory leaks as the containers which hold the cached screens (preload or behavior) are part of the route/screen hierarchy and they would get removed when the screen is removed.

TODO:
- [ ] - documentation updates
- [ ] - update demo. This must be done post-release as demo points to released version of Hyperview. See https://github.com/Instawork/hyperview/pull/1011

[TAD](https://instawork.atlassian.net/wiki/spaces/EN/pages/3671949446/TAD+-+Hyperview+Loading+Screens)

[Asana](https://app.asana.com/0/1204008699308084/1208750137800484/f)


src/core/components/hv-route/index.tsx
```diff
@@ -98,6 +98,29 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, ScreenState> {
       this.load();
       this.needsLoad = false;
     }
+
+    // Clean out loading screen cache after the screen has been rendered
+    if (
+      this.props.element ||
+      this.localDoc ||
+      this.props.route?.params?.isModal
```
flochtililoch, line 106: We want to clean up the cache _after_ the loading UI was removed.

A couple issues with the implementation:
- This will render every time any props changes including other props that are not part of this condition. This condition might be true multiple times, meaning this code will run more than what we need it to. Typically, we would need to compare the prevProps to only run the code when we know the condition to render the LoadScreen has changed.
- The condition needs to match exactly what's in the `render` method for this code to continue working in the future - it would be quite easy to break it if we don't pay attention, so ideally the condition should be shared between this code and the code in `render`. However I suspect this will become quite complicated to manage given that we also need to account for `prevProps`.

Looking at the implementation for the render method, we render a local `Load` component instance that is created either from the `loadingScreen` prop or the internal `Loading` component. Any way to wrap the creation of this component with another component that will implement in its "unmount" method the clean up for the cache?

src/core/components/hv-screen/index.js
```diff
@@ -218,6 +218,17 @@ export default class HvScreen extends React.Component {
       }
       const stylesheets = Stylesheets.createStylesheets(doc);
       this.navigation.setRouteKey(this.state.url, routeKey);
+
+      // Clean the preload screen out of cache after loading the screen
```
flochtililoch, line 222: is this redundant with the clean up done in hv-route, or complimentary?

hgray-instawork: This is because the thing which gets rendered may not be an `hv-screen`, it could be a `<navigator>`

src/core/components/hv-route/index.tsx
hgray-instawork: That's a good suggestion, I'll try that out and see if I can encapsulate the logic there

src/core/components/hv-route/index.tsx
```diff
@@ -302,8 +325,26 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, ScreenState> {
...
+        ? NavigatorService.getRouteById(
+            this.props.doc,
+            this.props.route?.params?.routeElementId,
+          )
+        : undefined;
+
+    const loadingScreen = this.props.loadingScreen || Loading;
+    return React.createElement(loadingScreen, {
+      element: behaviorElement || routeElement,
+    });
```
flochtililoch, line 347: minor: I think the imperative API to instantiate a component obfuscate the props required by the component itself from TS. I think it would also make sense to handle individual scenario to simplify logic, and reduce the diff too.

Would the following work?

```suggestion
    const LoadingScreen = this.props.loadingScreen || Loading;
    
    // Find the associated behavior element to show the loading screen
    const behaviorElement = this.props.route?.params?.behaviorElementId
      ? this.props.getPreload(this.props.route?.params?.behaviorElementId)
      : undefined;
    
    if (behaviorElement) {
      return <LoadingScreen element={behaviorElement} />;
    }
    
    // If no behavior element is found, check for a route element
    const routeElement =
      this.props.route?.params?.routeElementId &&
      this.props.doc
        ? NavigatorService.getRouteById(
            this.props.doc,
            this.props.route?.params?.routeElementId,
          )
        : undefined;
    if (routeElement) {
      return <LoadingScreen element={routeElement} />;
    }
    
    // No behavior element nor route element found
    return <LoadingScreen />;
```

src/core/components/hv-screen/index.js
```diff
@@ -208,6 +209,7 @@ export default class HvScreen extends React.Component {
       }
       const stylesheets = Stylesheets.createStylesheets(doc);
       this.navigation.setRouteKey(this.state.url, routeKey);
+
```
flochtililoch, line 212:
```suggestion
```

src/core/components/hv-screen/index.js
```diff
@@ -251,8 +263,15 @@ export default class HvScreen extends React.Component {
...
-      return React.createElement(loadingScreen);
+      return (
+        <Loader
+          behaviorElementId={this.props.route?.params?.behaviorElementId}
+          getPreloadScreen={this.navigation.getPreloadScreen}
+          loadingScreen={this.props.loadingScreen}
+          preloadScreen={this.props.route?.params?.preloadScreen}
+          removePreload={this.removePreload}
+        />
+      );
```
flochtililoch, line 274: Did you externalize the implementation for TS? The gain of typing might be canceled out by the added complexity of passing these local methods/props down to the component (and having to type them).

hgray-instawork: Revised approach incorporates this concept into the internal `loading` component

src/core/components/hv-navigator/index.tsx
```diff
@@ -301,7 +301,14 @@ export default class HvNavigator extends PureComponent<Props> {
...
         // This relies on the schema requirement that each <navigator> has at least one <nav-route>
         screens.push(
-          this.buildScreen(id, type, href || undefined, isModal, index === 0),
+          this.buildScreen(
+            id,
+            type,
+            href || undefined,
+            isModal,
+            index === 0,
+            id,
```
flochtililoch, line 310: did you intend to pass `id` as the extra param, or should it be `routeId`?

src/core/components/hv-route/index.tsx
```diff
@@ -298,12 +309,37 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, ScreenState> {
...
+          return this.props.route?.params?.routeId && this.props.doc
+            ? NavigatorService.getRouteById(
+                this.props.doc,
+                this.props.route?.params?.routeId,
+              )
+            : undefined;
+        }}
+        onUnmount={() => {
+          this.removePreload(this.props.route?.params?.behaviorElementId);
+        }}
```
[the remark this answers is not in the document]
hgray-instawork: Due to `hv-screen` providing its own internal caching, it seemed safer to allow the implementor to define the action. I can move this into `Loading` if we're not concerned about cleaning out the `hv-screen` cache.

src/core/components/hv-root/index.tsx
```diff
@@ -599,7 +566,37 @@ export default class Hyperview extends PureComponent<Types.Props> {
               reload: this.reload,
             }}
           >
-            <HvRoute />
+            <NavigatorMapContext.NavigatorMapProvider>
```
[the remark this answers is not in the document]
hgray-instawork: Two changes were required at the root level regarding contexts:
- The NavigatorMap was moved from inside each <navigator> up to the top level to simplify caching https://github.com/Instawork/hyperview/pull/1007/commits/4eef098a097f36932361f7843c5e22863ab8ff43
- Because `hvscreen` can be instantiated outside of the navigation hierarchy, but the loader requires access to both `loadingScreen` and `getPreload`, the context hierarchy needed to be updated to provide these https://github.com/Instawork/hyperview/pull/1007/commits/8734169c4b2b6dcb9880ccaac25cbbe00a1cd112

src/core/components/hv-route/index.tsx
```diff
@@ -298,12 +309,37 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, ScreenState> {
...
         if (component) {
-          return <>{component}</>;
+          return (
+            <>
+              {useEffect(() => {
+                return () => {
+                  // Perform cleanup when the preloadscreen component is unmounted
+                  this.removePreload(this.props.route?.params?.preloadScreen);
+                };
+              }, [])}
```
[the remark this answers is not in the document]
hgray-instawork: Thanks, I've revised with a cleaner approach: https://github.com/Instawork/hyperview/pull/1007/commits/4683cafeb83b82022a2e98231cc6be01f9cf4adc

src/core/components/loading/index.tsx
```diff
@@ -1,16 +1,70 @@
...
+
+  // Perform cleanup when the component is unmounted
+  useEffect(() => {
+    return () => {
+      if (props.behaviorElementId) {
+        navigatorMapContext?.removePreload(props.behaviorElementId);
+      }
+      if (props.preloadScreen) {
+        navigatorMapContext?.removePreload(props.preloadScreen);
+      }
```
flochtililoch, line 32: is there any scenario where both `behaviorElementId` and `preloadScreen` will be set together? If not maybe both these props can be collapsed into a single one?

src/core/components/loading/index.tsx
```diff
@@ -1,16 +1,70 @@
...
       </View>
     );
   }
-}
 
-export * from './types';
+  // Use the passed preloadScreen component
+  if (props.preloadScreenComponent) {
+    return <>{props.preloadScreenComponent}</>;
+  }
```
flochtililoch, line 52: this does not rely on the context - should this be handled before the previous condition?

src/core/components/loading/index.tsx
```diff
@@ -1,16 +1,70 @@
...
+
+  // The behavior element which triggered the load
+  const behaviorElement = props.behaviorElementId
+    ? navigatorMapContext?.getPreload(props.behaviorElementId)
+    : undefined;
+
+  // If the behavior element is not found, look for a route element
+  const element =
```
flochtililoch, line 60: is there a chance that element would not be defined here, and how should this be handled?

src/core/components/loading/index.tsx
hgray-instawork: The reason I'm tracking them as separate names is to ensure we don't try to render a behavior [here](https://github.com/Instawork/hyperview/blob/f5e702c4691789e7e752700d3ab4a3b91e4475a0/src/core/components/hv-route/index.tsx#L280-L304), [here](https://github.com/Instawork/hyperview/blob/f5e702c4691789e7e752700d3ab4a3b91e4475a0/src/core/components/hv-screen/index.js#L95-L110), or [here](https://github.com/Instawork/hyperview/blob/f5e702c4691789e7e752700d3ab4a3b91e4475a0/src/core/components/hv-screen/index.js#L145-L155).

An alternate approach would be to include a flag `isScreen` to indicate that the `preloadScreen` param is actually a `<screen>`. Then I'd change the three places above to check both `preloadScreen` and `isScreen === true`.

src/core/components/loading/index.tsx
hgray-instawork: Good catch. Resolved [here](https://github.com/Instawork/hyperview/pull/1007/commits/e69ebd0e63241d5fe929bf29e02af216b8e9d1c0). Note that I can't move it any higher as both the `useContext` and `useEffect` cannot be called conditionally.

src/core/components/loading/index.tsx
hgray-instawork: It is unlikely as every `navigate` call which is triggered by a behavior [will inject the behavior](https://github.com/Instawork/hyperview/blob/e69ebd0e63241d5fe929bf29e02af216b8e9d1c0/src/services/navigation/index.ts#L104-L106) (by unique id) into the params. And any route which is loading content can access itself from the doc.

In addition, the element is [an optional property](https://github.com/Instawork/hyperview/blob/e69ebd0e63241d5fe929bf29e02af216b8e9d1c0/src/core/components/loading/types.ts#L5) and the loader can function without it.

src/core/components/loading/index.tsx
flochtililoch: you can track them as separate name outside of the component, but that prop is not used in two different contexts within the component, so it seems there could just be a single prop? Or am I missing something?

flochtililoch (approved):

src/core/components/loading/index.tsx
hgray-instawork: Yup, thanks. Fixed here: https://github.com/Instawork/hyperview/pull/1007/commits/407cda948b0b5e323b89d60f0fbb55c3f0b8d34a

PR #1009, opened 2024-12-19
feat(scrolling): monitor scroll position

hgray-instawork:
Create a context which allows tracking the scroll position of child views and lists. Each scrollable component (view, list, section list, href) is wrapped as a context consumer and passes its scroll position up to the context. Any component which implement a provider can be notified of scroll position updates.

Data is cached by the id of the scrolling component. If no id is present, data is not logged.

Example observer:
```
<ScrollObserver.Context.Consumer>
  {context => (
    < ... content with view >
    <Text>{`context.offsets['view-id]`}</Text>
  )}
</ScrollObserver.Context.Consumer>
```

Based on work done by @flochtililoch in https://github.com/Instawork/hyperview/compare/master...florent/scroll-context, this 

---
  - https://app.asana.com/0/0/1208991383448367


flochtililoch:
Implemented in #1108

PR #1012, opened 2024-12-20
chore(navigation): simplify preload screen cache

hgray-instawork:
The existing cache behavior creates a unique id for each navigation action's `show-during-load` element and may cause an accumulation of cached items, particularly in base screens which don't get unmounted. There is a cache at each level of the navigation hierarchy.

This change simplifies each cache by only keeping a single preload screen per cache. Every time a `set` is called, the previous element is replaced in the cache.

[Asana](https://app.asana.com/0/1204008699308084/1209017355656018/f)


hgray-instawork:
Closing in favor of an alternate approach.

PR #1023, opened 2025-01-17
fix(demo): serialize forms in form field examples

flochtililoch:
Original demo app had built in support for serializing forms in field examples, but we had not re-implemented when we overhauled the codebase last summer.
This re-introduces support for serialization.

Implementation notes:
- restore the ability to override an endpoint with a dynamic JS handler
- provide a new global variable `__DEV__` to 11ty templates - value is set based on a environment variable
- set the env variable that causes `__DEV__` to be false when we deploy to GH pages
- conditionally render a serialize button at the bottom of each field

See commits breakdown for more details.

https://app.asana.com/0/1205761271270033/1209182447962484/f


demo/backend/_includes/templates/section-list.xml.njk
```diff
@@ -19,7 +19,7 @@
         {% set item_style = item_style ~ " item-last" %}
       {% endif %}
       <item
-        href="/hyperview/public/{{ screen.data.permalink }}"
+        href="/hyperview/public{{ screen.data.permalink }}"
```
hgray-instawork, line 22: does the permalink contain the leading slash?

hgray-instawork (approved):

demo/backend/_includes/templates/section-list.xml.njk
flochtililoch: yes although I see 2 templates out of 192 that don't have a leading slash

PR #1027, opened 2025-01-20
chore(loading screen): add documentation for loading screen

hgray-instawork:
[Asana](https://app.asana.com/0/1204008699308084/1209189437518136/f)


docs/reference_loading_screen.md
```diff
@@ -0,0 +1,74 @@
...
+/>;
+```
+
+The optional element which is passed into the loading screen provides a way to modify the appearance or functionality based on the XML structure which triggered the load. When loading is triggered via a navigation action, the `<behavior>` element is passed into the loading screen. When loading is triggered by a `<nav-route>` element, the `nav-route` is passed into the loading screen.
+
+### Example
+
+The following example shows how to use the `id` attribute of the `<behavior>` element to colorize the load indicator.
+
+```es6
```
flochtililoch, line 42:
```suggestion
```xml
```

flochtililoch (approved):

docs/reference_loading_screen.md
hgray-instawork: Thanks! Resolved https://github.com/Instawork/hyperview/pull/1027/commits/10b06ca7bbae7bf4640193cec6c462aa892bf0f6

PR #1028, opened 2025-01-21
fix(demo): navigation back/close examples

flochtililoch:
Address variety of issues with back and close examples.

See commits.

https://app.asana.com/0/1204008699308084/1209182447962502/f
https://app.asana.com/0/1204008699308084/1209182447962503/f


demo/backend/ui/ui-elements/forms/date-field/form.xml.js
```diff
@@ -1,12 +1,7 @@
-const urlParse = require('url-parse');
-const nunjucks = require('nunjucks');
-const { getContent } = require('../../../../_data/form.js');
+const { withFormData } = require('../../../../_data/render.js');
```
hgray-instawork, line 1: Should we clean up these long relative paths while we're in here?

hgray-instawork (approved):

demo/backend/ui/ui-elements/forms/date-field/form.xml.js
flochtililoch: this would require setting up the babel import plugin to allow absolute paths, so I'd rather not deal with this at the moment.

PR #1042, opened 2025-01-22
chore(loading screen): rename cache

hgray-instawork:
Refactoring the "navigation map" context and the "preload" naming to more appropriate "element" naming. See commits.

[Asana](https://app.asana.com/0/1204008699308084/1209189437518170/f)


src/core/components/hv-root/index.tsx
```diff
@@ -2,11 +2,11 @@ import * as Behaviors from 'hyperview/src/behaviors';
 import * as Components from 'hyperview/src/services/components';
 import * as Contexts from 'hyperview/src/contexts';
 import * as Dom from 'hyperview/src/services/dom';
+import * as ElementCacheContext from 'hyperview/src/contexts/element-cache';
```
flochtililoch, line 5: can we re-expose this context under the existing barrel file of `hyperview/src/contexts`, and save an additional import? I guess this would make sense to apply to any context that we import in this file, so I'm ok to defer this to a future PR if you'd rather leave this as is for now

flochtililoch (approved):

src/core/components/hv-root/index.tsx
hgray-instawork: See if this looks alright: https://github.com/Instawork/hyperview/pull/1042/commits/6090706925b2dd87dc09486c3939cb226bbd3762

PR #1048, opened 2025-01-22
feat(demo): add delay for localhost fetch

hgray-instawork:
After [fixing a bug in Hyperview](https://github.com/Instawork/hyperview/pull/1014), the demo application no longer shows loading indicators in localhost because it loads immediately.

This update introduces an artificial delay when running in localhost

[Asana](https://app.asana.com/0/1205761271270033/1209189437518211/f)

### DIFF
--- a/demo/src/Helpers/misc.ts
+++ b/demo/src/Helpers/misc.ts
@@ -1,3 +1,4 @@
+import Constants from 'expo-constants';
 import moment from 'moment';
 
 export const formatDate = (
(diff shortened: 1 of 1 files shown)


flochtililoch (commented):
We have the ability to configure some endpoints of the local server to return an artificial delay ([example](https://github.com/Instawork/hyperview/blob/master/demo/backend/navigation/behaviors/actions/changed-loading.xml.js#L8)). I don't think we should make this change on the client side and globally though, as this would affect the deployed demo app too, and probably would get in the way to testing.

hgray-instawork:
Closing in favor of alternate approach described above.

PR #1054, opened 2025-01-23
fix(demo): navigation/modal examples issues

hgray-instawork:
- remove delays
- fix language for modal and push screens (globally)
- adding example modal and push screens to resolve infinite loop and top button type

https://github.com/user-attachments/assets/3c6f5404-7683-4e51-9f8d-bfdfce3f3007

[Asana](https://app.asana.com/0/1205761271270033/1209182447962504/f)


flochtililoch (commented):
Looking at the example, I think the intention with the screen behaviors/action/modal/index.xml is to open as a modal (so the navigation action and the header should be updated). That example should show case both `new` ("Open another modal" button) and `push` ("Push on stack" button) actions on top of a modal. Let me know if you agree, or if you have a different understanding of this example.

PR #1087, opened 2025-02-18
chore(readme): update docs with node version

hgray-instawork:
Updating docs to indicate the current maximum `node` version.

[Asana](https://app.asana.com/0/1204008699308084/1209425378622141)


flochtililoch (approved):
I believe only the demo is incompatible with node 23+ (we don't run any node process on the library)

PR #1092, opened 2025-02-21
fix(navigation): resolve performance issue

hgray-instawork:
The `blur`, `focus`, and `state` listeners all can perform heavy dom mutations. Performing them at the same time the screen is rendering and transitioning can cause delays or visual glitches. To resolve, these listeners are delayed a small amount.

The use of `setTimeout` enables waiting for the screen to render before mutating the dom. The `navStateMutationsDelay` experiment value is used to enable the timeout. If the experiment is not set or has a value of "0", no timeout is used. See https://github.com/Instawork/hyperview/pull/1092/commits/35499d94ff3fd2b64f4dfc0c8d90deddc89b3812

| Before | After |
| -- | -- |
| before | after |

[Asana](https://app.asana.com/0/1204008699308084/1209403666139702)


flochtililoch (commented):
is setting `100ms` necessary, or would it work just as well if setting `0`, basically to schedule the code for the next tick?

flochtililoch (approved):

PR #1097, opened 2025-02-26
fix(web-view): support shared cookies

olegpesok:
Pass through shared cookies enabled prop for iOS.


olegpesok:
> It looks like the prop you're adding has existed in react-native-webview for 6+ years ([PR](https://github.com/react-native-webview/react-native-webview/pull/175)), so bumping the version might not be necessary. It also appears this prop is iOS only - does the same work on Android without an explicit prop?

* Yes, we don't need to bump the version, that's only for simulator support of this feature.
* Yes, this works with Android without an additional prop passed through.

flochtililoch (approved):

PR #1098, opened 2025-02-26
chore(navigation): consolidate dom mutation

hgray-instawork:
When a route gains focus, the dom may be mutated. This change combines them into a single method to simplify code.

[Asana](https://app.asana.com/0/1204008699308084/1209460354914779)


hgray-instawork:
Deemed unnecessary since each mutation is unique to the navigator type (stack vs. tab)

PR #1112, opened 2025-03-05
refactor(demo/scroll-opacity): use ranges to map position with opacity

flochtililoch:
To increase flexibiility of the component, read position (input) and opacity (output) ranges from XML attribute, and use these to calculate the expected opacity. Update the example to show-case two different use.
Also add the ability to pass styles to the wrapping view.

https://github.com/user-attachments/assets/11c12a96-55a9-46f9-bc2e-1e0968298d69

### DIFF
--- a/demo/schema/scroll-opacity.xsd
+++ b/demo/schema/scroll-opacity.xsd
@@ -3,17 +3,24 @@
   attributeFormDefault="qualified"
   elementFormDefault="qualified"
   targetNamespace="https://hyperview.org/scroll-opacity"
+  xmlns:scroll-opacity="https://hyperview.org/scroll-opacity"
   xmlns:hv="https://hyperview.org/hyperview"
 >
   <xs:import
     namespace="https://hyperview.org/hyperview"
     schemaLocation="hyperview.xsd"
   />
+  <xs:simpleType name="range">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="\[-?[0-9]+,\s?-?[0-9]+\]" />
+    </xs:restriction>
+  </xs:simpleType>
   <xs:element name="scroll-opacity">
     <xs:complexType>
       <xs:sequence>
         <xs:any minOccurs="0" maxOccurs="unbounded" />
       </xs:sequence>
+      <xs:attribute name="style" type="hv:styleList" form="unqualified" />
       <xs:attribute name="context-key" type="xs:token" />
       <xs:attribute name="axis">
         <xs:simpleType>
@@ -23,9 +30,9 @@
           </xs:restriction>
         </xs:simpleType>
       </xs:attribute>
-      <xs:attribute name="distance" type="xs:nonNegativeInteger" />
       <xs:attribute name="duration" type="xs:nonNegativeInteger" />
-      <xs:attribute name="initial-opacity" type="xs:nonNegativeInteger" />
+      <xs:attribute name="scroll-range" type="scroll-opacity:range" />
+      <xs:attribute name="opacity-range" type="scroll-opacity:range" />
     </xs:complexType>
   </xs:element>
 </xs:schema>
--- a/demo/schema/scroll-opacity.xsd
+++ b/demo/schema/scroll-opacity.xsd
@@ -10,9 +10,14 @@
     namespace="https://hyperview.org/hyperview"
     schemaLocation="hyperview.xsd"
   />
-  <xs:simpleType name="range">
+  <xs:simpleType name="scroll-range">
     <xs:restriction base="xs:string">
       <xs:pattern value="\[-?[0-9]+,\s?-?[0-9]+\]" />
+    </xs:restriction>
+  </xs:simpleType>
+  <xs:simpleType name="opacity-range">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="\[[0-9]\.?[0-9]*,\s?[0-9]\.?[0-9]*\]" />
     </xs:restriction>
   </xs:simpleType>
   <xs:element name="scroll-opacity">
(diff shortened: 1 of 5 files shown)


demo/schema/scroll-opacity.xsd
```diff
@@ -3,17 +3,24 @@
...
+  xmlns:scroll-opacity="https://hyperview.org/scroll-opacity"
   xmlns:hv="https://hyperview.org/hyperview"
 >
   <xs:import
     namespace="https://hyperview.org/hyperview"
     schemaLocation="hyperview.xsd"
   />
+  <xs:simpleType name="range">
+    <xs:restriction base="xs:string">
+      <xs:pattern value="\[-?[0-9]+,\s?-?[0-9]+\]" />
```
hgray-instawork, line 15: Can we assert the second number is higher than the first? Otherwise you could have a divide by zero error in `calculateOpacity`

demo/src/Components/ScrollOpacity/utils.ts
```diff
@@ -3,11 +3,48 @@ export const namespaceURI = 'https://hyperview.org/scroll-opacity';
...
+    throw new Error(`Invalid range attribute: ${value}`);
+  }
+};
+
+export const calculateOpacity = (
+  position: number,
+  inputRange: [number, number],
+  outputRange: [number, number],
+) => {
+  const [a, b] = inputRange;
```
hgray-instawork, line 36: Should you exit early if the values are the same?

hgray-instawork (approved):

adamstep:
@flochtililoch can the mapping of the scroll position to opacity be done using Animated interpolate helper? https://reactnative.dev/docs/animatedvalue#interpolate

demo/schema/scroll-opacity.xsd
flochtililoch: good catch, I updated the implementation to cover this case

demo/src/Components/ScrollOpacity/utils.ts
flochtililoch: this is now handled line 40

demo/src/Components/ScrollOpacity/ScrollOpacity.tsx
```diff
@@ -1,16 +1,28 @@
...
+  const opacityRange = getRangeAttr(props.element, 'opacity-range', [0, 1]);
   const duration = getNumberAttr(props.element, 'duration', 0);
-  const initialOpacity = getNumberAttr(props.element, 'initial-opacity', 1);
-  const opacity = useRef(new Animated.Value(initialOpacity)).current;
+  const style = Hyperview.createStyleProp(props.element, props.stylesheets, {
+    ...props.options,
+    styleAttr: 'style',
+  });
+  const opacity = useRef(
+    // Assumes the default scroll position is 0
```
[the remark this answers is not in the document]
flochtililoch: yes - one example covers it (the "title" that appears as a starting scroll value of 30)

demo/schema/scroll-opacity.xsd
```diff
@@ -23,9 +30,9 @@
           </xs:restriction>
         </xs:simpleType>
       </xs:attribute>
-      <xs:attribute name="distance" type="xs:nonNegativeInteger" />
       <xs:attribute name="duration" type="xs:nonNegativeInteger" />
-      <xs:attribute name="initial-opacity" type="xs:nonNegativeInteger" />
+      <xs:attribute name="scroll-range" type="scroll-opacity:range" />
+      <xs:attribute name="opacity-range" type="scroll-opacity:range" />
```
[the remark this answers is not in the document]
flochtililoch: makes sense, I'll update

demo/schema/scroll-opacity.xsd
hgray-instawork: Is it worth asserting in the XML too?

flochtililoch:
> @flochtililoch can the mapping of the scroll position to opacity be done using Animated interpolate helper? https://reactnative.dev/docs/animatedvalue#interpolate

That was my original attempt, but the value that needs to be interpolated with these range is a scroll position, which comes from the context. I believe the `interpolate` method to be only available on `Animated.Value`, which I think would make it a bit more challenging. Let me know if you think there's another way to use it without maintaining an additional state/value in the component.

demo/schema/scroll-opacity.xsd
hgray-instawork: Looks pretty gnarly, I think we can skip it for now since you're protecting against bad data.

demo/schema/scroll-opacity.xsd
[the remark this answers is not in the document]
flochtililoch: 4e021ce4cf41918a1a6b23ad587e22f3f077ab40

demo/schema/scroll-opacity.xsd
flochtililoch: actually there's a use case for the range to have the same start/end values (at least for the scroll range), so that there's no "slope" to apply to the calculated opacity. My assertion in the code just checks whether if the two numbers are equal and returns early.

adamstep:
> > @flochtililoch can the mapping of the scroll position to opacity be done using Animated interpolate helper? https://reactnative.dev/docs/animatedvalue#interpolate
> 
> That was my original attempt, but the value that needs to be interpolated with these range is a scroll position, which comes from the context. I believe the `interpolate` method to be only available on `Animated.Value`, which I think would make it a bit more challenging. Let me know if you think there's another way to use it without maintaining an additional state/value in the component.

Ah that makes sense, yes it only works on `Animated.Value`.

PR #1123, opened 2025-03-24
chore(hv-doc): create component and lift state

hgray-instawork:
Created a new `hv-doc` component. Lift the state and this.doc out of both `hv-route` and `hv-screen` into instances of `hv-doc`.

[Asana](https://app.asana.com/0/1204008699308084/1209669995625339/f)

### DIFF
--- a/src/core/components/hv-doc/context.ts
+++ b/src/core/components/hv-doc/context.ts
@@ -1,0 +1,9 @@
+import { StateContextProps } from './types';
+import { createContext } from 'react';
+
+export const StateContext = createContext<StateContextProps>({
+  getLocalDoc: () => null,
+  getState: () => ({}),
+  setLocalDoc: () => ({}),
+  setState: () => ({}),
+});
--- a/src/core/components/hv-doc/context.ts
+++ b/src/core/components/hv-doc/context.ts
@@ -3,7 +3,7 @@ import { createContext } from 'react';
 
 export const StateContext = createContext<StateContextProps>({
   getLocalDoc: () => null,
-  getState: () => ({}),
+  getScreenState: () => ({}),
   setLocalDoc: () => ({}),
-  setState: () => ({}),
+  setScreenState: () => ({}),
 });
--- a/src/core/components/hv-doc/index.tsx
+++ b/src/core/components/hv-doc/index.tsx
@@ -24,13 +24,18 @@ const HvDoc = (props: Props) => {
         getLocalDoc: () => localDoc.current,
         getState: () => state,
         setLocalDoc: (doc: Document | null) => {
-          if (doc !== null) {
+          if (doc !== undefined) {
             localDoc.current = doc;
           }
         },
         // * Override the state to clear the doc when an element is passed
-        setState: (newState: ScreenState) =>
-          setState({ ...newState, doc: props.element ? null : newState.doc }),
+        setState: (newState: ScreenState) => {
+          setState({
+            ...state,
+            ...newState,
+            doc: props.element ? null : newState.doc,
+          });
+        },
       }}
     >
       <>{props.children}</>
--- a/src/core/components/hv-doc/index.tsx
+++ b/src/core/components/hv-doc/index.tsx
@@ -38,7 +38,7 @@ const HvDoc = (props: Props) => {
         },
       }}
     >
-      <>{props.children}</>
+      {props.children}
     </StateContext.Provider>
   );
 };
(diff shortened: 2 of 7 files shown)


src/core/components/hv-doc/context.ts
```diff
@@ -0,0 +1,9 @@
+import { StateContextProps } from './types';
+import { createContext } from 'react';
+
+export const StateContext = createContext<StateContextProps>({
```
flochtililoch, line 4: Should this be `DocContext`? Also, all contexts currently live under src/contexts, I think we should follow the same pattern to remain consistent.

src/core/components/hv-doc/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const [state, setState] = useState<ScreenState>({
+    doc: undefined,
+    error: undefined,
+  });
+
+  return (
+    <StateContext.Provider
+      value={{
+        getLocalDoc: () => localDoc.current,
+        getState: () => state,
```
flochtililoch, line 25: is there any advantage in using proxy methods as opposed to direct references to the objects?

src/core/components/hv-doc/types.ts
```diff
@@ -0,0 +1,13 @@
+import { ScreenState } from 'hyperview/src/types';
+
+export type Props = {
+  children?: React.ReactNode;
+  element?: Element;
+};
+
+export type StateContextProps = {
+  getLocalDoc: () => Document | null;
+  getState: () => ScreenState;
```
flochtililoch, line 10: nit: as this is the public API of the context, `getState` / `setState` are a bit too generic. Maybe including `doc` or `screen` in the key names could help make it more obvious what is being accessed when these are called.

src/core/components/hv-doc/context.ts
hgray-instawork: We already have a `DocContext` [here](https://github.com/Instawork/hyperview/blob/4f49ce039e26b7be991e9e1af1e755b6818579bc/src/contexts/index.ts#L23). This is probably interim naming as I hope to incorporate the getDoc/setDoc functionality into `hv-doc` eventually.

Regarding location, I think we had discussed eventually moving them with their associated component, as was done for [the scroll component](https://github.com/Instawork/hyperview/blob/4f49ce039e26b7be991e9e1af1e755b6818579bc/src/core/components/scroll/context.tsx#L4). If that isn't the case here, I'll move this into the /contexts directory.

src/core/components/hv-doc/types.ts
hgray-instawork: good call. https://github.com/Instawork/hyperview/pull/1123/commits/829fce9bf29532b599e3d18e72419f2ca00c7085

src/core/components/hv-doc/index.tsx
hgray-instawork: Strangely, the state doesn't work when passed directly. We had the same problem when externalizing the `onUpdate` functionality [here](https://github.com/Instawork/hyperview/blob/829fce9bf29532b599e3d18e72419f2ca00c7085/src/core/components/hv-screen/index.js#L307).

src/core/components/hv-screen/index.js
```diff
@@ -348,5 +324,19 @@ export default class HvScreen extends React.Component {
   };
 }
 
+export default function HvScreen(props) {
```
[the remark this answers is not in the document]
hgray-instawork: I'll have a look

src/core/components/hv-doc/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+        setScreenState: (newState: ScreenState) => {
+          setState({
+            ...state,
+            ...newState,
+            doc: props.element ? null : newState.doc,
+          });
+        },
+      }}
+    >
+      <>{props.children}</>
```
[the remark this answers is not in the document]
hgray-instawork: Not needed. Removed: https://github.com/Instawork/hyperview/pull/1123/commits/2be75a25c28f4adf97d76256061e8b295723070f

src/core/components/hv-screen/index.js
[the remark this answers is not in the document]
hgray-instawork: It was part of an earlier idea. No longer needed, thanks for the catch. https://github.com/Instawork/hyperview/pull/1123/commits/5b7d9218fbaf0295d0be8cdfae9133de1993c5af

flochtililoch (approved):

PR #1132, opened 2025-03-27
chore(performance): memoize context values

hgray-instawork:
Based on suggestions from Cursor, updating the values that are passed into context providers. In most cases this was using a combination of `useMemo` and `useCallback`. For the class-based components, this involves using a get function to provide the value.

[Asana](https://app.asana.com/0/1204008699308084/1209802590586962/f)

### DIFF
--- a/demo/src/Components/BottomSheet/BottomSheet.tsx
+++ b/demo/src/Components/BottomSheet/BottomSheet.tsx
@@ -407,25 +407,29 @@ const BottomSheet = (props: HvComponentProps) => {
     </GestureDetector>
   );
 
+  const setContentSectionHeight = useCallback(
+    () => (index: number, contentSectionHeight: number) => {
+      setContentSectionHeights(prevHeights => {
+        const currentHeights = [...prevHeights];
+        currentHeights[index] = contentSectionHeight;
+        return currentHeights;
+      });
+    },
+    [],
+  );
+
+  const contextValue = useMemo(() => ({ setContentSectionHeight }), [
+    setContentSectionHeight,
+  ]);
+
   if (height === 0) {
     // Pre-render the content to trigger the layout calculation
     // The content should not be visible as we start with an opacity of 0
     return content;
   }
 
-  const setContentSectionHeight = (
-    index: number,
-    contentSectionHeight: number,
-  ) => {
-    setContentSectionHeights(prevHeights => {
-      const currentHeights = [...prevHeights];
-      currentHeights[index] = contentSectionHeight;
-      return currentHeights;
-    });
-  };
-
   return (
-    <BottomSheetContext.Provider value={{ setContentSectionHeight }}>
+    <BottomSheetContext.Provider value={contextValue}>
       <Modal onShow={animateOpen} transparent visible={visible}>
         <GestureHandlerRootView style={{ flex: 1 }}>
           <Overlay
--- a/src/core/components/hv-root/index.tsx
+++ b/src/core/components/hv-root/index.tsx
@@ -562,33 +562,33 @@ export default class Hyperview extends PureComponent<Types.Props> {
     }
   };
 
+  getContextValue = () => ({
+    behaviors: this.props.behaviors,
+    components: this.props.components,
+    elementErrorComponent: this.props.elementErrorComponent,
+    entrypointUrl: this.props.entrypointUrl,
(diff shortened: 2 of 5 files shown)


src/core/components/hv-root/index.tsx
```diff
@@ -562,33 +562,33 @@ export default class Hyperview extends PureComponent<Types.Props> {
...
-              onError: this.props.onError,
-              onParseAfter: this.props.onParseAfter,
-              onParseBefore: this.props.onParseBefore,
-              onRouteBlur: this.props.onRouteBlur,
-              onRouteFocus: this.props.onRouteFocus,
-              onUpdate: this.onUpdate,
-              reload: this.reload,
-            }}
-          >
+          <NavContexts.Context.Provider value={this.getContextValue()}>
```
flochtililoch, line 591: I don't think this will make a difference: the function `this.getContextValue()` will be called on every render, and will always return a new object, which in turn will make the `value` prop change

src/core/components/hv-screen/index.js
```diff
@@ -297,11 +301,7 @@ export default class HvScreen extends React.Component {
     }
 
     return (
-      <Contexts.DocContext.Provider
-        value={{
-          getDoc: () => this.doc,
-        }}
-      >
+      <Contexts.DocContext.Provider value={this.getDocContextValue()}>
```
flochtililoch, line 304: same here

src/core/components/scroll/context.tsx
```diff
@@ -12,17 +18,20 @@ export const Context = createContext<{
...
+  const updateOffset = useCallback((viewId: string, offset: ScrollOffset) => {
+    setOffsets(prev => ({
+      ...prev,
       [viewId]: offset,
-    });
-  };
+    }));
+  }, []);
+
+  const contextValue = useMemo(() => ({ offsets, updateOffset }), [
```
flochtililoch, line 28: were you able to test whether the scroll demos are still working with this change?

demo/src/Components/BottomSheet/BottomSheet.tsx
```diff
@@ -407,25 +407,29 @@ const BottomSheet = (props: HvComponentProps) => {
...
 
+  const setContentSectionHeight = useCallback(
+    () => (index: number, contentSectionHeight: number) => {
+      setContentSectionHeights(prevHeights => {
+        const currentHeights = [...prevHeights];
+        currentHeights[index] = contentSectionHeight;
+        return currentHeights;
+      });
+    },
+    [],
```
flochtililoch, line 418: shouldn't this require dependencies to be added here?

src/core/components/scroll/context.tsx
hgray-instawork: Yes, I tested the scroll opacity demo

src/core/components/hv-root/index.tsx
hgray-instawork: I'd have thought so too. Cursor suggested these changes with this info:

"I've added a getDocContextValue class method that returns the memoized context value object. Since this is a class component, we can't use useMemo, but the class method will maintain the same reference between renders unless this.doc changes. This should resolve the warning about the context value changing on every render."

flochtililoch: hmm, I don't think that's correct. You can probably test with the following:
```
console.log(this.getContextValue() === this.getContextValue())
```

demo/src/Components/BottomSheet/BottomSheet.tsx
hgray-instawork: I believe there are no dependencies. The only external reference is the `setContentSectionHeights` state setter, which doesn't need to be passed. Everything else is handled either as input args or the functional state setter which provides the previous state.

src/core/components/hv-root/index.tsx
hgray-instawork: You are correct. Revised approach captures the value in constructor and updates in `componentDidUpdate` when any values change. https://github.com/Instawork/hyperview/pull/1132/commits/de5b0485eb0a5c8f133d50e31469a6cc52181b4b

flochtililoch (approved):

flochtililoch (approved):

hgray-instawork:
We've decided to abandon this approach.

PR #1134, opened 2025-03-28
feat(demo): `safe-area` element

flochtililoch:
Add a new element that acts as a safe area container, then update the demo app server code to make use of it.

See commits breakdown for details

| iOS | Android |
|--|--|
| iOS | Android |


hgray-instawork (approved):
Nice!

PR #1139, opened 2025-04-01
feat(unload): add support for `unload` trigger

hgray-instawork:
Adding a new `unload` trigger which is triggered when a `view` is unmounted.

[Asana](https://app.asana.com/0/1204008699308084/1209860719801774/f)

https://github.com/user-attachments/assets/bf0f0079-ff02-41bf-903e-be9d6de9d36e

### DIFF
--- a/docs/reference_behavior_attributes.md
+++ b/docs/reference_behavior_attributes.md
@@ -105,6 +105,16 @@ These elements support the `load` trigger:
 - [`<text>`](/docs/reference_text)
 - [`<image>`](/docs/reference_image)
 
+### `unload`
+
+Triggers when the element is unloaded in the screen.
+
+> Elements with `hide="true"` are not unloaded in the screen. This means any hidden element (or child element) with the `unload` trigger will not be executed.
+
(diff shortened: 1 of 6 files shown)


docs/reference_behavior_attributes.md
```diff
@@ -105,6 +105,16 @@ These elements support the `load` trigger:
 - [`<text>`](/docs/reference_text)
 - [`<image>`](/docs/reference_image)
 
+### `unload`
+
+Triggers when the element is unloaded in the screen.
+
+> Elements with `hide="true"` are not unloaded in the screen. This means any hidden element (or child element) with the `unload` trigger will not be executed.
```
adamstep, line 112: Maybe to make this clear, we should say "Elements with hide=true are never loaded. Thus, they are not unloaded and will not trigger an `unload` behavior."

hgray-instawork:
We are not ready to add this feature now but will keep this around for possible later use. See [Asana ticket](https://app.asana.com/0/1204008699308084/1209860719801774/f) for further discussion.

PR #1158, opened 2025-05-06
chore(hv-doc): internalize loading logic

hgray-instawork:
Remove legacy load triggers from hv-route and hv-screen and internalize into hv-doc.
- Removed use of "needsLoad"
- Removed unnecessary lifecycle functions
- Verified current functionality including load, reload, append, modals, etc.

See commits for breakdown

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1209875001265866?focus=true)

### DIFF
--- a/src/core/components/hv-doc/index.tsx
+++ b/src/core/components/hv-doc/index.tsx
@@ -177,9 +177,6 @@ const HvDoc = (props: Props) => {
     ],
   );
 
-  const needsLoadCallback = useRef(() => {
-    // Noop
-  });
   // Monitor props.url changes
   useEffect(() => {
     if (
@@ -272,8 +269,10 @@ const HvDoc = (props: Props) => {
       getNavigation,
       getOnUpdate: () => onUpdate,
       getState: getScreenState,
-      setNeedsLoad: needsLoadCallback.current,
       setState: setScreenState,
+      updateUrl: (url: string) => {
+        setLoadingUrl(url);
+      },
     };
 
     return {
@@ -282,9 +281,6 @@ const HvDoc = (props: Props) => {
       onUpdate,
       onUpdateCallbacks: onUpdateCallbacksRef.current,
       reload,
-      setNeedsLoadCallback: (callback: () => void) => {
-        needsLoadCallback.current = callback;
-      },
       setScreenState,
     };
   }, [
--- a/src/core/components/hv-doc/index.tsx
+++ b/src/core/components/hv-doc/index.tsx
@@ -104,13 +104,14 @@ const HvDoc = (props: Props) => {
         handleError(new HvDocError('No URL provided'), targetUrl);
         return;
       }
+      const params = props.route?.params ?? {};
 
       try {
-        if (props.route?.params.delay) {
+        if (params.delay) {
           const delay =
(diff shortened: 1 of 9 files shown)


src/core/components/hv-doc/index.tsx
```diff
@@ -152,14 +158,51 @@ const HvDoc = (props: Props) => {
...
+    props.element,
+    props.route?.params.needsSubStack,
+    props.url,
+    state.url,
+  ]);
+
+  // Load the url when the state is updated
+  useEffect(() => {
+    if (loadingUrl) {
+      loadUrl(loadingUrl);
```
flochtililoch, line 202: is there any risk for this to be called more than what we need when `loadUrl` callback changes?

src/core/components/hv-doc/index.tsx
```diff
@@ -226,27 +269,24 @@ const HvDoc = (props: Props) => {
       getNavigation,
       getOnUpdate: () => onUpdate,
       getState: getScreenState,
-      setNeedsLoad: needsLoadCallback.current,
       setState: setScreenState,
+      updateUrl: (url: string) => {
+        setLoadingUrl(url);
+      },
```
flochtililoch, line 275: minor: can we pass a reference of the function instead of creating an anonymous wrapper?
```suggestion
      updateUrl: setLoadingUrl,
      },
```

src/core/components/hv-route/index.tsx
```diff
@@ -414,7 +343,13 @@ function HvRouteFC(props: Types.Props) {
   const backContext = useContext(BackBehaviorContext);
   const docContext = useContext(Contexts.DocContext);
 
-  const url = getRouteUrl(props, navigationContext);
+  const url =
+    props.navigation === undefined
+      ? navigationContext.entrypointUrl
+      : NavigatorService.cleanHrefFragment(
+          props.route?.params?.url || navigationContext.entrypointUrl,
+        );
```
flochtililoch, line 351: sanity check: the logic has changed to always call `NavigatorService.cleanHrefFragment` instead of conditionally returning `undefined` when `props.route?.params?.url` is falsy - is this intentional or could it have any unintended side effect?

flochtililoch (approved):

src/core/components/hv-doc/index.tsx
hgray-instawork: I was just looking at that. I think I can improve the state management here.

src/core/components/hv-doc/index.tsx
hgray-instawork: Will take a look

src/core/components/hv-doc/index.tsx
hgray-instawork: https://github.com/Instawork/hyperview/pull/1158/commits/b8a9d38a48b44ca4b826ef5f579904c21e659532

src/core/components/hv-doc/index.tsx
hgray-instawork: https://github.com/Instawork/hyperview/pull/1158/commits/55aa48a77c984e0f144c0ba32f08ba61e0ea6733

PR #1161, opened 2025-05-15
fix: njk templates linter

flochtililoch:
- Fix template linter to search for templates under the right path
- Lint templates that were missed since being relocated
- Make script more robust

### DIFF
--- a/demo/backend/ui/styling/simple-ui-kit/_uikit.xml.njk
+++ b/demo/backend/ui/styling/simple-ui-kit/_uikit.xml.njk
@@ -25,12 +25,12 @@
 <style id="btn-nav" height="44" minWidth="44" backgroundColor="#ffffff" alignItems="center" justifyContent="center"/>
 <style id="btn-nav-container">
   <modifier pressed="true">
-    <style backgroundColor="#ffffff"/>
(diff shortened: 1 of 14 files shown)


demo/backend/ui/styling/simple-ui-kit/_uikit.xml.njk
```diff
@@ -691,17 +691,17 @@
 <style id="shadow-footer" backgroundColor="white" shadowRadius="16" shadowOffsetY="-3.6"/>
 <style id="shadow-pressable">
   <modifier pressed="true">
-    <style elevation="1" shadowOffsetY="1" shadowRadius="1"/>
+  <style elevation="1" shadowOffsetY="1" shadowRadius="1"/>
```
adamstep, line 694: Just checking, is this the intended style? Does it have to do with self-closing tags?

adamstep (approved):

demo/backend/ui/styling/simple-ui-kit/_uikit.xml.njk
flochtililoch: I think the linter was designed primarily for html and might not handle every tags correctly

PR #1164, opened 2025-05-19
chore(behaviors): log behaviors only when run

hgray-instawork:
Behavior logging in the form of `core/components/hyperview | [behavior] | action: add-styles | <behavior trigger...` is being triggered regardless of whether the behavior was run or was cancelled (example, "once='true'" and already ran). To solve this, a new `onUpdateComplete` callback is added to the `onUpdate` flow. This is called with a "success" value of "true" when the update completed successfully, otherwise a "false" is sent. The behavior service can now wait for this callback and determine whether the log should be emitted.

See commits for breakdown.

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1205585476216008?focus=true)

---
  - https://app.asana.com/0/0/1209916822029079

### DIFF
--- a/src/core/components/hv-root/index.tsx
+++ b/src/core/components/hv-root/index.tsx
@@ -392,53 +392,57 @@ export default class Hyperview extends PureComponent<Types.Props> {
           onUpdateCallbacks,
           networkRetryAction,
           networkRetryEvent,
-        ).then(newElement => {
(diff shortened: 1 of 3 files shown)


hgray-instawork:
This approach is being put aside for now. We will remove the behavior logs completely and re-add them once we have a better behavior sequence.

PR #1169, opened 2025-05-21
chore(hv-element): create component for element rendering

hgray-instawork:
Improving performance by creating a new `hv-element` component as a replacement for Render.renderElement. This allows memoizing the options and props of the component to reduce unneeded re-renders.

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1210344460236268?focus=true)

### DIFF
--- a/src/core/components/hv-element/index.tsx
+++ b/src/core/components/hv-element/index.tsx
@@ -1,0 +1,157 @@
+import * as InlineContext from 'hyperview/src/services/inline-context';
+import * as Logging from 'hyperview/src/services/logging';
+import * as Namespaces from 'hyperview/src/services/namespaces';
+import { LOCAL_NAME, NODE_TYPE } from 'hyperview/src/types';
+import React, { useMemo } from 'react';
+import type { HvComponentProps } from 'hyperview/src/types';
+
+export default (
+  props: HvComponentProps,
+): React.ReactElement<HvComponentProps> | null | string => {
+  if (!props.element) {
+    return null;
+  }
+
+  if (props.element.nodeType === NODE_TYPE.ELEMENT_NODE) {
+    // Hidden elements don't get rendered
+    if (props.element.getAttribute('hide') === 'true') {
+      return null;
+    }
+  }
+  if (props.element.nodeType === NODE_TYPE.COMMENT_NODE) {
+    // XML comments don't get rendered.
+    return null;
+  }
+  if (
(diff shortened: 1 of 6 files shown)


src/core/components/hv-element/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+      props.element.namespaceURI &&
+      props.element.localName
+    ) {
+      return props.options.componentRegistry?.getComponent(
+        props.element.namespaceURI,
+        props.element.localName,
+      );
+    }
+    return undefined;
+  }, [props.element, props.options.componentRegistry]);
```
flochtililoch, line 84: minor: I think it's work extracting the attributes used in this context and apply memoization on them - `element` could change many times while the attributes used here will remain identical, but this would cause a re-compute.

src/core/components/hv-element/index.tsx
```diff
@@ -0,0 +1,157 @@
+import * as InlineContext from 'hyperview/src/services/inline-context';
+import * as Logging from 'hyperview/src/services/logging';
+import * as Namespaces from 'hyperview/src/services/namespaces';
+import { LOCAL_NAME, NODE_TYPE } from 'hyperview/src/types';
+import React, { useMemo } from 'react';
+import type { HvComponentProps } from 'hyperview/src/types';
+
+export default (
+  props: HvComponentProps,
+): React.ReactElement<HvComponentProps> | null | string => {
```
flochtililoch, line 10: can the return value be
```suggestion
): JSX.Element | null => {
```

flochtililoch (commented):
In order to deduplicate the code, can we also "re-wire" the existing `renderElement` method so that it instantiate the new `HvElement` component?

src/core/components/hv-element/index.tsx
hgray-instawork: https://github.com/Instawork/hyperview/pull/1169/commits/8df786f2be690b598e2b5d5c05f93f69276b8fae

src/core/components/hv-element/index.tsx
hgray-instawork: https://github.com/Instawork/hyperview/pull/1169/commits/f236348ea468249e2bfadbcd67c6a1a7a566b413

hgray-instawork:
> In order to deduplicate the code, can we also "re-wire" the existing `renderElement` method so that it instantiate the new `HvElement` component?

I'll delay this suggestion for now until we migrate the remaining render functions.

PR #1171, opened 2025-05-23
chore(render): simplify render.renderElement

hgray-instawork:
Consolidate the logic between `src/services/render` and the new `hv-element` component.
- Created a new utility to abstract the `null` render logic and logging
- Update render service to use logic and then return the component
- Update `hv-element` to use utility for logic and logging
- Export `hv-element` for access outside Hyperview
- Confirmed external use of `Hyperview.renderElement` with new logic still working (using existing nav-back demo)
- Update `nav-back` demo to use component instead of function

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1210368849843731?focus=true)

### DIFF
--- a/src/core/components/hv-element/index.tsx
+++ b/src/core/components/hv-element/index.tsx
@@ -7,62 +7,44 @@ import type { HvComponentProps } from 'hyperview/src/types';
 import { isRenderableElement } from '../../utils';
 
 export default (props: HvComponentProps): JSX.Element | null | string => {
-  const nodeType = useMemo(() => {
-    return props.element?.nodeType;
-  }, [props.element]);
-
-  const localName = useMemo(() => {
-    return props.element?.localName;
-  }, [props.element]);
-
-  const namespaceURI = useMemo(() => {
-    return props.element?.namespaceURI;
-  }, [props.element]);
-
-  const options = useMemo(() => {
-    return props.options || {};
-  }, [props.options]);
+  // eslint-disable-next-line react/destructuring-assignment
+  const { element, onUpdate, options, stylesheets } = props;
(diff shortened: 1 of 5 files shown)


src/core/components/hv-element/index.tsx
```diff
@@ -4,151 +4,88 @@ import * as Namespaces from 'hyperview/src/services/namespaces';
 import { LOCAL_NAME, NODE_TYPE } from 'hyperview/src/types';
 import React, { useMemo } from 'react';
 import type { HvComponentProps } from 'hyperview/src/types';
+import { isRenderableElement } from '../../utils';
```
flochtililoch, line 7: can you use an absolute import?

src/core/components/hv-element/index.tsx
```diff
@@ -4,151 +4,88 @@ import * as Namespaces from 'hyperview/src/services/namespaces';
...
+      stylesheets,
     };
-  }, [
-    formattingContext,
-    props.element,
-    props.onUpdate,
-    props.options,
-    props.stylesheets,
-  ]);
+  }, [formattingContext, options, element, onUpdate, stylesheets]);
```
flochtililoch, line 37: minor/code style: I tend to sort dependency list alphabetically, so we have a predictable way to add new dependencies (instead of just adding at the end of the list).

hgray-instawork: Good catches, both. I thought I caught all of these. I think we need to give Cursor some more guidance. It keeps re-adding these after I manually change it. https://github.com/Instawork/hyperview/pull/1171/commits/f4fd6ec6cfc15b00cd1534bf20d78fcaa1341056

PR #1172, opened 2025-05-27
chore(render): simplify child rendering

hgray-instawork:
Use a context to memoize the props used to create the children list used for creating elements to reduce the number of times the array is generated.
- Created a new `children-context` with a provider to memoize props and the child array
- Update the uses of `Render.renderChildren()`
- The provider accepts an optional `childNodes` to allow externally filtered content to be processed as a replacement for `renderChildNodes`.

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1210344460236290?focus=true)


hgray-instawork:
Closing to go in a different direction

PR #1173, opened 2025-05-28
chore(render): Replace renderChildren with component

hgray-instawork:
Creating a component which can replace the use of the render service `renderChildren`. Provides a mapping of child nodes to `<HvElement>` components.

- [Created the new component](https://github.com/Instawork/hyperview/commit/0ece9f6502dba35eade51d8097aee9df1ac439b7)
- [Switched four element types from ts to tsx](https://github.com/Instawork/hyperview/commit/dd518da6aa03fe22a3b71d587d7c5971015ed108)
- Replaced `React.createElement` implementations with declarative instances [using the new component to render children](https://github.com/Instawork/hyperview/commit/7c73c9bd2f3f45f3ee83ab77c3a9730d4f60928b)
- Updated the sticky header logic to use the rendered children and [added a verification of a valid `getAttribute`](https://github.com/Instawork/hyperview/commit/65d8b20ece3b2dba561752ab8911eb666faa8046)

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1210344460236290?focus=true)

---
  - https://app.asana.com/0/0/1210344460236290


src/components/hv-view/index.tsx
```diff
@@ -173,63 +173,81 @@ export default class HvView extends PureComponent<HvComponentProps> {
...
-      return React.createElement(
-        KeyboardAvoidingView,
-        { ...this.getCommonProps(), behavior: 'position' },
-        ...children,
+      return (
+        <KeyboardAvoidingView
+          {...{
+            ...this.getCommonProps(),
+            behavior: 'position',
+          }}
```
flochtililoch, line 237:
```suggestion
          {...this.getCommonProps()}
          behavior="position"
```

src/components/hv-view/index.tsx
```diff
@@ -173,63 +173,81 @@ export default class HvView extends PureComponent<HvComponentProps> {
...
-        {
-          element: this.props.element,
-          ...this.getCommonProps(),
-          ...this.getScrollViewProps(children),
-        },
-        ...children,
+      return (
+        <ScrollView
+          {...{
+            element: this.props.element,
```
flochtililoch, line 219:
```suggestion
          element={this.props.element}
          {...{
```

flochtililoch (commented):
minor/naming - would `otherProps` be better than `rest`?

src/core/hyper-ref/index.tsx
```diff
@@ -466,9 +466,14 @@ export const addHref = (
     return component;
   }
 
-  return React.createElement(
-    HyperRef,
-    { element, onUpdate, options, stylesheets },
-    ...Render.renderChildren(element, stylesheets, onUpdate, options),
+  return (
+    <HyperRef {...{ element, onUpdate, options, stylesheets }}>
```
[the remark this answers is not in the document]
hgray-instawork: https://github.com/Instawork/hyperview/pull/1173/commits/ab30fd0908df54de12d82205006290a3c71e0c20 Note: HyperRef doesn't seem to have need for children being passed in and is not handling them or rendering them. The child element has been removed from the implmentation.

src/components/hv-view/index.tsx
hgray-instawork: https://github.com/Instawork/hyperview/pull/1173/commits/7f5fc6de7286ac9912a9253c1decbd00d578b289 This change also required adding `children` to the props

src/components/hv-view/index.tsx
```diff
@@ -173,11 +173,10 @@ export default class HvView extends PureComponent<HvComponentProps> {
       }
     }
 
-    const children = Render.renderChildren(
-      this.props.element,
-      this.props.stylesheets,
-      this.props.onUpdate as HvComponentOnUpdate,
-      {
+    const children = HvChildren({
```
flochtililoch, line 176: can we use the JSX syntax here?

flochtililoch (approved):

src/components/hv-view/index.tsx
hgray-instawork: Unfortunately we can't as we need to iterate over the array of children to check their attributes. When we use the JSX element, the children are embedded inside the <HvChildren> component. The only alternative I found to this was to duplicate the map code from `hv-children` directly in `hv-view` but that seemed less ideal

PR #1183, opened 2025-06-09
chore(performance): optimize style props

hgray-instawork:
Optimizing style and props creation by replacing `createProps` and `createStyleProp` with hooks. After converting several class functions into functional components, the hooks were added. Additional optimizations include:
- destructuring props and options to help with memoization
- improved overall memoization through use of `useMemo` and `useCallback`. Grouped options and values together where possible to reduce overhead.

See commits for full breakdown.

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1210344460236267?focus=true)


flochtililoch (commented):
I'm not sure if you did a pass already to address what we discussed yesterday, but re-adding my feedback here (with some additional feedback):

I think we should remove the memoization for:
- simple attribute getters or objects accessor
- any low usage components (i.e. all form elements, webview)
Right now, memoization for these adds a lot of the verbosity in code, and I doubt they provide a real gain. I think memoization makes sense in some places, for heavy usage components like text, or view, and when there's computation involved (for loop, object creation etc.).

Going one step further in this migration, could we:
- define a return type for the new hooks, and remove use of `any` when these are used
- convert `createTestProps` to `useTestProps`, have it do internal memoization on element
- remove the code within legacy `create*` functions and have these delegate to their equivalent hook function (granted we would also need to silence the ESLint warning about not being in a React component)

And as future tasks (low priority but we should track):
- I see HvTextField disables rules of hooks linting rule, or does unnecessary re-creation of objects (`elementProps` is spread into `p`, which is spread again when applied to the component instance)
- I noted many components (`hv-switch`, `hv-select-single`) were reading attributes directly from `element`, but we're also able to get these attributes when calling `useProps`, so we should probably refactor to simplify and remove duplication.

PR #1187, opened 2025-06-13
chore(section-list): use FlatList for SectionList

hgray-instawork:
Example of using a react-native `FlatList` in place of `SectionList` for our `hv-section-list` component.

All functionality is retained. Code was able to be simplified especially around the scrolling behaviors.

>This PR isn't necessarily for merging, just an example of what is possible with FlatList.

| Section List | Flat List |
| -- | -- |
| s-l-sticky | f-l-sticky |
| s-l-scroll | f-l-scroll |

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1210525847101749?focus=true)

### DIFF
--- a/demo/backend/advanced/case-studies/basic-forms/index.xml.njk
+++ b/demo/backend/advanced/case-studies/basic-forms/index.xml.njk
@@ -22,6 +22,7 @@ hv_button_behavior: "back"
       />
     </view>
     <text
+      id="basic-forms-submit"
       action="replace"
       delay="500"
       href="/hyperview/public/advanced/case-studies/basic-forms/submit.xml"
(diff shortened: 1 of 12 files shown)


flochtililoch (commented):
I see a difference in behavior between the two implementation: the sticky header pushed the previous one with SectionList, while it overlap with the previous one with FlatList. Is this something we can fine tune?

hgray-instawork:
> I see a difference in behavior between the two implementation: the sticky header pushed the previous one with SectionList, while it overlap with the previous one with FlatList. Is this something we can fine tune?


I see the same "push up" animation in my clean react-native testbed. I'm not sure if the difference here is related to:
- the use of the [scroll component wrapper](https://github.com/Instawork/hyperview/blob/f4361f2e0fb6819d84423cfca843d0fbd4d0e04c/src/core/components/scroll/index.tsx#L12)
- the different version of expo or react-native in my test project
- some other style differences

I'll poke around some more and see if I can get them to match

PR #1201, opened 2025-06-20
chore(text): convert to Functional Component

hgray-instawork:
Convert `hv-text` to Functional Component. Confirm no degradation in render count.

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1210601789543021?focus=true)

### DIFF
--- a/src/elements/hv-text/index.tsx
+++ b/src/elements/hv-text/index.tsx
@@ -11,40 +11,38 @@ import { addHref } from 'hyperview/src/core/components/hyper-ref';
 import { createProps } from 'hyperview/src/services';
 
 const HvText = (props: HvComponentProps) => {
-  const Component = () => {
-    const componentProps = createProps(
-      props.element,
-      props.stylesheets,
-      props.options,
-    );
+  // eslint-disable-next-line react/destructuring-assignment
+  const { element, onUpdate, options, stylesheets } = props;
+  const { skipHref } = options || {};
+
+  const Component = (p: { element: Element }) => {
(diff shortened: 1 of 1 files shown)


src/elements/hv-text/index.tsx
```diff
@@ -4,55 +4,50 @@ import type {
...
 
-  static localName = LOCAL_NAME.TEXT;
-
-  Component = () => {
-    const props = createProps(
-      this.props.element,
-      this.props.stylesheets,
-      this.props.options,
-    );
+  const Component = (p: { element: Element }) => {
```
flochtililoch, line 18: was it necessary to change the interface of the component?

hgray-instawork: Without this change, I get a TS warning:

'element.getAttribute' is missing in props validation eslint[react/prop-types](https://github.com/yannickcr/eslint-plugin-react/tree/master/docs/rules/prop-types.md)

Would you rather I revert the change and ignore the warning?

flochtililoch: Actually one more issue: we're re-creating the component on every render, but previously it was an instance method. I assume it should be wrapped in a `useCallback`.

hgray-instawork: Adding the callback resolved the need for altering the component signature. https://github.com/Instawork/hyperview/pull/1201/commits/94a2a6512ec373be99f94a486e28e9d86c8db5e7

PR #1211, opened 2025-06-24
chore(doc-context): improve context implementation

hgray-instawork:
Improve the implementation of the context using the `Consumer` pattern.

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1210620686213505?focus=true)

### DIFF
--- a/src/elements/hv-doc/index.ts
+++ b/src/elements/hv-doc/index.ts
@@ -1,2 +1,2 @@
 export { default } from './hv-doc';
-export { Consumer as StateConsumer, useHvDoc } from './context';
+export { Consumer as HvDocContext, useHvDocContext } from './context';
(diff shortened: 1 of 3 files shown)


src/elements/hv-doc/index.ts
```diff
@@ -1,2 +1,2 @@
 export { default } from './hv-doc';
-export { Context as StateContext } from './context';
+export { Consumer as StateConsumer, useHvDoc } from './context';
```
flochtililoch, line 2: naming: what do you think if we were to match the hook and the consumer component (since they are equivalent for FC / Class components).
```suggestion
export { Consumer as HvDocContext, useHvDocContext } from './context';
```

hgray-instawork: Yeah, that makes sense. I'll go over the others too and see if they are matching.

flochtililoch (approved):

PR #1217, opened 2025-06-26
chore(hv-screen): remove loading features

hgray-instawork:
`hv-screen` is no longer responsible for performing loading and therefore does not need to handle the preloadScreen functionality or cache maintenance. These functions are performed by `hv-doc`. Removing all of these greatly simplifies the `hv-screen` implementation and reduces the number of props that need to be passed in.

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1209875001265863?focus=true)


flochtililoch (approved):
nice clean up!

PR #1218, opened 2025-06-26
chore(hyper-ref): replace hyper-ref injection

hgray-instawork:
Replace the way `hyper-ref` is injected into components

- Create a new util `needsHyperRef` which allows replacing the use of `addHref`
- Centralize `hyper-ref` implementation into `hv-element`
- Add new `supportsHyperRef` static to indicate which components need to support `hyper-ref`

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1210645181810282?focus=true)


flochtililoch (approved):
neat!

hgray-instawork:
@ehenighan, could you confirm how your custom components are set up? Do your components embed `HvView` directly into them?  I'm trying to make sure I correctly am able to reproduce your use case.

ehenighan:
All of which is to say, we weren't trying to solve any specific *problem* as such - it just happened to be the way we built the component out a while ago

hgray-instawork:
> All of which is to say, we weren't trying to solve any specific _problem_ as such - it just happened to be the way we built the component out a while ago

Thanks @ehenighan, I've merged in the documentation updates which should hopefully improve things for people going forward.

PR #1224, opened 2025-07-03
fix(demo): various broken examples

flochtililoch:
Some examples stopped  behavior code updates (TBC which one, we might want to fix this in Hyperivew) - the fix for these is to add an `id` attribute on the container of the behavior.
Some other examples had improper styling.

See commits breakdown.


hgray-instawork (approved):
It is probably best practice to use ids so we don't have to rely on generated uuids, but the fix is forthcoming in Hyperview.

PR #1234, opened 2025-07-09
chore(monorepo): `hyperview/` package

flochtililoch:
Move every files related to Hyperview package under a new `hyperview/` directory.


flochtililoch:
> Curious why we couldn't leave it as `src` and point the scripts there.

The monorepo architecture requires a different package.json file at the root of the project, which would collide with Hyperview's

PR #1238, opened 2025-07-14
chore(hv-doc): force state refresh

hgray-instawork:
An issue exists where the document state can become stale in the `hv-doc` component when mutations occur without cloning the document. For example, the auto-complete component is adding a `<text-field>` element [during construction](https://github.com/Instawork/mobile/blob/ccd92c5ab5ea1c8e30546d01a3894f493b767d9a/core-mobile/src/components/hyperview-autocomplete-input/index.tsx#L65).

This problem was previously hidden by the [state refresh](https://github.com/Instawork/hyperview/blob/ffd315ab2834f061573ce7e4b2114d600455aa4d/src/core/components/hv-screen/index.js#L61-L75) in `<hv-screen>`. This temporary fix restores this functionality.

| Before | After |
| -- | -- |
| before | after |

See videos:
[before](https://github.com/user-attachments/assets/17cbb69f-8bd3-46b6-9609-11f88233c265)
[after](https://github.com/user-attachments/assets/985e3be0-1f60-4cf7-9f32-59734483e25a)

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1210788582235847?focus=true)

### DIFF
--- a/src/elements/hv-route/hv-route.tsx
+++ b/src/elements/hv-route/hv-route.tsx
@@ -70,24 +70,6 @@ class HvRouteInner extends PureComponent<Types.InnerRouteProps, ScreenState> {
     );
   };
 
-  componentDidMount(): void {
-    // This legacy behavior is required to ensure the document state is updated
-    // after the initial load behaviors are applied.
-    // See: https://github.com/Instawork/hyperview/pull/1238
-    // TODO: remove this once we remove HvDoc localDoc
-
-    const needsSubStack = this.props.route?.params?.needsSubStack
-      ? this.props.route.params.needsSubStack
-      : false;
-
-    const renderElement: Element | undefined = needsSubStack
-      ? undefined
-      : this.getRenderElement();
-    if (renderElement?.localName === LOCAL_NAME.SCREEN) {
-      this.props.setScreenState({});
-    }
-  }
-
   /**
    * Build the <HvScreen> component with injected props
    */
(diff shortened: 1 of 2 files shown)


flochtililoch (commented):
Was this issue discovered with any other components than our custom autocomplete? Are we considering not cloning the doc before making mutations something that should work? In other words, is the issue at the Hyperview level or at the custom component level, and if the latter, should we fix the said component(s)?

hgray-instawork:
> Was this issue discovered with any other components than our custom autocomplete? Are we considering not cloning the doc before making mutations something that should work? In other words, is the issue at the Hyperview level or at the custom component level, and if the latter, should we fix the said component(s)?

This was the only component which failed automation tests but I can't know for sure if others are impacted. There are additional changes to the doc which may not be accurately reflected in the `localDoc`. I noticed the `ran-once` value wasn't present on the `localDoc` until after this force.

The way the behaviors and state are currently handled, we are expecting the clone in order to refresh state. The eventual goal is to rework the behaviors to not require the clone. That being said, the manipulation of dom elements by a component is probably bad practice and we should consider rewriting this to use a behavior instead of direct mutation.

flochtililoch (approved):
we now call `getRenderElement` twice - I don't know how long this temp fix should remain, but it might be easy to cache the result (assuming it doesn't change for a given instance).

hgray-instawork:
Relocated the state change back to `hv-screen` where it originally lived. This avoids the issue with duplication of code in `hv-route`. https://github.com/Instawork/hyperview/pull/1238/commits/32a803f73581db63f3ca7f16d07aa85a2a605534

PR #1242, opened 2025-08-09
fix: correct ElementErrorComponent type definition

### LINKED ISSUE #1128: ElementErrorComponent type incorrect in root type definition
https://github.com/Instawork/hyperview/blob/54697034455a8fed386fde5843e4201e5123742f/src/core/components/hv-root/types.ts#L34

```
elementErrorComponent?: ComponentType<ErrorProps>;
errorScreen?: ComponentType<ErrorProps>;
```

 -> this uses the same `ErrorProps` type definition as `errorScreen` but they actually have separate `Props` definitions. Causes Typescript to complain if you override with a custom load-element error component. Resolved by the following change to import the right type with a suitable alias and swap it in to the type definitions:

```
+ --> import type { Props as ElementErrorProps } from 'hyperview/src/core/components/load-element-error';
import type { Props as LoadingProps } from 'hyperview/src/core/components/loading';
...
--> elementErrorComponent?: ComponentType<ElementErrorProps>;
errorScreen?: ComponentType<ErrorProps>;
```

I'll raise a PR for it as and when I get a minute free, unless anyone else gets to it first. Any preference on target branch?

rajanarahul93:
# Fix ElementErrorComponent type definition

## Problem
Fixes issue #1128 where `elementErrorComponent` and `errorScreen` were both using the same `ErrorProps` type definition, causing TypeScript compilation errors when users tried to override with custom element error components.

The issue was that:
- `elementErrorComponent` should use props from `load-element-error` component: `{ error: Error; onPressReload: () => void; onPressClose: () => void; }`
- `errorScreen` should use props from `load-error` component: `{ back: () => void; error: Error | null | undefined; onPressReload: () => void; onPressViewDetails: (uri: string) => void; }`

## Solution
- Updated `elementErrorComponent` to use `ElementErrorProps` from `load-element-error` component
- Kept `errorScreen` using `ErrorProps` from `load-error` component
- Added proper exports for `load-element-error` component types
- Fixed component structure to follow project conventions

## Changes Made
- `src/types.ts`: Added import for `ElementErrorProps` and updated type definition
- `src/elements/hv-screen/types.ts`: Updated to use `ElementErrorProps`
- `src/elements/hv-route/types.ts`: Updated to use `ElementErrorProps`
- `src/components/load-element-error/`: Restructured to properly export types

## Testing
- ✅ TypeScript compilation passes
- ✅ ESLint and Prettier checks pass
- ✅ All unit tests pass (376 tests)
- ✅ No breaking changes to existing functionality

## Breaking Changes
None. This is a type-level fix that resolves compilation errors without changing runtime behavior.

Closes #1128


flochtililoch:
hi @rajanarahul93 - wanted to check if you'd be able to push your fix so we could get this PR integrated? Thanks!

rajanarahul93:
Hi @flochtililoch, could you please let me know if any further changes are needed on this PR before it can be merged?

hgray-instawork:
@rajanarahul93 , thanks very much for catching this issue. We are in the final stages of implementing a few changes based around your suggestions. We decided to avoid the deprecated flag and instead to just correct the types. We are hoping to release a new version of Hyperview in the next few days which should resolve this issue for you.

PR #1265, opened 2025-10-31
feat(scrollable): allow disabling "over scroll" effect

flochtililoch:
Scrollables by default will allow additional pull above the first / below the last item. The attribute `over-scroll` set to `false` allows to disable this behavior.

https://github.com/user-attachments/assets/553a3a06-6746-41b0-bb83-0f8af527b9d3


hgray-instawork (approved):
Looks good. Considering the terminology of "bounce" and the more common "elastic scrolling", I wonder if "over scroll" is the right name?

flochtililoch:
> Looks good. Considering the terminology of "bounce" and the more common "elastic scrolling", I wonder if "over scroll" is the right name?

the term comes from the android implementation (`overScrollMode`) which felt more related to scroll than "bounce", but I could see how `elastic-scroll` could work too. @adamstep any opinion?

adamstep:
I'm fine with overscroll as the term, like Florent said it's nice that it has "scroll" in the name.

PR #1292, opened 2026-02-05
feat(navigation): demonstrate nested tab navigators

hgray-instawork:
Creating a demo in the navigation section of the demo app to illustrate how navigators can be nested to provide a bottom tab/top tab UI. The nested navigator exists in a separate file to show have the navigation hierarchy can be dynamically expanded.

Added an example in Advanced/Case Studies for a top nav bar.

Updated the demo to use nested navigation for top tabs.

| Demo | Nested | Top tab bar |
| -- | -- | -- |
| img | vid | top |

[Asana](https://app.asana.com/1/47184964732898/project/1204008699308084/task/1213137896710762?focus=true)


flochtililoch (commented):
Would it make sense to update the implementation of the demo app so it also implements this technique for its top tabs?

PR #1307, opened 2026-07-08
feat: `collapsable`/`collapsable-children` attributes on view-like elements

flochtililoch:
These control the equivalent RN's `View` props, which prevents view from being collapsed when explicitly set to `false` (default to `true` when new architecture is enabled). In addition, automatically add the prop when a test ID is set (so the view remains selectable by the test automation framework).

### DIFF
--- a/demo/backend/ui/ui-elements/view/view/index.xml.njk
+++ b/demo/backend/ui/ui-elements/view/view/index.xml.njk
@@ -63,6 +63,4 @@ hv_button_behavior: "back"
       </view>
     </view>
   </view>
-
-  </view>
 {% endblock %}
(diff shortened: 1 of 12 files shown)


adamstep:
What's an example in which we would want to explicitly set this to `false`? Perhaps the docs can shed some light on that, the RN docs also don't mention any use cases.

flochtililoch:
> What's an example in which we would want to explicitly set this to `false`? Perhaps the docs can shed some light on that, the RN docs also don't mention any use cases.

For my initial pass on automation failures with new arch, transparent overlay views with attached behaviors need that prop to continue working correctly. This is still in flux: this is a feature we know we need with new arch, and there might be other use-cases we haven't discovered yet.

hgray-instawork (approved):

PR #1308, opened 2026-07-10
chore(hv-route): improve back behavior handling

hgray-instawork:
Address several issues with back behaviors:

1. hv-route.tsx (listener refactor) — Splits the single monolithic useEffect into four focused effects with precise dependencies.
2. Introduces isFocusedRef (updated by focus/blur) to reliably gate back behavior dispatch mid-swipe, and deferredNavEventRef to coalesce state events and defer URL updates until transitionEnd, reducing animation hitches from synchronous setDoc calls during native animations.
3. hv-route.tsx (post-animation state mutations) — handleFocus and handleState now defer their setDoc calls to transitionEnd, replacing the navStateMutationsDelay experiment flag with a reliable event-driven mechanism.
4. hv-route.tsx (overflow fix) — handleBeforeRemove now filters the back behavior registry to only elements whose ancestors are not hidden. The hide+close pattern used by back behavior views relies on action="hide" setting hide="true" synchronously before action="close" fires. The visibility check reads that attribute directly, so the second handleBeforeRemove invocation from close sees no visible elements and allows navigation through — this avoids any timing issues.

---
  - https://app.asana.com/0/0/1216436336425045


flochtililoch (approved):
Thanks for the fix. I assume this needs pre-release testing?

hgray-instawork:
Testing completed in https://github.com/Instawork/mobile/pull/8582

PR #1329, opened 2026-08-05
chore(demo): navigate 'pop' behavior

hgray-instawork:
Illustrate navigate `pop` behavior to pop back to an existing route instead of adding a duplicate.

| iOS | Android |
| -- | -- |
| iOS | android |


flochtililoch (approved):
Minor: I assume the "pop" terminology comes from React Navigation, but Hyperview docs uses ["unwind"](https://hyperview.org/docs/reference_behavior_attributes#navigate) to describe this action.

PR #1331, opened 2026-08-10
feat(navigation): support React Navigation 7

hgray-instawork:
## Summary

Support React Navigation 7 while retaining React Navigation 6 compatibility and Hyperview's existing public navigation behavior. Consumers can continue using the same HXML actions, bare route IDs, deep links, and integration APIs under either major version.

## Changes and rationale

- **Support both custom navigator contracts.** React Navigation 7 adds `describe`, locale direction, and updated state/type requirements. Hyperview detects v7-only capabilities at runtime so the same stack and tab implementations remain valid under v6.
- **Resolve bare nested route IDs explicitly.** React Navigation 7 no longer searches child navigators for `navigate('route-id')`. Hyperview resolves mounted navigation state first, then uses an HXML fallback for unloaded routes reached by deep links and external navigation refs.
- **Preserve existing navigation semantics.** Compatible `pop: true` payloads retain Hyperview's unwind behavior, while follow-up actions after `close` are redirected from stale child navigators to their active parent.
- **Preserve screen presentation and transitions.** Screens provide both v6 `animationEnabled` and v7 `animation`, explicitly distinguish card and modal presentation, and retain the existing card interpolators so RN6 transitions remain unchanged. Explicitly requested sub-stacks carry a stable route ID and flag into `HvNavigator` so nested modal routes render consistently under both versions.
- **Fix reload route-state synchronization exposed during React Navigation 7 testing.** A replacement document could render while the route params still contained the previous URL. Hyperview now updates route identity only after a successful load.

## Validation

- Manually validated Hyperview demo and MDS navigation flows on iOS and Android.

### DIFF
--- a/src/components/navigator-stack/index.tsx
+++ b/src/components/navigator-stack/index.tsx
@@ -1,7 +1,11 @@
 import * as CustomStackRouter from 'hyperview/src/components/navigator-stack/router';
 import * as React from 'react';
-import type { NavigationState, ParamListBase } from '@react-navigation/routers';
-import type { Props, StackOptions } from './types';
+import type {
+  CompatibleStackViewProps,
+  NavigationBuilderWithDescribe,
+  Props,
+  StackOptions,
+} from './types';
 import {
   StackActionHelpers,
   StackNavigationState,
@@ -13,20 +17,22 @@ import {
   StackNavigationOptions,
   StackView,
 } from '@react-navigation/stack';
-import type { EventMapBase } from '@react-navigation/native';
+import {
+  isReactNavigation7,
+  useCompatibleLocale,
+} from 'hyperview/src/services/navigator/helpers';
+import type { ParamListBase } from '@react-navigation/routers';
 import { useHvDocContext } from 'hyperview/src/elements/hv-doc';
 import { useHyperview } from 'hyperview/src/contexts/hyperview';
+
+const CompatibleStackView = StackView as React.ComponentType<CompatibleStackViewProps>;
 
 const CustomStackNavigator = (props: Props) => {
   const { getSourceDoc } = useHvDocContext();
   const { entrypointUrl } = useHyperview();
+  const { direction } = useCompatibleLocale();
 
-  const {
-    state,
-    descriptors,
-    navigation,
-    NavigationContent,
-  } = useNavigationBuilder<
+  const builder = useNavigationBuilder<
     StackNavigationState<ParamListBase>,
     StackOptions,
     StackActionHelpers<ParamListBase>,
@@ -40,11 +46,15 @@ const CustomStackNavigator = (props: Props) => {
     initialRouteName: props.initialRouteName,
     screenOptions: props.screenOptions,
   });
+  const { state, descriptors, navigation, NavigationContent } = builder;
+  const { describe } = (builder as unknown) as NavigationBuilderWithDescribe;
 
   return (
     <NavigationContent>
-      <StackView
+      <CompatibleStackView
+        describe={isReactNavigation7 ? describe : undefined}
         descriptors={descriptors}
+        direction={isReactNavigation7 ? direction : undefined}
         navigation={navigation}
         state={state}
       />
@@ -52,9 +62,4 @@ const CustomStackNavigator = (props: Props) => {
   );
 };
 
-export default createNavigatorFactory<
-  Readonly<NavigationState>,
-  StackNavigationOptions,
-  EventMapBase,
-  typeof CustomStackNavigator
->(CustomStackNavigator);
+export default createNavigatorFactory(CustomStackNavigator);
--- a/src/components/navigator-stack/index.tsx
+++ b/src/components/navigator-stack/index.tsx
@@ -1,4 +1,5 @@
 import * as CustomStackRouter from 'hyperview/src/components/navigator-stack/router';
+import * as NavigatorService from 'hyperview/src/services/navigator';
 import * as React from 'react';
 import type {
   CompatibleStackViewProps,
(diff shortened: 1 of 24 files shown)


test/helpers/navigation.ts
```diff
@@ -0,0 +1,14 @@
...
+
+type StackState = Omit<StackNavigationState<ParamListBase>, 'preloadedRoutes'>;
+
+export const createStackNavigationState = (
+  state: StackState,
+): StackNavigationState<ParamListBase> =>
+  (({
+    ...state,
+    preloadedRoutes: [],
+  } as unknown) as StackNavigationState<ParamListBase>);
```
flochtililoch: Do you know if we still need this hack? Technically, we are returning a type that's not `StackNavigationState<ParamListBase>`

src/components/navigator-stack/index.tsx
```diff
@@ -13,20 +17,22 @@ import {
   StackNavigationOptions,
   StackView,
 } from '@react-navigation/stack';
-import type { EventMapBase } from '@react-navigation/native';
+import {
+  isReactNavigation7,
+  useCompatibleLocale,
+} from 'hyperview/src/services/navigator/helpers';
```
flochtililoch: I don't think we should reach to the internals of this service (might require updating the module's public API)

```suggestion
} from 'hyperview/src/services/navigator';
```

test/integration/navigation.test.tsx
flochtililoch, line 1: is the location for this file adequate?

src/components/navigator-tab/router.tsx
```diff
@@ -0,0 +1,33 @@
...
+} from '@react-navigation/routers';
+import {
+  expandNestedNavigate,
+  findPathFromDom,
+} from 'hyperview/src/services/navigator/helpers';
+import type { TabNavigationState } from '@react-navigation/native';
+import type { TabOptions } from './types';
+import { TabRouter } from '@react-navigation/native';
+
+export const Router = (options: TabOptions) => {
```
flochtililoch: can we type the returned value for this function?

flochtililoch (commented):
Couple minor comments

test/integration/navigation.test.tsx
hgray-instawork: This seemed like the best location for integration tests for Hyperview’s navigation components, custom routers, and React Navigation. The coverage spans multiple areas, so it doesn’t fit cleanly under services/navigator or a single component. Keeping it separate also avoids adding navigation-specific setup to hyperview.test.tsx. Do you have a suggested other location?

src/components/navigator-tab/router.tsx
hgray-instawork: https://github.com/Instawork/hyperview/pull/1331/commits/1b40c867bc05b197e3db95141c8d511bc28f3953

src/components/navigator-stack/index.tsx
hgray-instawork: https://github.com/Instawork/hyperview/pull/1331/commits/71ac22f5a65bfbecb6a612eb494bcefb11dbfdea

test/helpers/navigation.ts
hgray-instawork: https://github.com/Instawork/hyperview/pull/1331/commits/737c1b03e39322c1d7d46d80a241d22c0e9c7b08

test/integration/navigation.test.tsx
flochtililoch: I would suggest a new `integration` directory under `test/`. Maybe we don't need the `hyperview.` prefix either.

src/components/navigator-stack/router.tsx
```diff
@@ -3,14 +3,20 @@ import type {
   RouterConfigOptions,
 } from '@react-navigation/routers';
 import type { RouterRenameOptions, StackOptions } from './types';
+import {
+  expandNestedNavigate,
+  findPathFromDom,
+} from 'hyperview/src/services/navigator/helpers';
```
flochtililoch: same issue I previously flagged

```suggestion
} from 'hyperview/src/services/navigator';
```

src/components/navigator-stack/types.ts
```diff
@@ -1,6 +1,10 @@
 import * as React from 'react';
-import type { RouterConfigOptions } from '@react-navigation/native';
-import type { StackNavigationOptions } from '@react-navigation/stack';
+import type {
+  StackNavigationOptions,
+  StackView,
+} from '@react-navigation/stack';
+import type { Locale } from 'hyperview/src/services/navigator/types';
```
flochtililoch: same here

```suggestion
import type { Locale } from 'hyperview/src/services/navigator';
```

src/components/navigator-tab/router.tsx
```diff
@@ -0,0 +1,33 @@
+import type {
+  ParamListBase,
+  RouterConfigOptions,
+} from '@react-navigation/routers';
+import {
+  expandNestedNavigate,
+  findPathFromDom,
+} from 'hyperview/src/services/navigator/helpers';
```
flochtililoch: same here

```suggestion
} from 'hyperview/src/services/navigator';
```

hgray-instawork: https://github.com/Instawork/hyperview/pull/1331/commits/85241434a045f8b7b343af7e292c21eceed6f298

src/components/navigator-stack/types.ts
hgray-instawork: https://github.com/Instawork/hyperview/pull/1331/commits/85241434a045f8b7b343af7e292c21eceed6f298

src/components/navigator-stack/router.tsx
hgray-instawork: https://github.com/Instawork/hyperview/pull/1331/commits/85241434a045f8b7b343af7e292c21eceed6f298

test/integration/navigation.test.tsx
hgray-instawork: Thanks. https://github.com/Instawork/hyperview/pull/1331/commits/4cd9ab02d9793bdd29ec571c3ab716b6185504e8

flochtililoch (approved):

PR #1336, opened 2026-08-19
feat(hyper-ref): allow disabling touchables

flochtililoch:
Support a new `disabled` attribute on elements that support behaviors. When set to `false`, carries forward as a prop on the `TouchableOpacity` instance. This allows automation frameworks (UIAutomator2, XCUITest) to read the disabled state of a button and select/assert on its value.

### DIFF
--- a/schema/core.xsd
+++ b/schema/core.xsd
@@ -481,6 +481,7 @@
     <xs:attribute name="id" type="hv:ID" />
     <xs:attribute name="key" type="hv:KEY" />
     <xs:attribute name="hide" type="xs:boolean" />
+    <xs:attribute name="disabled" type="xs:boolean" />
     <xs:attribute name="avoid-keyboard" type="xs:boolean" />
     <xs:attribute name="safe-area" type="xs:boolean" />
     <xs:attribute name="value" type="xs:string" />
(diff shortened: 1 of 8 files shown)


schema/core.xsd
```diff
@@ -481,6 +481,7 @@
     <xs:attribute name="id" type="hv:ID" />
     <xs:attribute name="key" type="hv:KEY" />
     <xs:attribute name="hide" type="xs:boolean" />
+    <xs:attribute name="disabled" type="xs:boolean" />
```
hgray-instawork, line 484: Will this attribute carry forward to the view aliases like body, form, and item?

hgray-instawork (approved):

schema/core.xsd
flochtililoch: Yes, any element that renders a touchable will support this.

hgray-instawork: Should we update docs to reflect that?

flochtililoch: which part? I already updated the docs for view, text and image

hgray-instawork: docs/reference_form.md, docs/reference_item.md might be worthwhile.

PR #1338, opened 2026-08-25
feat(demo): switch from Expo Go to a local development build

hgray-instawork:
## Summary
- Switch the demo from Expo Go to a local development build (`expo-dev-client` + `expo run:ios` / `expo run:android`).
- Display changed to **Hyperview**.
- Default XML server URL is `http://127.0.0.1:8085`. `0.0.0.0` is a listen address; iOS will not load it. Physical devices still pass `BASE_URL` with the machine’s LAN IP.
- Native bundle / application ID is `org.hyperview.demo`.
- Demo HXML sets `color="black"` on shared text styles. Expo Go’s Android theme defaults unstyled text to black; a development build uses DayNight `textColorPrimary` (faded).
- `update-demo.sh` installs `expo-dev-client` and `react-native-worklets`.

| iOS | Android |
| -- | -- |
| iOS | android |

[Asana](https://app.asana.com/0/0/1217876198464400)


demo/app.config.ts
```diff
@@ -1,20 +1,30 @@
...
+      NSAppTransportSecurity: {
+        NSAllowsLocalNetworking: true,
+      },
+    },
     supportsTablet: true,
   },
-  name: 'demo',
-  newArchEnabled: true,
+  name: 'Hyperview Demo',
+  newArchEnabled: false,
```
flochtililoch: we probably don't want to switch that off?

demo/app.config.ts
```diff
@@ -1,20 +1,30 @@
...
+      backgroundColor: '#000000',
+      foregroundImage: './assets/icon.png',
     },
+    package: 'org.hyperview.demo',
+    usesCleartextTraffic: true,
   },
   assetBundlePatterns: ['**/*'],
   extra: {
-    baseUrl: process.env.BASE_URL || 'http://0.0.0.0:8085',
+    baseUrl: process.env.BASE_URL || 'http://127.0.0.1:8085',
```
flochtililoch, line 12: was this change necessary?

demo/app.config.ts
hgray-instawork: See the bullet above in the description. There's a known bug with our current combination of packages.

flochtililoch: I see, are you re-enabling it in your upcoming PR that uses native stack?

demo/app.config.ts
```diff
@@ -1,20 +1,30 @@
...
+      CFBundleDisplayName: 'Hyperview Demo',
+      NSAppTransportSecurity: {
+        NSAllowsLocalNetworking: true,
+      },
+    },
     supportsTablet: true,
   },
-  name: 'demo',
-  newArchEnabled: true,
+  name: 'Hyperview Demo',
```
flochtililoch: Minor nit: since the name is truncated on Android, might be best renaming to "Hyperview" or "HV Demo"

flochtililoch (commented):
Do we also need to update scripts/update-demo.sh? Will you take care of updating to the latest Expo in a follow up PR?

demo/app.config.ts
hgray-instawork: Expo Go accepted `0.0.0.0` on iOS; the development build does not (Network request failed). Android still works via `adb reverse`. We either change this default to 127.0.0.1, or iOS simulator launches need `BASE_URL=http://127.0.0.1:8085`.

demo/app.config.ts
hgray-instawork: `native-stack` won't fix it. The issue is with `react-native-screens`. We can't re-enable the flag until we test with a newer version (>= 4.18.0).

hgray-instawork: Removed as part of upgrade to Expo 57

demo/app.config.ts
hgray-instawork: Updated to "Hyperview" to fit better

flochtililoch (approved):

PR #1352, opened 2026-09-15
chore(scroll): support content insets

hgray-instawork:
## Summary

Adds opt-in `content-insets` support for `<view>`, `<list>`, and `<section-list>` so content remains reachable while drawing beneath translucent native UI.

## Why

Hardcoded or broadly applied padding cannot account for dynamic native UI and can shorten the viewport, prevent full-bleed content, or affect the wrong nested scroll view. The `content-insets` attribute enables automatic handling only on scroll views that explicitly opt in.

## Implementation

- Adds the optional `content-insets` schema attribute and documentation.
- Uses React Native's [`contentInsetAdjustmentBehavior="automatic"`](https://reactnative.dev/docs/0.77/scrollview#contentinsetadjustmentbehavior-ios) for opted-in native scroll views.
- Updates the demo to allow tab routes to render full bleed.

## Compatibility

The feature is optional and controlled per scroll element. Documents without `content-insets="true"` remain unchanged. The attribute has no effect on Android.

| iOS 18 | iOS 26 | Android
| -- | -- | -- |
| ios18 | ios26 | android |


flochtililoch (approved):
LGTM

